PR #24, opened 2019-04-04
Fixed #21: Fixed incorrect Anaconda detection in install and uninstall scripts.

### LINKED ISSUE #21: Installation issue on Mac
Following https://docs.hummingbot.io/installation/source/
Without anaconda installed:

```
Aarons-iMac:hummingbot polymorpher$ ./install 
usage: dirname path
env: list: No such file or directory
env: create: No such file or directory
```

martinkou:
As title.

The old `${CONDA_EXE} info` detection method would always succeed, even without Anaconda installed, because `info` is a valid shell command by itself. The fix is to test whether the detected Anaconda path to be empty or not.

### DIFF
--- a/install
+++ b/install
@@ -7,7 +7,7 @@ if [ "${CONDA_EXE} " == " " ]; then
     CONDA_EXE=$((find /opt/conda/bin/conda || find ~/anaconda3/bin/conda) 2>/dev/null)
 fi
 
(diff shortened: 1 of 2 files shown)


wpank:
Tested with Ubuntu 18.10, ZSH, and without anaconda. This fix now gives the expected error message :+1:

PR #44, opened 2019-04-10
Use batch order api for Radar Relay to minimize api requests

chowryan:
Radar Relay has an api for returning the latest 20 orders of an ethereum address. We can use this to look up the status of tracked orders and the single order status api as a fallback.

### DIFF
--- a/wings/radar_relay_market.pyx
+++ b/wings/radar_relay_market.pyx
@@ -7,6 +7,7 @@ import logging
 import math
 import time
 from typing import (
+    Any,
     Dict,
     List,
     Optional,
@@ -78,7 +79,7 @@ cdef class TradingRule:
         public int amount_decimals              # Max amount of decimals in quote token (amount)
 
     @classmethod
-    def parse_exchange_info(cls, markets: List[Dict[str, any]]) -> List[TradingRule]:
+    def parse_exchange_info(cls, markets: List[Dict[str, Any]]) -> List[TradingRule]:
         cdef:
             list retval = []
         for market in markets:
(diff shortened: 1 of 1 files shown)


wings/radar_relay_market.pyx
```diff
@@ -339,6 +340,30 @@ cdef class RadarRelayMarket(MarketBase):
                 self._trading_rules[trading_rule.symbol] = trading_rule
             self._last_update_trading_rules_timestamp = current_timestamp
 
+    async def get_account_orders(self) -> List[Dict[str, Any]]:
+        list_account_orders_url = f"{RADAR_RELAY_REST_ENDPOINT}/accounts/{self._wallet.address}/orders"
+        return await self._api_request(http_method="get", url=list_account_orders_url)
+
+    async def get_order(self,
```
martinkou: Adding the account orders map to `get_order()` here violates the single responsibility principle... Ideally `get_order()` should only get a single order. The logic to ignore already obtained orders should go in `_get_order_updates()`

wings/radar_relay_market.pyx
```diff
@@ -339,6 +340,30 @@ cdef class RadarRelayMarket(MarketBase):
...
+        return await self._api_request("get", url=order_url)
+
+    async def _get_order_updates(self) -> List[Dict[str, Any]]:
+        account_orders_list = await self.get_account_orders()
+        account_orders_map = {}
+        for account_order in account_orders_list:
+            account_orders_map[account_order["orderHash"]] = account_order
+        tasks = [self.get_order(o.exchange_order_id, account_orders_map)
+            for o in list(self._in_flight_limit_orders.values())
+            if o.exchange_order_id]
```
martinkou: You should check for the non-existence of the order id in the `accounts_order_map` here instead of adding logic to `get_order()`

chowryan:
made a diff PR for this change, closing this one

PR #299, opened 2019-05-24
(fix) Fix for trading tokens which require more than 5/6 digits of pr…

veecos2:
…ecision

**Optional**:
- Related Github issue:
- Related Clubhouse Story:

### DIFF
--- a/hummingbot/market/binance/binance_market.pyx
+++ b/hummingbot/market/binance/binance_market.pyx
@@ -972,8 +972,12 @@ cdef class BinanceMarket(MarketBase):
             self.c_start_tracking_order(order_id, -1, symbol, True, decimal_amount, order_type)
             order_result = None
             if order_type is OrderType.LIMIT:
+                order_price = self.quantize_order_price(symbol, price)
+                #order_price = f"{price:f}"
                 order_price = f"{Decimal(str(price)):f}"
                 order_decimal_amount = f"{Decimal(str(decimal_amount)):f}"
+                #order_decimal_amount = f"{decimal_amount:f}"
+                #self.logger().info(f"decimal_amount:{decimal_amount} order_price:{order_price}")
                 order_result = await self.query_api(self._binance_client.order_limit_buy,
                                                     symbol=symbol,
                                                     quantity=order_decimal_amount,
@@ -1042,11 +1046,16 @@ cdef class BinanceMarket(MarketBase):
 
 
         try:
+
             self.c_start_tracking_order(order_id, -1, symbol, False, decimal_amount, order_type)
             order_result = None
             if order_type is OrderType.LIMIT:
+                order_price = self.quantize_order_price(symbol, price)
                 order_price = f"{Decimal(str(price)):f}"
+                #order_price = f"{price:f}"
+                #self.logger().info(f"decimal_amount:{decimal_amount} order_price:{order_price}")
                 order_decimal_amount = f"{Decimal(str(decimal_amount)):f}"
+                #order_decimal_amount = f"{decimal_amount:f}"
                 order_result = await self.query_api(self._binance_client.order_limit_sell,
                                                     symbol=symbol,
                                                     quantity=order_decimal_amount,
--- a/hummingbot/market/binance/binance_market.pyx
+++ b/hummingbot/market/binance/binance_market.pyx
@@ -1,4 +1,5 @@
 import math
+from numpy import format_float_positional
 import aiohttp
 from aiokafka import (
     AIOKafkaConsumer,
@@ -975,7 +976,7 @@ cdef class BinanceMarket(MarketBase):
                 order_result = await self.query_api(self._binance_client.order_limit_buy,
                                                     symbol=symbol,
                                                     quantity=str(decimal_amount),
-                                                    price=price,
+                                                    price=format_float_positional(price),
                                                     newClientOrderId=order_id)
             elif order_type is OrderType.MARKET:
                 order_result = await self.query_api(self._binance_client.order_market_buy,
@@ -1037,6 +1038,7 @@ cdef class BinanceMarket(MarketBase):
             raise ValueError(f"Sell order amount {decimal_amount} is lower than the minimum order size "
                              f"{trading_rule.min_order_size}.")
 
+
         try:
             self.c_start_tracking_order(order_id, -1, symbol, False, decimal_amount, order_type)
             order_result = None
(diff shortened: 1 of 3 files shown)


hummingbot/market/binance/binance_market.pyx
```diff
@@ -975,7 +976,7 @@ cdef class BinanceMarket(MarketBase):
                 order_result = await self.query_api(self._binance_client.order_limit_buy,
                                                     symbol=symbol,
                                                     quantity=str(decimal_amount),
-                                                    price=price,
+                                                    price=format_float_positional(price),
```
martinkou: Please use `c_quantize_order_price()` instead. Don't hack this with a random numpy function since you may go over the maximum allowed precision and cause order failure.

martinkou:
Also, please add a unit test case.

martinkou:
Wait... this is not ok.

hummingbot/market/binance/binance_market.pyx
```diff
@@ -972,15 +972,18 @@ cdef class BinanceMarket(MarketBase):
             self.c_start_tracking_order(order_id, -1, symbol, True, decimal_amount, order_type)
             order_result = None
             if order_type is OrderType.LIMIT:
+                order_price = f"{Decimal(str(price)):f}"
```
martinkou: This is no good. You should quantize the price first, and then convert to decimal string.

veecos2: order price is already quantized, but the problem is if you do Decimal(0.002485), its converted to 0.002485000000999 or some garbage value

test/test_binance_market.py
```diff
@@ -323,6 +323,29 @@ def test_cancel_all(self):
...
+        quantized_amount: Decimal = self.market.quantize_order_amount(symbol, amount)
+        self.market.buy("IOSTETH", amount)
+
+        # Intentionally setting invalid price to prevent getting filled
+        quantize_bid_price: Decimal = self.market.quantize_order_price(symbol, bid_price * 0.7)
+        quantize_bid_price -= Decimal('0.00000005')
+        quantize_ask_price: Decimal = self.market.quantize_order_price(symbol, ask_price * 1.5)
+        quantize_ask_price += Decimal('0.00000005')
+
+        self.market.buy(symbol, quantized_amount, OrderType.LIMIT, quantize_bid_price)
```
martinkou: This is a bad test case. The buy and sell functions are supposed to have floating point inputs because the strategies shouldn't care about these market details. What you should do is to give it a floating point number that goes beyond the required precision (i.e. an invalid order) and test that the market connector would make it a valid order again.

hummingbot/market/binance/binance_market.pyx
veecos2: So first convert it to string so that you can use the format :f for decimal

martinkou: order price is not quantized. See the type definition of c_buy() and c_sell() in `MarketBase`. It is intentionally designed that way to offload the order quantization work from strategy code.

veecos2: but let me try that method as well

martinkou: And by quantize I'm meaning you need to go call `c_quantize_order_price()` as I've commented before. Just converting a floating point to random decimal precision is NOT 'quantizing'. You need to consider the exchange's rules about that particular trading symbol's allowed precision for price and order sizes. That's why that function exists. Don't just go `Decimal(str(price))` and tell me that's quantization.

veecos2: The Pure MM strategy already quantizes it before sending it to binance market, which is what I meant not the above

veecos2: I will quantize it again in the binance market class

martinkou: The market classes is supposed to work for all strategies, how the pure MM handles it is irrelevant. In fact... if you pass in a price or order size that's not a floating point then that's wrong because you're violating the argument type specs.

test/test_binance_market.py
martinkou: e.g. Binance API specifies filter rules about what the minimum price and order size increments are for every trading pair. So for IOSTETH the minimum price tick is 1e-8 ETH and the minimum order size quantum is 1 IOST.

So the expectation here is, you should be able to give the Binance market connector an order to buy 5.012 IOST at 1.2321e-7 ETH and it'll quantize the size to 5 IOST and the price to 0.00000012. And you should check that the order went through.

test/test_binance_market.py
```diff
@@ -316,12 +316,37 @@ def test_cancel_all(self):
...
             self.assertEqual(cr.success, True)
+        self.market.sell("LOOMETH", amount)
+
+    def test_order_price_precision(self):
+        symbol = "IOSTETH"
+        bid_price: float = self.market.get_price(symbol, True)
+        ask_price: float = self.market.get_price(symbol, False)
+        amount: float = 0.02 / bid_price
+        quantized_amount: Decimal = self.market.quantize_order_amount(symbol, amount)
+        self.market.buy("IOSTETH", amount)
```
martinkou: Market order functions are asynchronous... it's generally not true that you'll have the IOST tokens in your wallet right after calling this function.

The way we do these market tests is to assume the wallet have the tokens needed to carry out the tests to begin with. That's why, e.g. the `test_buy_and_sell()` function starts by asserting you have a certain amount of ETH to start with. I'd advise you to do a similar assert in the beginning to make sure the account has enough ETH and IOST to make the limit orders first, and make the limit orders far, far away from the trading price s.t. they won't be traded.

hummingbot/market/binance/binance_market.pyx
veecos2: I have added the quantization. However just using the :f tag to the string drops precision from 8 to 6 digits.

veecos2: So I have kept the original method. Let me know if you want to implement it in some other way.

veecos2: Oh so basically this is from the bug yingdan posted for iost token, so I am testing it there

martinkou: This is not a reason to keep the original method. You should find out why the quantization gave you the 6 digits price, and find out whether that's correct or not by comparing it to Binance's quantization rules from their PRICE_FILTER spec. We've tested today that the :f tag can give you >24 digits so that's not the cause of it.

hummingbot/market/binance/binance_market.pyx
```diff
@@ -1042,11 +1046,16 @@ cdef class BinanceMarket(MarketBase):
...
+
             self.c_start_tracking_order(order_id, -1, symbol, False, decimal_amount, order_type)
             order_result = None
             if order_type is OrderType.LIMIT:
+                order_price = self.quantize_order_price(symbol, price)
                 order_price = f"{Decimal(str(price)):f}"
+                #order_price = f"{price:f}"
+                #self.logger().info(f"decimal_amount:{decimal_amount} order_price:{order_price}")
                 order_decimal_amount = f"{Decimal(str(decimal_amount)):f}"
+                #order_decimal_amount = f"{decimal_amount:f}"
```
martinkou: This is not broken, don't change it.

martinkou:
I'm going to fix this patch and the unit test case tonight.

PR #344, opened 2019-06-05
fix/balance_check_and_binance_min_notional_check

veecos2:

**Optional**:
- Related Github issue:
https://github.com/CoinAlpha/hummingbot/issues/337
https://github.com/CoinAlpha/hummingbot/issues/341

- Related Clubhouse Story:
https://app.clubhouse.io/coinalpha/story/4251/pure-market-error-display-issue-display-correct-error-for-low-balances
https://app.clubhouse.io/coinalpha/story/4258/filled-orders-still-being-shown-in-active-orders-even-after-cancel-order-wait-time

1. checks whether binance limit orders pass the min_notional as per trading rules
2. fixes order failed event tag
3. For calculating sufficient balance looks at the price you are placing the buy order instead of top bid price in Pure MM
4. Stops tracking filled orders


hummingbot/market/binance/binance_market.pyx
```diff
@@ -1161,6 +1169,8 @@ cdef class BinanceMarket(MarketBase):
 
         # Add 1% as a safety factor in case the prices changed while making the order.
         if notional_size < float(trading_rule.min_notional_size) * 1.01:
+            self.logger().error(f"The order value is less than the minimum order value allowed and is not placed. "
```
martinkou: This is not an error. Strategies should be allowed to calculate any size they want, and they're just using this function to check whether it makes sense to emit a certain order at this market.

hummingbot/market/binance/binance_market.pyx
```diff
@@ -974,6 +978,10 @@ cdef class BinanceMarket(MarketBase):
             order_decimal_amount = f"{decimal_amount:f}"
             if order_type is OrderType.LIMIT:
                 decimal_price = self.c_quantize_order_price(symbol, price)
+                notional_size = decimal_amount * decimal_price
```
martinkou: On second thought... I think this is not the right approach here - even though I suggested this solution yesterday.

The problem has to do with how the quantize functions are normally used. e.g. in arbitrage strategy. The strategy checks whether it can expect to send a certain order to both markets, before it sends it. If you simply error one of the orders out like this, then it'll cause the user to be exposed.

We should allow these checks to proceed as intended without raising errors. So I guess the correct solution is to change the interface of `c_quantize_order_amount()` to accept an optional price keyword argument. This means the interface in market base will need to be changed and all market classes will need to have it.

hummingbot/market/binance/binance_market.pyx
veecos2: I believe the user should know why an order is not placed in the market. This is just logging to show they did not meet the notional value and that's why order is not placed.

hummingbot/market/binance/binance_market.pyx
veecos2: okays, will update this

hummingbot/market/binance/binance_market.pyx
martinkou: It's up to the strategy to decide whether that's an unrecoverable error to the user or not, not the market connector.

In general, anything we log as `.error()` should be bugs. If someone, e.g., has too little balance in his wallet s.t. the strategy cannot make orders - that's a recoverable error because it can be fixed by user action. Such a thing should be a `logger().network()` call emitted from the strategy instead.

veecos2:
made necessary changes

PR #345, opened 2019-06-05
feat / add fees to historynn

yzhang1994:

- Related Clubhouse Story: https://app.clubhouse.io/coinalpha/story/3786/add-exchange-fees-and-gas-info-to-performance-analytics


mifeng:
If possible, I think we should represent the fees with the following columns:
* Fee
* Fee coin
* Other costs

For trades in the ONE/BTC pair on Binance (if user is paying fees in BNB), these columns would be:
* `0.0023434`
* `BNB`
* `None`

For trades in the ONE/BTC pair on Binance (if user is **not** paying fees in BNB), these columns would be:
* `0.00004554`
* `BTC`
* `None`

For trades in the WETH/DAI pair on Radar, these columns would be:
* `0`
* `None`
* `[(ETH, 0.003443)]`

PR #354, opened 2019-06-07
feat / pure mm v2

martinkou:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:
https://app.clubhouse.io/coinalpha/story/4118/pure-mm-strategy-improvements

- Implemented basic plugin based architecture for pure marketing making V2 strategy
- Implemented the V1 pure MM order pricing and sizing logic as plugins
- Fixed the self stopping issue with V1 pure MM strategy


yzhang1994:
I'm running this on DDEX. Seems to be working.

hummingbot/strategy/pure_market_making/pure_market_making_v2.pyx
```diff
@@ -0,0 +1,709 @@
...
+
+    cdef object c_get_orders_proposal_for_market_info(self, object market_info, list active_orders):
+        cdef:
+            double last_trade_price
+            MarketBase maker_market = market_info.market
+            int actions = 0
+            list cancel_order_ids = []
+
+        # Before doing anything, ask the filter delegate whether to proceed or not.
+        if not self._filter_delegate.c_should_proceed_with_processing(self, market_info, active_orders):
```
veecos2: As exceptions in filter delegate, return True, shouldn't this be checking the opposite condition. Like if there is an exception in the filter delegate, then NO_Op_Orders is proposed

hummingbot/strategy/pure_market_making/pure_market_making_v2.pyx
```diff
@@ -0,0 +1,709 @@
...
+        self._owner.c_did_cancel_order(arg)
+
+
+cdef class PureMarketMakingStrategyV2(StrategyBase):
+    BUY_ORDER_COMPLETED_EVENT_TAG = MarketEvent.BuyOrderCompleted.value
+    SELL_ORDER_COMPLETED_EVENT_TAG = MarketEvent.SellOrderCompleted.value
+    ORDER_FILLED_EVENT_TAG = MarketEvent.OrderFilled.value
+    ORDER_CANCELLED_EVENT_TAG = MarketEvent.OrderCancelled.value
+    ORDER_EXPIRED_EVENT_TAG = MarketEvent.OrderExpired.value
+    TRANSACTION_FAILURE_EVENT_TAG = MarketEvent.TransactionFailure.value
```
veecos2: This is actually the orderfailure event.

veecos2:
Tested it working on ddex, binance and radar relay.

hummingbot/strategy/pure_market_making/pure_market_making_v2.pyx
martinkou: No, if there's an exception from filter delegate, then the exception is raised. The `True?` doesn't mean you'd see a `True` being returned from Python's side. What it means is that the C++ code would return a `True` and set an exception. Cython would automatically check for the exception when it sees a True and raise the exception on Python's side. Thus, you won't see the return value when an exception is raised.

martinkou:
Yes, further modifications will be needed for having multiple bids and asks. The modifications I made for this week is meant to be conservative s.t. we can release it faster.

martinkou:
@veecos2 I've back-ported one of your recent fixes to the tracking issue w.r.t. filled orders.

PR #392, opened 2019-06-13
Feat/allow multiple orders pure mm v2

veecos2:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:


hummingbot/client/hummingbot_application.py
```diff
@@ -881,6 +885,10 @@ def start(self, log_level: Optional[str] = None):
                 strategy_logging_options = PureMarketMakingStrategyV2.OPTION_LOG_ALL
 
                 self.strategy = PureMarketMakingStrategyV2(market_infos=[self.market_info],
+                                                           number_of_orders= number_of_orders,
```
martinkou: Please don't put plugin arguments in the pure market making strategy's constructor. Once we've changed the architecture to plugins, then the way to do it is to give the arguments to plugin constructor and then give the plugins to the strategy.

martinkou: Note the compatibility arguments for this thing have been renamed "legacy_xxx" s.t. it can serve as a drop-in replacement for v1 in case the user doesn't want to use it in the plugin way. But that's the only place where this is allowed and it's well labeled.

hummingbot/strategy/pure_market_making/pure_market_making_config_map.py
```diff
@@ -34,7 +34,12 @@ def is_valid_maker_market_symbol(value: str) -> bool:
                                                   prompt="What is your preferred quantity per order (denominated in "
                                                          "the base asset, default is 1)? >>> ",
                                                   default=1.0,
-                                                  type_str="float"),
+                                                 type_str="float"),
```
martinkou: Feels like it's gonna be a multiple-level choice here, instead of just once level. If you try to ask all questions for all plugins here, then most of the questions here will be irrelevant for users.

We're gonna need a different question sequence rather than trying to ask the user everything up-front.

(After choosing pure market making strategy)
- Do you want to configure pricing plugin?
- Do you want to configure sizing plugin?
(Assuming yes to the above questions)
- Which pricing plugin?
- Which sizing plugin?
- (plugin-specific questions)

The plugin maker will likely need to come up with the config options and include those in a specific variable name in his module. We'll need to discuss w/ @yzhang1994 on how to do this.

martinkou: For v0.9.0... we probably don't need to implement the full plugin config architecture. But we should start by asking the initial questions like "Do you want to configure staggered orders for market making?". If it's no then we proceed with the old constant spread / size questions. Basically we should start by having the multiple-level choice architecture.

Then in the next next version (say, 0.10.0)... we can build incrementally upon that and have a more complete architecture.

veecos2:
Made the necessary changes. All unit tests have passed and tested working on Binance.

PR #414, opened 2019-06-20
feat / add performance analysis

anish-neervannan:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:


hummingbot/client/hummingbot_application.py
```diff
@@ -1139,6 +1140,85 @@ def compare_balance_snapshots(self):
         lines = ["", "  Performance:"] + ["    " + line for line in str(df).split("\n")]
         self._notify("\n".join(lines))
 
+    def add_balances(self, msp, asset_type, asset_time, base, quote):
```
chowryan: add types

chowryan: can you also add a comment summarizing what this function does

chowryan: also try to avoid acronyms that arent common (i dont know what msp is)

hummingbot/client/hummingbot_application.py
```diff
@@ -1139,6 +1140,85 @@ def compare_balance_snapshots(self):
...
+
+        if asset_time == "starting":
+            amount = self.starting_balances.get(asset).get(market_name)
+        elif asset_time == "current":
+            amount = self.balance_snapshot().get(asset).get(market_name)
+        else:
+            raise ValueError("asset_time must be either 'starting' or 'current'")
+
+        if len(asset_list) == 0:
+            asset_list.insert(0, asset)
```
chowryan: use list.append instead

chowryan: seems like your list only ever has 2 items. code will be more clear if you use individual variables

hummingbot/client/hummingbot_application.py
```diff
@@ -1139,6 +1140,85 @@ def compare_balance_snapshots(self):
...
+            quote = asset_list
+
+        return base, quote
+
+    def analyze_performance(self):
+        if len(self.starting_balances) == 0:
+            self._notify("  Performance analysis is not available before bot starts")
+            return
+
+        starting_base = list()
```
chowryan: minor nitpick - i think ppl usually initialize with starting_base = []

hummingbot/client/hummingbot_application.py
```diff
@@ -1139,6 +1140,85 @@ def compare_balance_snapshots(self):
...
+                )
+                current_base, current_quote = self.add_balances(
+                    msp=msp,
+                    asset_type=asset_type,
+                    asset_time="current",
+                    base=current_base,
+                    quote=current_quote
+                )
+
+        market_pair_info = self.market_symbol_pairs[0]
```
chowryan: why are we only doing this for the first pair?

hummingbot/client/hummingbot_application.py
```diff
@@ -1215,4 +1295,3 @@ def bounty(self, register: bool = False, status: bool = False, terms: bool = Fal
         self.app.change_prompt(prompt=">>> ")
         self.app.toggle_hide_input()
         self.placeholder_mode = False
-
```
chowryan: don't delete empty line, standard practice is to leave empty line at end of a file

hummingbot/client/hummingbot_application.py
anish-neervannan: Replaced list with tuple

hummingbot/client/hummingbot_application.py
anish-neervannan: We use the first market_symbol_pair because if the trading pairs are different, such as WETH-DAI and ETH-USD, the tuples above will contain the information in terms of the first trading pair based on the way that the balances were computed in the for loop above.

anish-neervannan:
@chowryan made changes. Let me know if there are any other clarifications to make

hummingbot/client/hummingbot_application.py
```diff
@@ -1141,6 +1142,99 @@ def compare_balance_snapshots(self):
         lines = ["", "  Performance:"] + ["    " + line for line in str(df).split("\n")]
         self._notify("\n".join(lines))
 
+    def add_balances(self, market_symbol_pair, is_base, is_starting, base, quote):
+        """ Adds the balance of either the base or the quote in the given market symbol pair token to the corresponding
+        tuple pair.
+
+        :param market_symbol_pair: type market_symbol_pair
```
martinkou: The modern way to do type annotation in Python >= 3.5 is with the ':' notation inside the function arguments. You can still keep the text description here, but type annotation should be in the arguments list now.

hummingbot/client/hummingbot_application.py
```diff
@@ -1141,6 +1142,99 @@ def compare_balance_snapshots(self):
...
+
+        if is_base:
+            asset_tuple = base
+            asset = market_symbol_pair.base_asset
+        else:
+            asset_tuple = quote
+            asset = market_symbol_pair.quote_asset
+
+        if is_starting:
+            amount = self.starting_balances.get(asset).get(market_name)
```
martinkou: If you're sure the asset or market name exists in the dicts, you should just use `[]` instead of `get()`. `get()` should only be used when you're not sure and you want a sensible default value.

hummingbot/client/hummingbot_application.py
```diff
@@ -1141,6 +1142,99 @@ def compare_balance_snapshots(self):
...
+        # Each of these is in the format (currency, amount)
+        starting_base = tuple()
+        starting_quote = tuple()
+        current_base = tuple()
+        current_quote = tuple()
+
+        for market_symbol_pair in self.market_symbol_pairs:
+            for is_base in [True, False]:
+                starting_base, starting_quote = self.add_balances(
+                    market_symbol_pair=market_symbol_pair,
```
martinkou: You don't need to, and shouldn't use keywords if the argument isn't a keyword argument. If the function's argument name is changed later, your call will break.

hummingbot/client/hummingbot_application.py
```diff
@@ -1141,6 +1142,99 @@ def compare_balance_snapshots(self):
...
+                    quote=current_quote
+                )
+
+        # Compute the current exchange rate. We use the first market_symbol_pair because
+        # if the trading pairs are different, such as WETH-DAI and ETH-USD, the tuples
+        # above will contain the information in terms of the first trading pair based on
+        # the way that the balances were computed in the for loop above.
+        market_pair_info = self.market_symbol_pairs[0]
+        market = market_pair_info.market
+        buy_price = market.get_price(symbol=market_pair_info.trading_pair, is_buy=True)
```
martinkou: Again, don't add unnecessary keywords. It actually makes the code more fragile than without.

hummingbot/client/hummingbot_application.py
martinkou: Also.. the type annotation for market_symbol_pair is invalid.

hummingbot/client/hummingbot_application.py
martinkou: But wouldn't you miss the changes in the ETH-USD pair?

hummingbot/client/hummingbot_application.py
```diff
@@ -1141,6 +1142,99 @@ def compare_balance_snapshots(self):
...
+        return base, quote
+
+    def analyze_performance(self):
+        """ Determine the profitability of the trading bot. """
+        if len(self.starting_balances) == 0:
+            self._notify("  Performance analysis is not available before bot starts")
+            return
+
+        # Each of these is in the format (currency, amount)
+        starting_base = tuple()
```
martinkou: The general design of these data structures feel too complicated for the task to me... also, is this performance analyzer applicable to all strategies?

anish-neervannan: I changed to using named tuples. Without some sort of organization in the data structures, I would have 8 individual variables for this task, which I felt was a bit much.

hummingbot/client/hummingbot_application.py
anish-neervannan: In this case, I would convert the values of ETH and USD into WETH and DAI to have one "combined" balance across both exchanges.

anish-neervannan:
Making more changes to refactoring

anish-neervannan:
@martinkou @chowryan I just finished refactoring - please take a look and let me know if there are any other suggestions

hummingbot/client/performance_analysis.py
```diff
@@ -0,0 +1,44 @@
...
+            return self._starting_quote
+        elif is_base and not is_starting:
+            return self._current_base
+        else:
+            return self._current_quote
+
+    def add_balances(self, asset_name: str, amount: float, is_base: bool, is_starting: bool):
+        """ Adds the balance of either the base or the quote in the given market symbol pair token to the corresponding
+        CurrencyAmount object. """
+        currency_amount = self._get_currency_amount_pair(is_base, is_starting)
```
martinkou: There's an underlying assumption here that says the base / quote tokens between different market pairs are equivalent, and so they can be added to the same container.

This should be clearly documented. Because it's not obvious upon a casual reading of the code.

hummingbot/client/performance_analysis.py
```diff
@@ -0,0 +1,44 @@
...
+        if currency_amount.get_token() is None:
+            currency_amount.set_token(asset_name)
+            currency_amount.set_amount(amount)
+        else:
+            if currency_amount.get_token() == asset_name:
+                currency_amount.add_amount(amount)
+            else:
+                erc = ExchangeRateConversion.get_instance()
+                temp_amount = erc.convert_token_value(amount, asset_name, currency_amount.get_token())
```
martinkou: This is also related to the equivalency assumption above - only that for cases like DAI / USDT we need this conversion function.

So, again, please document why this is needed and how it's related to the equivalency assumption.

hummingbot/client/performance_analysis.py
```diff
@@ -0,0 +1,44 @@
...
+                currency_amount.add_amount(temp_amount)
+
+    def compute_profitability(self, price: float) -> float:
+        """ Compute the profitability of the trading bot based on the starting and current prices"""
+        starting_amount = (self._starting_base.get_amount() * price) + self._starting_quote.get_amount()
+        current_amount = (self._current_base.get_amount() * price) + self._current_quote.get_amount()
+        percent = ((current_amount / starting_amount) - 1) * 100
```
chowryan: can starting_amount ever be 0? (divide by 0)

chowryan:
looks great! let me know when youre done with unit tests.

anish-neervannan:
@martinkou made changes that you suggested. @chowryan finished unit tests.

carlolm:
add to whitelist

carlolm:
ok to test

PR #420, opened 2019-06-21
feat / bounty submit trades

yzhang1994:

- Allow bounty users to submit trades to database


hummingbot/market/markets_recorder.py
```diff
@@ -211,6 +212,7 @@ def _did_fill_order(self,
                                                  price=evt.price,
                                                  amount=evt.amount,
                                                  trade_fee=TradeFee.to_json(evt.trade_fee))
+        LiquidityBounty.get_instance().did_receive_new_trade_records([trade_fill_record])
```
martinkou: This forms a circular dependency between `MarketsRecorder` and `LiquidityBounty`... since each is calling the other's functions. While it works for now, it's architecturally a little bit messy.

It'll be a little bit cleaner if you turn `MarketsRecorder` into a `PubSub` child class, and trigger an event on trade fills instead. And have `LiquidityBounty` listen for the event. This way, the dependency becomes one way, and you don't need much code change to get there.

yzhang1994: Haha I'm going to just remove it then

PR #441, opened 2019-06-26
fix / fix xemm profit calc:

mifeng:

These fixes the profitability calculation for cross-exchange market making. The profitability should not take into account top depth tolerance, since it's displayed to the user in the status output. Otherwise, the actual profitability will be overly inflated when the order book is think or if the user sets a high top depth tolerance, as here:
unknown

For use of profitability as a metric to decide whether or not place orders, there's a downstream check c_check_if_still_profitable that uses top depth tolerance.


mifeng:
In testing, the bot seems to respect `top_depth_tolerance` so I'm not sure why the test is failing. Will check with @martinkou.

carlolm:
test this please

PR #497, opened 2019-07-03
Fix/handle fees pure mm

veecos2:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:


hummingbot/strategy/pure_market_making/constant_size_sizing_delegate.pyx
```diff
@@ -36,20 +41,27 @@ cdef class ConstantSizeSizingDelegate(OrderSizingDelegate):
...
             bint has_active_bid = False
             bint has_active_ask = False
 
+        buy_fees = market.c_get_fee(market_info.base_currency, market_info.quote_currency,
+                                    OrderType.MARKET, TradeType.BUY,
+                                    bid_order_size, pricing_proposal.buy_order_prices[0])
+
         if market.name == "binance":
-            bid_order_size = market.c_quantize_order_amount(market_info.symbol, self.order_size, pricing_proposal.buy_order_prices[0])
+            bid_order_size = market.c_quantize_order_amount(market_info.symbol, (self.order_size * (1+float(buy_fees.percent))), pricing_proposal.buy_order_prices[0])
```
mifeng: This seems to adjust the order size for fees - I thought we are trying to not adjust the order size

veecos2: Removed Binance order adjustment as discussed

hummingbot/strategy/pure_market_making/constant_size_sizing_delegate.pyx
```diff
@@ -36,20 +41,27 @@ cdef class ConstantSizeSizingDelegate(OrderSizingDelegate):
...
+
         if market.name == "binance":
             bid_order_size = market.c_quantize_order_amount(market_info.symbol, self.order_size, pricing_proposal.buy_order_prices[0])
             ask_order_size = market.c_quantize_order_amount(market_info.symbol, self.order_size, pricing_proposal.sell_order_prices[0])
+            required_quote_asset_balance = pricing_proposal.buy_order_prices[0] * bid_order_size
 
         else:
             bid_order_size = market.c_quantize_order_amount(market_info.symbol, self.order_size)
             ask_order_size = market.c_quantize_order_amount(market_info.symbol, self.order_size)
+            required_quote_asset_balance = pricing_proposal.buy_order_prices[0] * (1+float(buy_fees.percent)) * bid_order_size
```
mifeng: I don't think this line is necessary. Other exchanges do not require order sizes to be adjusted for buy fees.

Binance was the only exchange that took fees out of the base asset when you buy it, so that's why we originally adjusted order sizes. We aren't taking that into account anymore, and that's what this PR does. But we don't need to adjust order sizes in other markets for fees, since they never did that in the first place.

hummingbot/strategy/pure_market_making/staggered_multiple_size_sizing_delegate.pyx
```diff
@@ -65,22 +70,40 @@ cdef class StaggeredMultipleSizeSizingDelegate(OrderSizingDelegate):
...
+                #For binance fees is calculated in base token, so need to adjust for that
+                buy_order_size = market.c_quantize_order_amount(market_info.symbol, current_order_size, pricing_proposal.buy_order_prices[idx])
+                #Check whether you have enough quote tokens
+                required_quote_asset_balance += ( float(buy_order_size) * float(pricing_proposal.buy_order_prices[idx]) )
+
             else:
-                current_order_size = market.c_quantize_order_amount(market_info.symbol, current_order_size)
+                buy_order_size = market.c_quantize_order_amount(market_info.symbol, current_order_size)
+                #For other exchanges, fees is calculated in quote tokens, so need to ensure you have enough for order + fees
+                required_quote_asset_balance += ( float(buy_order_size) * float(pricing_proposal.buy_order_prices[idx]) *(1+float(buy_fees.percent)) )
```
mifeng: same comment as above

mifeng: Just to make sure that this is clear, the issue is NOT that we need to adjust the order size to reflect fees, either on the base or the quote side. 

The issue is that because Binance takes fees out of the base asset amount when you execute a buy order, you end up with a mismatched amount than expected if you are doing cross-exchange or arbitrage trades. Originally, we incremented the buy order size for Binance orders to make the amounts line up. However, this creates bugs and is confusing, so we are not doing this anymore. Other exchanges do not need any adjustment.

hummingbot/strategy/pure_market_making/constant_size_sizing_delegate.pyx
veecos2: The PR is to address the edge case of having enough balance but not enough balance for fees, would case insufficient balance issues. For example in Coinbase, you might have enough USD for 2 ETH but since fees is taken out in USD, your order might not go through as exchange would only allow to buy 1.999 ETH. The balance calculation earlier did not include fees and now I have incorporated that. This does not affect order sizing.

hummingbot/strategy/pure_market_making/staggered_multiple_size_sizing_delegate.pyx
veecos2: Same comment as above.

hummingbot/strategy/pure_market_making/constant_size_sizing_delegate.pyx
mifeng: Got it - you're right that this is necessary for non-Binance exchanges.

veecos2:
This does not break any Binance Unit tests

PR #528, opened 2019-07-10
Feat/list trades sqlite

veecos2:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:


hummingbot/client/command/list_command.py
```diff
@@ -65,16 +67,35 @@ def list(self,  # type: HummingbotApplication
 
         elif obj == "trades":
             lines = []
-            if self.strategy is None:
-                self._notify("No strategy available, cannot show past trades.")
+            #To access the trades from Markets Recorder you need the file path and strategy name
+            if in_memory_config_map.get("strategy_file_path").value is None or \
+                    in_memory_config_map.get("strategy").value is None:
+                self._notify("Kindly Configure the bot first")
```
martinkou: nitpick

"Please configure Hummingbot first."

hummingbot/client/command/list_command.py
```diff
@@ -65,16 +67,35 @@ def list(self,  # type: HummingbotApplication
 
         elif obj == "trades":
             lines = []
-            if self.strategy is None:
-                self._notify("No strategy available, cannot show past trades.")
+            #To access the trades from Markets Recorder you need the file path and strategy name
```
martinkou: PEP8: block comments should have a space after the '#'.

hummingbot/client/command/list_command.py
```diff
@@ -65,16 +67,35 @@ def list(self,  # type: HummingbotApplication
...
+            if in_memory_config_map.get("strategy_file_path").value is None or \
+                    in_memory_config_map.get("strategy").value is None:
+                self._notify("Kindly Configure the bot first")
             else:
-                if len(self.strategy.trades) > 0:
-                    df = Trade.to_pandas(self.strategy.trades)
-                    df_lines = str(df).split("\n")
-                    lines.extend(["", "  Past trades:"] +
-                                 ["    " + line for line in df_lines])
+                markets_recorder = MarketsRecorder(
```
martinkou: You should be able to use `self.markets_recorder` here instead of building your own.

hummingbot/core/data_type/trade_fills.py
```diff
@@ -0,0 +1,57 @@
...
+    trade_type: TradeType
+    price: float
+    amount: float
+    order_type: OrderType
+    market: str
+    timestamp: float
+    trade_fee: TradeFee
+
+    @classmethod
+    def to_pandas(cls, trades: List):
```
martinkou: This should be added to the `TradeFill` model class. A new data type class doesn't seem needed for now.

hummingbot/client/command/list_command.py
veecos2: Already did that, that does not work unless a strategy is running

hummingbot/model/trade_fill.py
```diff
@@ -1,8 +1,14 @@
 #!/usr/bin/env python
-import numpy
+import numpy, pandas as pd
```
martinkou: Importing two different modules in one line is disallowed by PEP8.

hummingbot/client/command/list_command.py
```diff
@@ -65,16 +67,33 @@ def list(self,  # type: HummingbotApplication
...
+                config_file = in_memory_config_map.get("strategy_file_path").value
+                queried_trades = markets_recorder.get_trades_for_config(config_file)
+                df = TradeFill.to_pandas(queried_trades)
+                if len(df) > 0:
+                    if len(df) <MAXIMUM_TRADE_FILLS_DISPLAY_OUTPUT:
+                        df_lines = str(df).split("\n")
+                        lines.extend(["", "  Past trades:"] +
+                                     ["    " + line for line in df_lines])
+                    else:
+                        df_lines = str(df[:MAXIMUM_TRADE_FILLS_DISPLAY_OUTPUT]).split("\n")
```
martinkou: If it's possible for the list of trades to go beyond the screen's capacity, it would be more useful to list from the newest one first rather than the oldest.

I think we should change the `.get_trades_for_config()` function a bit here:

 1. The results should be sorted in descending order of timestamp, rather than ascending order we have right now.
 2. There should be an optional parameter in the function to allow limiting the number of rows returned.

Once you have these, you can then query for n+1 rows, n being the maximum number of rows to display, and then check whether you actually got >n rows. If you have >n then you print the notify message about having more rows below.

martinkou:
@yzhang1994 we need your approval before this patch can be merged.

PR #541, opened 2019-07-12
Fix/Fix_lost_orders

veecos2:

**Optional**:
- Related Github issue:
https://github.com/CoinAlpha/hummingbot/issues/493
- Related Clubhouse Story:
https://app.clubhouse.io/coinalpha/story/4828/example-of-a-lost-order-on-binance
https://app.clubhouse.io/coinalpha/story/4747/duplicate-order-bug-pure-mm-multi

The pure MM strategy is supposed to cancel the orders only after receiving the Order cancellation event from the Market class. Due to a bug in the logic, the strategy was not accounting these orders as active even when the Order cancellation event was not received from the market class.

This will return the order as active until a successful cancel event has occurred. This should hopefully fix orders being lost.


mifeng:
I've run this branch with 8 orders refreshed every 20 secs for the past few hours and have not yet seen any lost orders.

PR #638, opened 2019-08-08
fix / flat fees key error

yzhang1994:

**Optional**:
- Related Github issue: https://github.com/CoinAlpha/hummingbot/issues/636


mifeng:
I'm testing this now and will try to generate some trades on Radar Relay to make sure that the history command is working properly.

mifeng:
Actually, I just noticed that `history` for Radar Relay trades shows **None** under flat fees / gas. @yzhang1994 is this correct?

yzhang1994:
@mifeng it is. The fee is 0 for limit orders. You only pay gas if you submit market orders.

PR #680, opened 2019-08-16
fix / reduce web3 usage for infura

chowryan:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:


hummingbot/wallet/ethereum/web3_wallet_backend.py
```diff
@@ -296,11 +303,12 @@ def stop(self):
...
+        # Assume connected if received new blocks in last 2 minutes
+        if time.time() - self._last_timestamp_received_blocks > 60 * 2:
+            return NetworkStatus.CONNECTED
+
         try:
-            await self._ev_loop.run_in_executor(hummingbot.get_executor(),
-                                                getattr,
-                                                self._w3.eth,
-                                                "blockNumber")
+            await self._update_gas_price()
```
martinkou: This is not needed - we're already updating gas prices at new block.

hummingbot/wallet/ethereum/web3_wallet_backend.py
```diff
@@ -296,11 +303,12 @@ def stop(self):
             self._check_transaction_receipts_task = None
 
     async def check_network(self) -> NetworkStatus:
+        # Assume connected if received new blocks in last 2 minutes
+        if time.time() - self._last_timestamp_received_blocks > 60 * 2:
```
martinkou: This is wrong... while block times average 15 seconds per block, it's not rare for some blocks time to be more than 2 minutes. Block times are expected to form a Gaussian random distribution, rather than being perfectly regular.

You should use the last successful block check's timestamp instead, not the new block's timestamp.

hummingbot/wallet/ethereum/web3_wallet_backend.py
chowryan: yea it's not needed but might as well do a task that is useful? rather than just checking blockNumber and ignoring result.

hummingbot/wallet/ethereum/web3_wallet_backend.py
chowryan: this is just a quick check, if we received a block within the last 2 min, assume our connection is still good, if not ping for gas price just to verify that connection to web3 provider is still good.

martinkou: ok, I think I get it then. It wasn't clear when looking at just the diffs.

gabririgo:
this is a much-needed efficiency improvement. I understand this:
https://github.com/CoinAlpha/hummingbot/blob/ee17287b9e892c03a09f5dc07bdb17fb7b319c65/hummingbot/wallet/ethereum/watcher/new_blocks_watcher.py#L140
probably reduces the overall number of calls by [60% 70%].

I also see that you are reducing the number of calls by setting a default gas price and some other optimizations.

One suggestion I have to further reducing the number of calls when updating balances is to use a smart contract helper instead of a loop of rpc calls. Since each cross is composed by 2 tokens, you can save 50% number of calls to the endpoint. Here:
https://github.com/CoinAlpha/hummingbot/blob/ee17287b9e892c03a09f5dc07bdb17fb7b319c65/hummingbot/wallet/ethereum/watcher/account_balance_watcher.py#L112

Since you also query for the ETH balance, you can also add that to the return values from the helper contract. This would mean reducing the calls from 3 to 1.

Contract Specification

Function:
- function queryMultipleBalances

Inputs:
- address[] token
- address account

Return values:
- address[] tokens uint256[] tokenBalances
- uint256 etherBalance

Function specification
- loop to call Token(token).balanceOf(account)
- account.balance

Furthermore, you could also decide to query the allowances and return them, therefore further reducing the calls. An example of how I implemented one such helper in solidity can be found here:
https://github.com/RigoBlock/rigoblock-monorepo/blob/master/packages/contracts/src/examples/helpers/HGetMultipleBalances.sol

PR #693, opened 2019-08-23
feat / Implement Bittrex connector

CrimsonJacket:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:
https://app.clubhouse.io/coinalpha/story/4547/implement-bittrex-connector

Added Bittrex exchange connector into Hummingbot client


carlolm:
add go whitelist

carlolm:
add to whitelist

hummingbot/market/bittrex/bittrex_auth.py
```diff
@@ -0,0 +1,63 @@
...
+        :param params: Dictionary of url parameters to be included in the api request. USED ONLY IN SOME CASES
+        :param body: Dictionary representing the values in a request body.
+        :param subaccount_id: String value of subaccount id.
+        :return: Dictionary containing the final 'params' and its corresponding 'signature'.
+        """
+
+        # Appends params the url
+        def append_params_to_url(url: str, params: Dict[str, any] = {}) -> str:
+            if params:
+                param_str = "&".join([f"{param}={value}" for (param, value) in params.items()])
```
martinkou: It's dangerous to generate the query string this way - because it doesn't account for escape characters in urlencode format. You should use [`urllib.parse.urlencode()`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode) instead.

hummingbot/market/bittrex/bittrex_auth.py
```diff
@@ -0,0 +1,63 @@
...
+                return f"{url}?{param_str}"
+            return url
+
+        def construct_content_hash(body: Dict[str, any] = {}) -> str:
+            if body:
+                return hashlib.sha512(ujson.dumps(body).encode()).hexdigest()
```
martinkou: You should preferably only do the JSON encode() once - and make sure the JSON you send out and the the JSON you signed are exactly the same. While doing it twice "should" return the same result, it's both more risky and performs unnecessary computations.

CrimsonJacket: The reason for performing `encode()` here is because its must be encoded before hashing. If not, `hashlib` would throw a `TypeError`

hummingbot/market/bittrex/bittrex_in_flight_order.pyx
```diff
@@ -0,0 +1,67 @@
...
+            initial_state
+        )
+
+    @property
+    def is_done(self) -> bool:
+        return self.last_state in {"CLOSED"}
+
+    @property
+    def is_failure(self) -> bool:
+        return NotImplementedError
```
martinkou: `is_failure` should be implemented and include cancellation. In your case, it likely means the same as `is_cancelled`

See: https://docs.hummingbot.io/developers/connectors/order-lifecycle/

hummingbot/market/bittrex/bittrex_in_flight_order.pyx
```diff
@@ -0,0 +1,67 @@
...
+
+    @property
+    def is_cancelled(self) -> bool:
+        return self.last_state in {"CLOSED"}
```
martinkou: This looks concerning... as it means we can't distinguish between cancelled and completed.

hummingbot/market/bittrex/bittrex_active_order_tracker.pyx
```diff
@@ -0,0 +1,163 @@
+# distutils: language=c++
```
martinkou: This whole file may be unneeded... if the diff messages we receive from Bittrex is already in (price, quantity) format, then you can simply use `apply_diff()` to update the order book. You can refer to the Binance market connector for how to do this - Binance doesn't have an active order tracker, it simply does `apply_diff()` in the order book tracker.

hummingbot/market/bittrex/bittrex_api_order_book_data_source.py
```diff
@@ -0,0 +1,394 @@
...
+    async def _socket_stream(self, conn: signalr_aio.Connection) -> AsyncIterable[str]:
+        try:
+            while True:
+                async with timeout(MESSAGE_TIMEOUT):
+                    yield await conn.msg_queue.get()
+        except asyncio.TimeoutError:
+            self.logger().warning("Message recv() timed out. Going to reconnect...")
+            return
+
+    async def _transform_raw_message(self, msg) -> Dict[str, Any]:
```
martinkou: This function is repeated a few times in the data source code. You should refactor it out.

hummingbot/market/bittrex/bittrex_api_order_book_data_source.py
```diff
@@ -0,0 +1,394 @@
...
+            except Exception:
+                self.logger().error("Unexpected error.", exc_info=True)
+            finally:
+                connection.close()
+
+    async def listen_for_order_book_diffs(self, ev_loop: asyncio.BaseEventLoop, output: asyncio.Queue):
+        while True:
+            connection: Optional[signalr_aio.Connection] = None
+            try:
+                connection = signalr_aio.Connection(BITTREX_WS_FEED, session=None)
```
martinkou: As discussed earlier today... there's likely no need to create separate connections for snapshot and diffs. You can actually share the same connection and manage its liveliness via the empty ping messages from the exchange.

hummingbot/market/bittrex/bittrex_auth.py
martinkou: No, I mean you can do it like this..

```
def construct_content_hash(body: bytes) -> str:
   ...

...
json_bytes: bytes = ujson.dumps(body).encode()
content_hash = construct_content_hash(json_bytes)
...
return {"headers": headers, "body": json_bytes, "url": url}
```

The encode I mentioned meant converting from Dict to JSON bytes. If you only do the conversion once, then you can save

martinkou:
I think this patch is pretty much done. Just need to resolve the merge conflict and the `asyncio.gather()` comment above and it should be ready to merge.

PR #695, opened 2019-08-24
Feat/xemm place orders based on min profitablity

veecos2:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:
https://app.clubhouse.io/coinalpha/story/5132/change-xemm-to-always-place-orders-deep-in-the-order-book

Basically had to rewrite the entire XEMM strategy.
- Removed all the profitability calculations as it will always place an order in the maker order book based on the price
- Added vwap price for taker order book and profitability calculations
- Removed unusused params


veecos2:
Tested it working on DDEX, but needs further testing.

carlolm:
@martinkou @mifeng @veecos2 we should discuss this.
I think the simplest strategy would be if I can hedge on the taker exchange at 100 and I have a 0.5% min profitability, it should be pretty simple; I place my bid at 99.5.

I wouldn't care that the previous best bid is 99 (so no need to adjust my price to 99.1), or care if there's already a better bid at 99.6.  I'd still just simply want to place my bid at 99.5.

veecos2:
Addressed Martin's comments. Tested it working with 1. make Binance ONE/PAX, take Binance ONE/USDT 2. make ddex ZRX/WETH, take Binance ZRX/ETH. @mifeng: The Bamboo relay connector is broken and needs to be fixed.

veecos2:
@carlolm : I basically rewrote many parts of the strategy. Now the price for which it places the orders is based on the VWAP of executing a market order on the taker side and added the min profitability to it. So the price on maker is only dependent now on how much you can hedge it for. 

I can add a parameter which will make it so that the price is the expected price and not the top bid/ask, which would be more profitable.

carlolm:
thanks @veecos2 , I think that making adjusting price to jump best bid/ask optional that can be toggled would be great.  I'd initial users of the strategy would find it more intuitive, but also think that does serve the mirroring use case more and improving pricing on the maker exchange / actually making it more in line with the taker exchange.

mifeng:
@carlolm @veecos2 Reading this, I'm not sure you guys are talking about the same thing:

* The hedging price should always be based on VWAP on the taker book, not the top bid/ask on the taker book. Top bid/ask doesn't take into account size, so it's always a less accurate number than VWAP.
* Making top depth tolerance (penny jumping logic) optional is something that we should eventually do, but it is outside the scope of what this PR is intended to do. I think it's fine if it's a simple change, but if it's going to push this PR into next release, I don't think it's a good idea.

veecos2:
@mifeng : So earlier it was not based on vwap, the taker price would be based off the worst price at which you could buy/sell. Now I have changed it to vwap. 

Carlo is proposing two things here,
1. make the maker price as independent of  top bid/ask. Basically the maker price is only based on the taker price and min profitability.
2. being independent of top bid/ask would make it be not affected by penny jumping.

I believe this can be done for the upcoming release.

martinkou:
I can see some improvements compared to last week's patch. e.g. `c_get_market_making_price()` is now able to compare the top of order book prices and the min profitability prices and take the wider one - instead of always making offers at min profitability.

There're still a few design issues as of today:

 1. As discussed last week, we want to keep the top depth tolerance logic because only looking at the top of order book prices would make the strategy vulnerable to dust orders. This is missing as of today.

 2. The price calculation logic of `c_get_market_making_price()` and `c_check_if_price_has_drifted()` are inconsistent. In particular, `c_check_if_price_has_drifted()` uses the price samples to determine the effective top of order book price, while `c_get_market_making_price()` only uses the current top of order book price. This can cause orders to get cancelled frequently because `c_check_if_price_has_drifted()` may constantly detect new orders has having drifted, if other market makers have been constantly adding or deleting orders near the top of the book.

 3. The logic in `c_check_if_price_has_drifted()` is not checking the price of the active limit orders, but the output of `c_get_market_making_price()` - this is definitely wrong.

I'd recommend you add back top depth tolerance in calculating the effective top of order book prices - but decrease the default value of top depth tolerance by 10x+, just as discussed last week. Also, always account for previous price samples in calculating the top of order book prices as well. This should make it similar, or the same, to pre-patch logic for top-of-order book price... basically the effective top-of-order-book price has three layers: current order book prices -> top depth tolerance -> recent price samples. You always use the effective top-of-order book price after the three layers when checking for price drifts and also making new orders.

The new logic of comparing always making order at min profitability, is the fourth layer of pricing logic on top of the original three layers.

veecos2:
As discussed with Martin, we will use the top depth tolerance (in base asset) to calculate top bid and top ask prices, store this in price samples and take from the price samples for calculating top bid and ask across the strategy.

hummingbot/strategy/cross_exchange_market_making/cross_exchange_market_making.pyx
```diff
@@ -880,137 +797,143 @@ cdef class CrossExchangeMarketMakingStrategy(StrategyBase):
...
+                taker_price *= self._exchange_rate_conversion.convert_token_value(1,
+                                                                                  market_pair.taker.quote_asset,
+                                                                                  market_pair.maker.quote_asset)
+
+            # you are buying on the maker market and selling on the taker market
+            maker_price = taker_price / (1 + self._min_profitability)
+
+            if self._adjust_orders_enabled:
+                # If your bid is higher than highest bid price, reduce it to one tick above the top bid price
+                if maker_price > price_above_bid:
```
martinkou: `maker_price = min(maker_price, price_above_bid)` would be simpler and easier to read.

hummingbot/strategy/cross_exchange_market_making/cross_exchange_market_making.pyx
```diff
@@ -880,137 +797,143 @@ cdef class CrossExchangeMarketMakingStrategy(StrategyBase):
...
+            # You are buying on the taker market and selling on the maker market
+            maker_price = taker_price * (1 + self._min_profitability)
 
-            raw_size_limit = min(
-                taker_order_book_size_limit,
-                maker_balance_size_limit,
-                taker_balance_size_limit
+            if self._adjust_orders_enabled:
+                # If your ask is lower than the the top ask, increase it to just one tick below top ask
+                if maker_price < next_price_below_top_ask:
```
martinkou: Similarly, `maker_price = max(maker_price, next_price_below_top_ask)` would be better here.

hummingbot/strategy/cross_exchange_market_making/cross_exchange_market_making.pyx
```diff
@@ -647,230 +707,87 @@ cdef class CrossExchangeMarketMakingStrategy(StrategyBase):
...
-        there's not enough balance for the maker order or the hedging trade; or if it's not possible to hedge the
-        trade profitably, then the returned order size will be 0.
+        The price returned is calculated by adding the profitability to vwap of hedging it on the taker side. 
+        or if it's not possible to hedge the trade profitably, then the returned order price will be none.
 
         :param market_pair: The cross exchange market pair to calculate order price/size limits.
         :param is_bid: Whether the order to make will be bid or ask.
-        :param own_order_depth: Market depth caused by existing order issued by ourselves.
-        :return: a (Decimal, Decimal) tuple. First item is price, second item is size limit.
+        :param size: size of hte order.
```
martinkou: typo

hummingbot/strategy/cross_exchange_market_making/cross_exchange_market_making.pyx
```diff
@@ -647,230 +707,87 @@ cdef class CrossExchangeMarketMakingStrategy(StrategyBase):
...
+        :return: a Decimal which is the price
```
martinkou: add ", or None if the order cannot be hedged on the taker market"

veecos2:
@chowryan : What is linter ?. How do you pass it ?.

veecos2:
All necessary changes addressed.

chowryan:
@veecos2 run install again and it should add flake8. itll warn you where your code doesnt conform to the style guide

hummingbot/strategy/cross_exchange_market_making/cross_exchange_market_making.pyx
```diff
@@ -64,29 +66,32 @@ cdef class CrossExchangeMarketMakingStrategy(StrategyBase):
...
                  min_profitability: float,
-                 trade_size_override: Optional[float] = 0.0,
+                 order_amount: Optional[float] = 0.0,
                  order_size_taker_volume_factor: float = 0.25,
                  order_size_taker_balance_factor: float = 0.995,
                  order_size_portfolio_ratio_limit: float = 0.1667,
                  limit_order_min_expiration: float = 130.0,
-                 cancel_order_threshold: float = -1,
-                 active_order_canceling: bint = True,
+                 adjust_order_enabled: bool = True,
```
martinkou: There's no documentation for what the `adjust_order_enabled` flag does, and it's still not clear to me after reviewing the code.

veecos2:
@chowryan : Already did that

hummingbot/strategy/cross_exchange_market_making/cross_exchange_market_making.pyx
```diff
@@ -1023,51 +926,89 @@ cdef class CrossExchangeMarketMakingStrategy(StrategyBase):
...
+
+        if self._top_depth_tolerance == 0:
+
+            top_bid_price = maker_order_book.c_get_price(False)
+
+            top_ask_price = maker_order_book.c_get_price(True)
+
+        else:
+            # Use bid entries in maker order book
+            top_bid_price = maker_order_book.c_get_vwap_for_volume(False, self._top_depth_tolerance)
```
martinkou: The top depth tolerance setting means the order book depth to ignore, not average out. So getting the vwap here is wrong. It should be `c_get_price_for_volume()` instead for this.

hummingbot/strategy/cross_exchange_market_making/cross_exchange_market_making.pyx
```diff
@@ -519,6 +498,78 @@ cdef class CrossExchangeMarketMakingStrategy(StrategyBase):
...
+
+        suggested_price = float(self.c_get_market_making_price(market_pair, is_buy, order_quantity))
+
+        top_bid_price, top_ask_price = self.c_get_top_bid_ask_from_price_samples(market_pair)
+
+        if is_buy:
+
+
+            if self._adjust_orders_enabled:
+                adjusting_condition = suggested_price > order_price and suggested_price > top_bid_price
```
martinkou: This only checks for the case where the market moved narrower, but it won't update the price if the market moved wider, to get more potential profit.

I think the `adjust_orders_enabled` flag is not really needed here. But what's needed is that you should do a `suggested_price != order_price` check here instead (make sure they're properly quantized). If they're not equal, then adjust the order.

hummingbot/strategy/cross_exchange_market_making/cross_exchange_market_making.pyx
veecos2: Refer to carlo's comments. This is the feature which he requested, that does not change based on the top bid / ask. I will add the doc.

hummingbot/strategy/cross_exchange_market_making/cross_exchange_market_making.pyx
veecos2: if the market moves the other way, then it will not be profitable and it will get cancelled. its already handled.

martinkou: No. If `suggested_price` is lower than order price, it means you can adjust the order to make more profit.

martinkou: Or put it another way...

The suggested price can become lower because the top of the order book widened, because `c_get_market_making_price()` is using both the effective top of order book price and the taker side hedge price.

Right now, this logic wouldn't adjust the bid pricing in this case, to capture more potential profits.

veecos2: will it not make the bot keep on cancelling and placing orders again and again and make the bot overall unstable ?. I think we should not chase the top order. If someone places above us, then you can chase it.

veecos2: The suggested price is linked to the taker order book as well as the top bid, ask prices. I think making this change, would make it keep on cancelling and replacing the bid and ask prices.

martinkou: That's why we have the anti hysteresis timer. It will only do that every 60 seconds.

martinkou:
Approved, but we don't have enough time to test this for next Monday's release. So I'm marking this as DO NOT MERGE for now.

documentation/docs/strategies/cross-exchange-market-making.md
```diff
@@ -32,6 +32,16 @@ The end result: Hummingbot has sold the same asset at \$102 (❶) and purchased
...
+### Adjusting Orders
+
+The maker price is calculated based on the price at which you can buy or sell on the taker exchange with a min_profitability added to it.
+
+For example if you can sell 1 ETH on taker exchange for 100 USDT on a volume weighted average price basis, then if min_profitability is 0.05, then maker price at which limit buy will be placed is 95 USDT.
+
+Now its possible that the top bid on the exchange is much lower, say 90 USDT. In this case if you enable adjust orders it will place at 90.01 (one price level above top bid).
+
+If you disable adjust orders, it will place order at 95 USDT instead of 90.01 USDT. 
+
```
chowryan: "adjust orders" seems too generic. Can we come up with a better term?

documentation/docs/strategies/cross-exchange-market-making.md
```diff
@@ -32,6 +32,16 @@ The end result: Hummingbot has sold the same asset at \$102 (❶) and purchased
 
 Initially, we assume that the maker exchange is an Ethereum-based decentralized exchange and that the taker exchange is Binance.
 
+### Adjusting Orders
+
+The maker price is calculated based on the price at which you can buy or sell on the taker exchange with a min_profitability added to it.
+
+For example if you can sell 1 ETH on taker exchange for 100 USDT on a volume weighted average price basis, then if min_profitability is 0.05, then maker price at which limit buy will be placed is 95 USDT.
```
chowryan: confusing syntax.
suggestion -
Example: You can currently sell 1 ETH on the taker exchange for 100 USDT on a volume weighted average price basis. If your 'min_profitability' is 0.05, hummingbot will place a buy limit order at 95 USDT on the maker exchange.

documentation/docs/strategies/cross-exchange-market-making.md
```diff
@@ -32,6 +32,16 @@ The end result: Hummingbot has sold the same asset at \$102 (❶) and purchased
...
+
+Now its possible that the top bid on the exchange is much lower, say 90 USDT. In this case if you enable adjust orders it will place at 90.01 (one price level above top bid).
```
chowryan: *it's*

what is a "price level"?

documentation/docs/strategies/cross-exchange-market-making.md
veecos2: what name do you suggest ?

documentation/docs/strategies/cross-exchange-market-making.md
```diff
@@ -48,9 +58,12 @@ The following walks through all the steps when running `config` for the first ti
...
 | `Enter the token symbol you would like to trade on [taker exchange name] >>>`: | Enter the token symbol for the *taker exchange*.<br/>Example input: `ZRX-ETH`<br/><table><tbody><tr><td bgcolor="#ecf3ff">**Note**: ensure that (1) the pair corresponds with the token symbol entered for the maker exchange, and that (2) is a valid pair for the exchange.  *Note in the example, the use of `ETH` instead of `WETH`*.</td></tr></tbody></table>|
 | `What is the minimum profitability for your to make a trade? (Enter 0.01 to indicate 1%) >>>`: | This sets `min_profitability` (see [definition](/strategies/cross-exchange-market-making/#configuration-parameters)). |
+| `Do you want to adjust the prices to be above top bid/ask instead of the expected price, if profitable ? (Default is True) >>>`: | This sets `adjust_order_enabled` (see [definition](/strategies/cross-exchange-market-making/#configuration-parameters)). |
```
chowryan: confusing, what is expected price? what does it mean by "if profitable"? will hummingbot place orders if it's not profitable?

documentation/docs/strategies/cross-exchange-market-making.md
veecos2: https://www.grammarly.com/blog/its-vs-its/

documentation/docs/strategies/cross-exchange-market-making.md
veecos2: I think it makes sense.

documentation/docs/strategies/cross-exchange-market-making.md
```diff
@@ -48,9 +58,12 @@ The following walks through all the steps when running `config` for the first ti
...
 | `Do you want to actively adjust/cancel orders (Default True) >>>`: | This sets `active_order_canceling` (see [definition](/strategies/cross-exchange-market-making/#configuration-parameters)). |
 | `What is the minimum profitability to actively cancel orders? (Default to 0.0, only specify when active_order_cancelling is disabled, value can be negative) >>>`: | This sets the `cancel_order_threshold` (see [definition](/strategies/cross-exchange-market-making/#configuration-parameters)). |
+| `"What is your preferred trade size? (denominated in " "the base asset) >>>`: | This sets the `order_amount` (see [definition](/strategies/cross-exchange-market-making/#configuration-parameters)). |
```
chowryan: are the double quotes in the wrong place?

documentation/docs/strategies/cross-exchange-market-making.md
veecos2: If its profitable to place above top bid/ask it will place it there. How do you want to change this ?

documentation/docs/strategies/cross-exchange-market-making.md
veecos2: https://en.wikipedia.org/wiki/Order_book_(trading)#Price_levels

documentation/docs/strategies/cross-exchange-market-making.md
chowryan: no blank space before `?`

documentation/docs/strategies/cross-exchange-market-making.md
```diff
@@ -62,8 +75,9 @@ The following parameters are fields in Hummingbot configuration files (located i
...
-| **top_depth_tolerance** | An amount expressed in quote currency of maximum aggregate amount of orders at a better price than Hummingbot's order that are allowed before Hummingbot modifies its order.<br/><br/>*Example: assuming a top depth tolerance of `100` and a Hummingbot bid price of 20, if there exist an aggregate amount of orders of more than 100 at a better (higher) price than Hummingbot's bid, Hummingbot will cancel its order at 20 and re-evaluate the next opportunity to place a new bid.*
-| **trade_size_override** | An amount expressed in quote currency of maximum allowable order size.  If not set, the default value is 1/6 of the aggregate value of quote and base currency balances across the maker and taker exchanges.<br/><br/>*Example: assuming a trade size override of `100` and a token symbol of ETH/DAI, the maximum allowable order size is one that has a value of 100 DAI.*
+| **top_depth_tolerance** | An amount in base currency, which is the specifies how much deep you want to go to calculate the top bid and ask. This is used for getting the top bid and ask, ignoring dust orders on top of the order book. _Default value: 0_.
```
chowryan: this definition is unclear. why did we modify the original?

chowryan: this also needs an example

hummingbot/core/data_type/order_book.pyx
```diff
@@ -328,6 +328,33 @@ cdef class OrderBook(PubSub):
...
+            double total_cost  = 0
+            double total_volume = 0
+            double result_vwap = NaN
+
+        if is_buy:
+            for order_book_row in self.ask_entries():
+                total_cost += order_book_row.amount * order_book_row.price
+                total_volume += order_book_row.amount
+                if total_cost >= quote_volume:
+                    incremental_base_asset = (total_cost-quote_volume)/order_book_row.price
```
chowryan: are you sure you have flake8 setup? it should warn for no spaces between operator.

our standard is to have `a / b` not `a/b`

hummingbot/strategy/cross_exchange_market_making/cross_exchange_market_making.pxd
```diff
@@ -22,44 +23,59 @@ cdef class CrossExchangeMarketMakingStrategy(StrategyBase):
...
+                                                object market_pair)
+    cdef object c_get_market_making_size(self,
+                                        object market_pair,
+                                        bint is_bid)
+    cdef object c_get_market_making_price(self,
+                                         object market_pair,
+                                         bint is_bid,
+                                         double size)
+    cdef object c_calculate_effective_hedging_price(self,
+                                         object market_pair,
```
chowryan: line up all the variables properly please

hummingbot/strategy/cross_exchange_market_making/start.py
```diff
@@ -1,43 +1,33 @@
-import re
-from typing import (
-    List,
-    Tuple,
-)
+from typing import List, Tuple
```
chowryan: wrong syntax

hummingbot/strategy/cross_exchange_market_making/start.py
```diff
@@ -1,43 +1,33 @@
...
 
 from hummingbot.client.config.global_config_map import global_config_map
 from hummingbot.strategy.market_symbol_pair import MarketSymbolPair
 from hummingbot.strategy.cross_exchange_market_making.cross_exchange_market_pair import CrossExchangeMarketPair
 from hummingbot.strategy.cross_exchange_market_making.cross_exchange_market_making import (
-    CrossExchangeMarketMakingStrategy
+    CrossExchangeMarketMakingStrategy,
```
chowryan: single line for single imports

hummingbot/strategy/cross_exchange_market_making/start.py
```diff
@@ -1,43 +1,33 @@
...
 )
 from hummingbot.strategy.cross_exchange_market_making.cross_exchange_market_making_config_map import (
-    cross_exchange_market_making_config_map
+    cross_exchange_market_making_config_map,
```
chowryan: single line for single imports

hummingbot/strategy/cross_exchange_market_making/start.py
```diff
@@ -47,28 +37,33 @@ def start(self: "HummingbotApplication"):
         return
     self._initialize_wallet(token_symbols=list(set(maker_assets + taker_assets)))
     self._initialize_markets(market_names)
+    self.logger().info("wallet works")
```
chowryan: do we need this??

documentation/docs/strategies/cross-exchange-market-making.md
chowryan: "Now it is possible".

documentation/docs/strategies/cross-exchange-market-making.md
chowryan: chaining multiple "then"s is awkward.

documentation/docs/strategies/cross-exchange-market-making.md
chowryan: we need to explain the term on our website.

veecos2: both are correct

hummingbot/core/data_type/order_book.pyx
veecos2: Look at my comment before, I already did ./install. I will change this.

documentation/docs/strategies/cross-exchange-market-making.md
veecos2: It is clear. The previous definition does not make sense to me. I will add example.

chowryan: read the sentence again. youre missing words.

veecos2:
@chowryan: It is clear to me. It is very similar to the old xemm, so i'm not sure where the confusion stems from.

documentation/docs/strategies/cross-exchange-market-making.md
chowryan: see google - 

Screen Shot 2019-09-09 at 4 57 12 PM

chowryan: dont mark stuff as resolved if its not.

veecos2: I thought it was resolved. Ok, you want to be a grammar nazi I guess.

veecos2: Grammar police

documentation/docs/strategies/cross-exchange-market-making.md
martinkou: Vignesh changed the semantics from quote currency to base currency. But agreed the new doc is harder to understand and has grammatical errors (e.g. "which is the specifies ..."). The old doc is easier to understand in my opinion. We should probably just reuse that but changing the "quote currency" to "base asset".

martinkou:
@veecos2 the flake8 pre-commit hook would only check changed files by default. I think you're not seeing the warnings because you're not changing all the files in the new commits. You can manually call `flake8` and check the cross exchange market making strategy files with 

`$ flake8 hummingbot/strategy/cross_exchange_market_making/`

veecos2:
@martinkou: I already updated the code and re-requested the review.

veecos2:
@mifeng: Removed the prompt questions.

martinkou:
@chowryan please take a look at Vignesh's updates. The PR is now blocked by you.

documentation/docs/strategies/cross-exchange-market-making.md
```diff
@@ -62,8 +75,13 @@ The following parameters are fields in Hummingbot configuration files (located i
...
+| **order_amount** | An amount expressed in base currency of maximum allowable order size.  If not set, the default value is 1/6 of the aggregate value of quote and base currency balances across the maker and taker exchanges as determined by the order_size_portfolio_ratio_limit.<br/><br/>*Example: assuming an order amount of `1` and a token symbol of ETH/DAI, the maximum allowable order size is one that has a value of 1 ETH.*
+| **adjust_order_enabled** | `True` or `False`<br/>If enabled (parameter set to `True`), the strategy will place the order on top of the top bid and ask if its more profitable to place it there. If this is set to `False`, strategy will ignore the top of the maker order book for price calculations and only place the order based on taker price and min_profitability. Refer to Adjusting orders section above. _Default value: True_
```
chowryan: still wrong `it's`

PR #722, opened 2019-09-02
docker install and update scripts with ethereum local node

gabririgo:
- added 2 scripts which install a hummingbot docker instance pointing to a local node running on the local machine (outside of docker containers).
- updated docker-commands/README.md to add the new scripts (names ending -web3).

the documentation regarding the setup using a local node is not detailed. It might require a bit of troubleshooting to be able to running hummingbot pointing to a local node. Therefore, I created two additional script files and updated the README.md for the docker setup in local node mode, which will save a bit of time to devs/market markers.


carlolm:
OK to test

PR #799, opened 2019-09-18
Feat/puremm adjust other order on fill

veecos2:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:
https://app.clubhouse.io/coinalpha/story/5076/pure-mm-update-both-bid-ask-orders-when-a-trade-on-either-side-is-filled
https://app.clubhouse.io/coinalpha/story/4492/binary-request-adjust-orders-upon-fill
https://app.clubhouse.io/coinalpha/story/5688/filter-failure-percent-price-when-using-multiple-orders-in-puremm

1. Added the filled order delay
a. Created a filter delegate which will only allow orders to be created after a certain timestamp.
b. Whenever orders are filled, this filter timestamp is set, so that new orders are not immediately created. This only affects order creation and not order cancellation.
2. Added stop order cancellation of orders on the other side for order filled events.
3. Added documentation for the same.
4. Refactored all pure MM files to PEP8 standards.
5. Added unit tests for the features.
6. Added validation of percentage inputs for pure MM config file where input is required to be a percentage.


veecos2:
Tested it working on Binance on ETH/USDT.

hummingbot/strategy/pure_market_making/pass_through_filter_delegate.pyx
```diff
@@ -1,7 +1,21 @@
...
+                         OrdersProposal)
 
 cdef class PassThroughFilterDelegate(OrderFilterDelegate):
+
+    def __init__(self, order_placing_timestamp: float):
+        super().__init__()
+        self._order_placing_timestamp = order_placing_timestamp
+
+    @property
+    def order_placing_timestamp(self) -> float:
```
martinkou: This interface is not specified in the filter delegate. Filter delegates are intended to be plugins. If you specify something in an implementation class and make the owner of the delegate (i.e. PureMarketMakingStrategyV2) use that... then you've effectively required all other filter delegate plugin to implement it.

You should either add this interface in order filter delegate s.t. it's required and handled automatically for all filter delegate plugins, or you should find a different way to implement this without breaking the interface contract.

hummingbot/strategy/pure_market_making/pure_market_making_v2.pyx
```diff
@@ -152,6 +157,10 @@ cdef class PureMarketMakingStrategyV2(StrategyBase):
     def sizing_delegate(self) -> OrderSizingDelegate:
         return self._sizing_delegate
 
+    @property
+    def sizing_delegate_name(self) -> str:
+        return self._sizing_delegate.name
```
martinkou: Again, if you expect all sizing delegates to have a name, you must add it to the interface s.t. it's properly defined for all plugins of this type. The way you do it now is architecturally incorrect.

hummingbot/strategy/pure_market_making/pure_market_making_v2.pyx
```diff
@@ -346,6 +363,31 @@ cdef class PureMarketMakingStrategyV2(StrategyBase):
             object market_info = self._sb_order_tracker.c_get_market_pair_from_order_id(order_id)
             LimitOrder limit_order_record
 
+        # Replenish Delay and order filled stop cancellation are only for single order mode
+        # (identified using name of the delegate)
+        if self.sizing_delegate_name in self.SINGLE_ORDER_SIZING_DELEGATES:
```
martinkou: Anybody should be able to write new order sizing delegates and have it behave correctly. You should not make special cases here and make two specific sizing delegates have special functions.

veecos2: Hey Martin, the idea is to basically roll these out for single order mode for now. I can make the strategy always pass in the mode and check with it, instead of checking with the name of the delegates.

martinkou:
As discussed earlier today... a few changes are still needed.

 1. The checks for single vs. multiple is only a temporary thing for testing this new feature behinds flags. It **must not** involve any interface changes as interface changes are permanent.
 2. The filter delegate must stay as a plugin. Do not directly integrate it into the pure market making strategy.
 3. Remove the `mode` hack.
 4. All interface definitions should stay proper and allow other users to write and use their own plugins without problem.

PR #800, opened 2019-09-18
feat / asyncio wrapper

yzhang1994:

- Wrap asyncio.gather() and asyncio.ensure_future() to always report errors on background tasks


hummingbot/core/utils/async_utils.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def asyncio_ensure_future(*args, **kwargs):
+    try:
+        return asyncio.ensure_future(*args, **kwargs)
+    except Exception as e:
+        logging.getLogger().error(f"Unhandled error in background task: {str(e)}", exc_info=True)
+
+
+async def asyncio_gather(*args, **kwargs):
```
martinkou: The wrapped gather function should still raise errors s.t. outer layers can catch them again.

So, right after the debug log, there should be a `raise` to forward the error to outside.

hummingbot/core/utils/async_utils.py
```diff
@@ -0,0 +1,16 @@
+import asyncio
+import logging
+
+
+def asyncio_ensure_future(*args, **kwargs):
+    try:
+        return asyncio.ensure_future(*args, **kwargs)
```
martinkou: This is not effective.

The `ensure_future()` call itself is not the problem - it won't cause any exceptions. It's the coroutine that gets scheduled that may have silent errors.

If you look at the definition of `ensure_future()` from Python's documentations, the first argument is the coroutine or await-able object. What you need to do is to wrap that thing in another async function.

e.g.

```
  async def safe_wrapper(c):
    try:
      return await c
    except asyncio.CancelledError:
      raise
    except Exception:
      # do some error log here

  # assume coro is defined here
  return asyncio.ensure_future(safe_wrapper(coro))
```

hummingbot/core/utils/async_utils.py
```diff
@@ -0,0 +1,16 @@
+import asyncio
+import logging
+
+
+def asyncio_ensure_future(*args, **kwargs):
```
martinkou: The name should reflect the purpose of the function. I'd recommend naming it `safe_ensure_future()` instead.

hummingbot/core/utils/async_utils.py
martinkou: Similarly, this should be named `safe_gather()`

hummingbot/core/utils/async_utils.py
yzhang1994: Ok fixed.

PR #829, opened 2019-09-25
refactor / unit tests for jenkins | fix / idex tests

carlolm:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:

- reorganize jenkins files: binance
- add idex to jenkins tests
- read keys from environment for tests
- cleanup idex tests


conf/__init__.py
```diff
@@ -45,6 +45,9 @@
 
 # IDEX Tests
 idex_api_key = os.getenv("IDEX_API_KEY")
+test_idex_erc20_token_address_1 = os.getenv("IDEX_TOKEN_ADDRESS_1")
+test_idex_erc20_token_address_2 = os.getenv("IDEX_TOKEN_ADDRESS_2")
```
carlolm: @chowryan I don't think these token addresses are actually used in idex tests.
Only the ticker?

test/integration/test_idex_market.py
```diff
@@ -79,17 +74,17 @@ class IDEXMarketUnitTest(unittest.TestCase):
...
+                                                       conf.test_idex_erc20_token_address_2],
                                 chain=EthereumChain.MAIN_NET)
         cls.market: IDEXMarket = IDEXMarket(
             idex_api_key=conf.idex_api_key,
             wallet=cls.wallet,
-            ethereum_rpc_url=conf.test_ddex_web3_provider_list[0],
+            ethereum_rpc_url=conf.test_web3_provider_list[0],
             order_book_tracker_data_source_type=OrderBookTrackerDataSourceType.EXCHANGE_API,
-            symbols=[ETH_FXC]
+            symbols=[ETH_ZRX]
```
chowryan: change FXC to ZRX will break the tests (minimum order sizes)

conf/__init__.py
chowryan: dont understand your question. why are you adding this?

chowryan:
be careful of selecting trading pairs on these small exchanges too. ZRX-ETH on IDEX is currently <1 ETH 24 hour trade volume which means the spread is probably terrible too. potentially lose a decent amount of ETH if you do a market buy and sell

conf/__init__.py
carlolm: @chowryan added those because the original tests refer to:

```
erc20_token_addresses=[conf.test_ddex_erc20_token_address_1,
conf.test_ddex_erc20_token_address_2],
```


https://github.com/CoinAlpha/hummingbot/blob/4941ec38b8090b8f458d19ca5ad98d667f561b4e/test/integration/test_idex_market.py#L84

Maybe this is a carry over from copying over the ddex tests, that wasn't updated or cleaned up.

I added these to conf so the variables (that are referenced in the test) need to be sourced from the environment variables to run in jenkins.

Therefore my question, are these actually being used in the tests? (they seem to be not)
image

carlolm:
> cant just change FXC to ZRX without changing buy / sell amounts that meet the minimum order size

See this commit: sizes were updated:
https://github.com/CoinAlpha/hummingbot/pull/829/commits/c3c1f347ac5a3c1c205d4fd26e3416834e2f42eb

Ideally these tests should actually get the price and set quantity based on 0.15 / price, so that mins are maintained without manually having to hard code and change quantities if prices drop.

The market trades were the only ones where the quantity needed updating:
image



Since these other quantities are by definition > 0.15 ETH:
image

carlolm:
@chowryan @veecos2 agreed, we can change to a different pair.  FXC no longer works because it's not available for US users

carlolm:
@chowryan @veecos2 changed to QNT

chowryan:
the prices are not hardcoded, only the amounts are hardcoded as there is no easy way to get minimum order size from market classes. It's consistent with all other market tests. If we want to prevent hardcoding a reasonable order size amount close to minimum from each exchange, we'll need to add that functionality to each market class.

PR #873, opened 2019-10-04
feat / performance analysis using trade history

michaelc9:

**Optional**:
- Related Github issue: https://github.com/CoinAlpha/hummingbot/issues/732
- Related Clubhouse Story:
https://app.clubhouse.io/coinalpha/story/5284/calculate-performance-based-on-trade-history-rather-than-inventory

Design Decision:
- Trade performance is independent of inventory balance changes so that deposits/withdrawals and manual trades will not affect calculation. 
- Profit percentage is calculated by assets spent and acquired in trades (unused balance is irrelevant).

Terms:
- ```session_trades```: trades produced by current strategy during Hummingbot runtime
- ```spent_amount```: total amount of asset used from trades
- ```acquired_amount```: total amount of asset acquired from trades

**Performance Calculation:**

- For each market, calculate total spent and total acquired amount for each asset from ```session_trades```.
Final delta percentage for each asset is defined by: ```acquired / spent```
- Portfolio performance is calculated from both base and quote asset delta, using latest price to convert base asset to quote asset.

For example:
- Trades: 
  - buy 1 ETH with 100 DAI
  - sell 0.9 ETH and acquired 100 DAI
  - buy 0.1 ETH with 10 DAI
  - sell 1 ETH and acquired 100 DAI

- Total ```spent_amount``` for DAI = 110 and ```acquired_amount``` = 200
- Total ```spent_amount``` for ETH = 1.9 and ```acquired_amount``` = 1.1

- Performance for DAI = 200/110 = 81.8 %
- Performance for ETH = 1.1/1.9 = - 42.1 %
- Performance for ETH_DAI portfolio (assuming ETH price is 100 DAI)  
=  ( 200 + (1.1 * 100) ) / ( 110 + (1.9 * 100) )  = 310 / 300
=  3.3 %


hummingbot/client/command/config_command.py
```diff
@@ -114,6 +114,9 @@ def get_all_available_config_keys() -> List[str]:
         self.placeholder_mode = False
 
         if choice.lower() in {"y", "yes"}:
+            # Clear application states that are specific to config
+            self.starting_balances = {}
```
yzhang1994: I think this line belongs in `self._stop_loop`

hummingbot/client/performance_analysis.py
```diff
@@ -1,17 +1,37 @@
-from typing import Tuple
-
+import logging
+from typing import (
+    Tuple, Dict, List
```
yzhang1994: Change to multiline import

hummingbot/client/performance_analysis.py
```diff
@@ -80,3 +100,133 @@ def compute_return(self, price: float) -> float:
...
+                    source="default"
+                )
+        if trade.trade_type == TradeType.SELL.name:
+            net_base_delta: float = trade.amount
+            net_quote_delta: float = trade.amount * trade.price * (1 - float(trade_fee["percent"])) - total_flat_fees
+        elif trade.trade_type == TradeType.BUY.name:
+            net_base_delta: float = trade.amount * (1 - float(trade_fee["percent"])) - total_flat_fees
+            net_quote_delta: float = trade.amount * trade.price
+        else:
+            raise Exception(f"Unsupported trade type {trade.trade-type}")
```
yzhang1994: should be f"...{trade.trade_type}..."

hummingbot/client/performance_analysis.py
```diff
@@ -80,3 +100,133 @@ def compute_return(self, price: float) -> float:
...
+                    asset_stats[quote_asset]["spent"] += quote_delta
+
+            market_trading_pair_stats[market_trading_pair_tuple] = {
+                "starting_quote_rate": queried_trades[0].price,
+                "asset": asset_stats
+            }
+
+        return market_trading_pair_stats
+
+    def calculate_trade_performance(self,
```
yzhang1994: For larger functions like this, could you add some in-code comments?
(I know we don't really have a standard now, but would be good to start doing that in case someone needs to make changes in the future).

hummingbot/model/trade_fill.py
```diff
@@ -21,7 +16,7 @@
     Float,
     JSON
 )
-from sqlalchemy.orm import relationship
+from sqlalchemy.orm import relationship, Session
```
yzhang1994: multiline import

hummingbot/client/command/config_command.py
michaelc9: If we reset the starting balance in stop loop, we will lose starting balance data when we start and stop strategies.
We only want to reset starting balance when we reconfig markets/strategies.

hummingbot/client/performance_analysis.py
martinkou: It is still `trade-type` as of today, which is incorrect.

hummingbot/client/performance_analysis.py
```diff
@@ -80,3 +100,133 @@ def compute_return(self, price: float) -> float:
         _, delta = self.compute_delta(price)
         percent = (delta / starting_amount) * 100
         return percent
+
+    @staticmethod
+    def calculate_trade_asset_delta_with_fees(trade: TradeFill):
```
martinkou: Add annotation for return types, `Tuple[float, float]`

PR #904, opened 2019-10-11
docs / Add documentation for Simple Trade strategy

yuna878:

Add markdown documentation for Simple Trade strategy


documentation/docs/developers/strategies/simple-trade.md
```diff
@@ -0,0 +1,12 @@
+# Simple Trade
+
+## Overview
+
+Simple trade strategy expands upon Perform Trade strategy. It has a few additional features such as time constraints and order specific loggings.
```
martinkou: It's better to describe the differences in terms of what the tutorial reader can learn from the strategy's code. The primary purpose of the tutorial is to teach a reader new tricks he or she can do inside a strategy, or a new mechanism with the strategy code that can be used - in this case, using the `c_tick()` mechanism to add a time delay before the trade.

martinkou: Another thing I'd suggest to add is a simple flow chart, like the perform trade documentation. Visual aids are useful for developers to grasp new concepts quickly.

documentation/docs/developers/strategies/simple-trade.md
```diff
@@ -0,0 +1,12 @@
...
+
+#### Time restriction
+There is a time restriction `_time_delay` set that will ensure that there is a given amount of time in between the trades. When an order is placed in `c_process_market()`, it checks that the current timestamp is greater than previous order’s timestamp plus delay time. If it is a valid time to submit new order, the order will be processed.
```
martinkou: Again, it's better to list out the differences in point form. It's easier for readers to scan through the documentation in that format.

You should also talk about how these two new attributes are related to the arguments in the constructor.

documentation/docs/developers/strategies/simple-trade.md
```diff
@@ -0,0 +1,12 @@
...
+## Overview
+
+Simple trade strategy expands upon Perform Trade strategy. It has a few additional features such as time constraints and order specific loggings.
+
+#### Time restriction
+There is a time restriction `_time_delay` set that will ensure that there is a given amount of time in between the trades. When an order is placed in `c_process_market()`, it checks that the current timestamp is greater than previous order’s timestamp plus delay time. If it is a valid time to submit new order, the order will be processed. 
+There is also a restriction `_cancel_order_wait_time` placed to cancel active orders. If the current timestamp is greater than order’s start time plus cancel wait time, the order is cancelled.
+
+#### Logging
+Helpful logging about order placement status is provided. The functions `c_did_fill_order()`, `c_did_complete_buy_order()`, and `c_did_complete_sell_order()` will log when order is either filled or completed.
```
martinkou: And again... it's better to describe this in terms of a new mechanism that the reader can use to develop his/her own strategies. In this case, it's the event handling mechanism.

When you go into the details of what each of these functions is doing, it may be better to use point form as well.

carlolm:
OK to test

martinkou:
@yuna878 Can you add a simple SVG flow chart as well?

martinkou:
I think the overall contents are of the diagrams are good. But still some changes needed:

 1. In flow charts, diamonds are decision points and rounded rectangles are "things that happen". The syntax you used in the current diagrams are the opposite of what people are used to seeing.

 2. You've reversed the two diagrams. The first one that intends to say order processed, has order cancelled as the end state.

 3. "Order processed" is actually triggered by an external event. The way your diagram currently says is really that the order would get processed after a time delay - which is not true. You can refer to the bottom diagram in https://docs.hummingbot.io/developers/strategies/cross-exchange-market-making/ for how we express external events - we use the "Document" symbol to do that. It's in the draw.io palette of flow chart symbols.

 4. There should be some textual description before or after the diagrams saying they belong to two possible flows. One is the order timed out and had to be cancelled. The other is the order got filled before the timeout.

PR #906, opened 2019-10-11
feat / log rotation

yzhang1994:

Expected behaviors:
- A new log file will be created every day
- When a new log file is created, if there are more than 7 files, the oldest ones will be deleted.


hummingbot/templates/hummingbot_logs_TEMPLATE.yml
```diff
@@ -23,11 +23,14 @@ handlers:
         formatter: simple
         stream: ext://sys.stdout
     file_handler:
-        class: logging.FileHandler
+        class: logging.handlers.TimedRotatingFileHandler
         level: DEBUG
         formatter: simple
         filename: $PROJECT_DIR/logs/hummingbot_logs_$DATETIME.log
```
martinkou: I think the log rotation algorithm wouldn't be able to detect the old logs of their base names (i.e. the name before the timestamp suffix from the log rotator) are different.

We should probably change the log filename scheme here... maybe use `hummingbot_logs_{strategy_file_name}` instead... we don't have a tag defined for the strategy file name right now. So to do that, you'd need to add the replace tag to `init_logging()` as well.

yzhang1994: @martinkou the issue is that users select the strategy after logging starts. We can put a default value there, but that will result in a weird separation between logs produced before and after the user selects a strategy. If we don't want to preserve logs past 7 days, we can just name the file `hummingbot_logs` without any suffix.

martinkou: Just did some discussion in office about this... the conclusion we have is this:

 1. Before we know which strategy config file is being used, just log to a general log file, but with process id included in individual lines to separate the lines from different instances.
 2. Once we know which strategy config file is being used, incorporate a truncated conf file name to the log file's filename, and do `init_logging()` again to make sure all new logs would go there.

For 2, we should still keep the process ID in the log lines, s.t. log lines between the general log before the strategy was initiated, and those inside the more specific log file can be correlated. This means we'll just have one log format string for both types of log files.

yzhang1994: Implemented accordingly

hummingbot/__init__.py
```diff
@@ -102,7 +102,10 @@ def add_remote_logger_handler(loggers):
         root_logger.error("Error adding remote log handler.", exc_info=True)
 
 
-def init_logging(conf_filename: str, override_log_level: Optional[str] = None, dev_mode: bool = False):
+def init_logging(conf_filename: str,
+                 override_log_level: Optional[str] = None,
+                 dev_mode: bool = False,
+                 strategy_file_path: str = "conf_unselected_strategy"):
```
martinkou: nitpick... how about making the default name not sound like it's a strategy? Something generic like "logs_hummingbot.log" (which means the suffix is "hummingbot") should do the job and cleanly separate the general log file from the strategy specific ones.

yzhang1994: @martinkou Sure. Changed.

PR #913, opened 2019-10-14
Arctek/fix bamboo 912

Arctek:
Addresses issue https://github.com/CoinAlpha/hummingbot/issues/912.

Radar Relay connector is likely also going to need the `_account_available_balances` change as well, otherwise the market stalls with a 0 availability amount.


yzhang1994:
test this please

yzhang1994:
ok to test

PR #944, opened 2019-10-19
docs / user manual - strategies

dennisocana:

Going over the strategies section of the User Manual to make the documentation more encouraging to read and easier to understand especially to non-technical users. Also checking on some outdated items like prompts on configuration walkthrough.

I conducted a survey for a small group of people I know about our existing documentation. I taught them first about the basic terms and definitions prior to making them read the pure market making strategy. I think they're the best group we can compare to new non-technical users who are interested to try Hummingbot. IMO the existing docs are fine because I already understand how Hummingbot works but a new user's point of view can be different. This rework is mostly on the presentation of the docs based on the feedback I received.

- Generalized prompts for API and secret keys instead of having one row each for Binance, IDEX, etc. If we'll be adding more exchanges in the future, doing this will be more convenient and doesn't have to take too much area in the docs.
- Removed other available options for strategy. If the article is about pure market making, the description would only say 'Enter `pure_market_making`'.
- Changed some headings level to adjust the table of contents.
- Promoted penny jumping mode to be removed under order adjustment based on filled events.
- Minor edits like using uppercase and lowercase, inconsistencies of using period.


mifeng:
Great work @dennisocana! I really like how you polled your friends to see if the docs make sense to them. Overall, I think all the features we've added recently, particularly to the pure market making strategy, might make it intimidating for new users.

In the future, I think we should have a basic / advanced split for that strategy, so that the **Configuration Walkthrough** only has the parameters needed for basic users: exchange, token_symbol, order_amount,  cancel_order_wait_time, and bid/ask spreads. Then, we can have an Advanced section with single/multiple orders, inventory skew, penny-jumping, etc. I will create a story for this in our backlog.

dennisocana:
@mifeng Thanks! I agree, same as what we did to cross-exchange where the advanced features are like "hidden". We'll keep working on improving the docs to make it less intimidating to new users while I'm working on creating new tutorial videos for installing Hummingbot using the scripts.

PR #960, opened 2019-10-22
Add Missing Templates for dev strategies

HazemElAgaty:
The problem:
The `dev_*` strategies do not have templates with the same naming convention used causing errors when trying to run those strategies from the UI.

Changes:
- Fixed the naming of the `dev_*` templates.
- Fixed wrong config map import in `dev_0_hello_world` strategy.


carlolm:
OK to test

PR #964, opened 2019-10-22
feat / add conf version

yzhang1994:

- Added version number to config files 

Expected behaviour:
- When the bot starts, it will load config vars from `conf_global.yml`. If the global config version is outdated, it will load contents of the template, and copy existing configs from the old file to the new one. (So that the user doesn't have to re-config API keys). Any outdated config variables will be deleted, and any empty variables will be prompted.
- When the user chooses a strategy, the same will happen to the strategy config file.

Potential issue:
- If there is a user error -- imports the wrong config file (e.g. conf_arbitrage.yml for discovery strategy), the old config file will be overwritten. This may be annoying, but since it is the strategy config, there shouldn't be any essential info and the user just has to re-config.
- If there is a mismatch between the conf template and the config map object, there will be an error and the missing variable from the template will not be saved. In order to prevent this, I added unit tests to ensure all templates are in sync with the config map object.


hummingbot/client/command/config_command.py
```diff
@@ -207,10 +207,6 @@ def get_all_available_config_keys() -> List[str]:
             self._notify('Invalid choice. Please enter "create" or "import".')
             strategy_path = await self._import_or_create_strategy_config()
 
-        # Validate response
```
martinkou: Why is this deleted?

yzhang1994: @martinkou I realized that for this particular prompt, the answer is validated twice -- at line 234 in the same file. Normally this is ok. But in this particular case, since we try to re-load and rewrite the config file after a set of strategy config is selected, it causes an async bug

PR #998, opened 2019-10-29
Fix for available balances.

Arctek:
Fix for if API fails to store an order, proactively cancel and untrack.

Fix for available balances on the Bamboo market class.
Fix for a rare case where the API could respond with a 404 for a tracked limit order.


martinkou:
ok to test

Issue #1032, opened 2019-11-05
[BUG] Cannot create orders for pairs with very low price in Bittrex


dennisocana:
**Describe the bug**

This was reported by a user trading FXC-BTC. Getting the following errors in logs.

```
2019-11-05 11:30:02,983 - 15812 - hummingbot.strategy.pure_market_making.pure_market_making_v2 - INFO - (FXC-BTC) Creating limit bid orders at (Size, Price): ['1794.70736022 FXC, 3.1E-7 BTC'] [clock=2019-11-05 03:30:03+00:00]
2019-11-05 11:30:04,080 - 15812 - hummingbot.market.bittrex.bittrex_market - NETWORK - Error submitting buy LIMIT order to Bittrex for 1794.70736022 FXC-BTC 3.1E-7.
Traceback (most recent call last):
  File "hummingbot\market\bittrex\bittrex_market.pyx", line 741, in hummingbot.market.bittrex.bittrex_market.BittrexMarket.execute_buy
  File "hummingbot\market\bittrex\bittrex_market.pyx", line 702, in place_order
  File "hummingbot\market\bittrex\bittrex_market.pyx", line 996, in _api_request
  File "hummingbot\market\bittrex\bittrex_market.pyx", line 1004, in hummingbot.market.bittrex.bittrex_market.BittrexMarket._api_request
OSError: Error fetching response from POST-https://api.bittrex.com/v3/orders. HTTP Status Code 400: {'code': 'BAD_REQUEST', 'detail': 'Refer to the data field for specific field validation failures or extra fields in the body that were rejected.', 'data': {'newOrder.limit': ['Error converting value "3.1E-7" to type \'System.Nullable`1[System.Decimal]\'. Path \'limit\', line 1, position 102.']}}
2019-11-05 11:30:04,082 - 15812 - hummingbot.market.bittrex.bittrex_market - WARNING - Failed to submit buy order to Bittrex. Check API key and network connection.
```

Used the same configs on two different pairs and it worked fine. Probably when the trading pair's price is too low it results to this error. Current price of FXC-BTC is 0.00000030

```
Error submitting buy LIMIT order to Bittrex for 1794.70736022 FXC-BTC 3.1E-7
```

```
Error converting value "3.1E-7" to type \'System.Nullable`1[System.Decimal]
```

**Screenshots**

[...]

PR #1050, opened 2019-11-08
fix / Bittrex unable to place orders with low prices

CrimsonJacket:

**Optional**:
- Related Github issue:
#1032 
- Related Clubhouse Story:

Bug #1032: 
Placing orders with extremely low prices(`3.17E-7`) on Bittrex would fail.

Current Fix:
Enforce f-string literals when crafting signatures and request parameters.


yzhang1994:
@CrimsonJacket If you are just fixing a bug in Bittrex, might not be a good idea to change the other market connectors. Is there a reason why they need to be changed?

CrimsonJacket:
> @CrimsonJacket If you are just fixing a bug in Bittrex, might not be a good idea to change the other market connectors. Is there a reason why they need to be changed?

This bug potentially affects other connectors considering that they use `Decimal` for price. I did make sure to test that placing of orders across the connectors still works.

martinkou:
Discussed this w/ Daniel last week. The `str({decimal object})` method has a problem that it'll output scientific notation when the decimal value is too small. That, as I understand, is a problem for all exchange APIs.

One superficially risky change here is changing the output type of price and size from float to str in Coinbase Pro connector. But according to their API specs, that's actually correct as they expect a string there.

martinkou:
I think I'd merge this and then allow for testing on Friday.

PR #1052, opened 2019-11-11
feat/vwap

yuna878:

Add new VWAP strategy and test cases.


yzhang1994:
ok to test

martinkou:
The VWAP strategy should wait for the previous order to be completely filled before issuing a follow up order.

In the current way that `c_process_market()` is written... the VWAP strategy would just split the order according to volume metrics but issue an order every tick. Please fix.

hummingbot/strategy/dev_5_vwap/dev_5_vwap.pyx
```diff
@@ -0,0 +1,446 @@
...
+
+from libc.stdint cimport int64_t
+from hummingbot.core.data_type.order_book cimport OrderBook
+from datetime import datetime
+
+NaN = float("nan")
+s_decimal_zero = Decimal(0)
+ds_logger = None
+
+cdef class Dev5TwapTradeStrategy(StrategyBase):
```
martinkou: Change the name to Dev5VWAPTradeStrategy

test/test_vwap.py
```diff
@@ -0,0 +1,370 @@
...
+        fill_events = self.maker_order_fill_logger.event_log
+        self.assertEqual(1, len(fill_events))
+        bid_fills: List[OrderFilledEvent] = [evt for evt in fill_events if evt.trade_type is TradeType.BUY]
+        self.assertEqual(1, len(bid_fills))
+
+        # test whether number of orders is two after time delay
+        # check whether the order is buy
+        # check whether the price is correct
+        # check whether amount is correct
+        if bid_order.quantity < 100:
```
martinkou: This should be an assert rather than an if statement. If the bid order size is incorrect... then if statement would just skip it and not print an error during unit testing.

martinkou:
Remember during last week's phone call, I talked about the need to check for canceled / failed orders. The reason is, if you don't check for those, and the user cancelled the order on the exchange manually... then this strategy would wait forever.

Let's do something like this..

 1. Write a common function that validates the `order_id` - making sure it was made by this strategy object - and sets `_has_outstanding_order` to False. Something like this...

    ```
    cdef c_acknowledge_last_order_completion_or_failure(self, object order_event):
                cdef:
                    str order_id = order_event.order_id  # this works because all order events have order_id
                    object market_info = self._sb_order_tracker.c_get_market_pair_from_order_id(order_id)
                if market_info is not None:
                    self._has_outstanding_order = False
    ```
 2. Add the event handlers `c_did_fail_order()`, `c_did_cancel_order()` and `c_did_expire_order()`, and call the function above to unblock the strategy.

hummingbot/strategy/dev_5_vwap/dev_5_vwap.pyx
```diff
@@ -0,0 +1,474 @@
...
+                                          f"making may be dangerous when markets or networks are unstable.")
+
+            for market_info in self._market_infos.values():
+                self.c_process_market(market_info)
+        finally:
+            self._last_timestamp = timestamp
+
+    cdef c_check_last_order(self, object order_event):
+        """
+        Check latest order to see it if was completed. If not completed, set self._has_outstanding_order to False
```
martinkou: The function is doing the right thing, but the comments here is a bit misleading... the `if` statement in this function is really just checking whether the order was made by this strategy or not - it's possible for a market connector to emit order events that were not emitted by the strategy object. e.g. the order may be made by the user manually, or made by another bot on the same account.

PR #1106, opened 2019-11-22
feat/KuCoin connector part 2: UserStreamTracker

vic-en:
KuCoin UserStreamTracker & Tests


CrimsonJacket:
ok to test

conf/__init__.py
```diff
@@ -72,10 +72,6 @@
 
 test_web3_provider_list = [os.getenv("WEB3_PROVIDER")]
 
-# Liquid Tests
-liquid_api_key = os.getenv("LIQUID_API_KEY")
-liquid_secret_key = os.getenv("LIQUID_SECRET_KEY")
-
 # Wallet Tests
```
CrimsonJacket: You should not have to deal with other connectors

vic-en: That wasn't intentional. I have corrected it.

PR #1127, opened 2019-11-26
docs/vwap

anyachopra97:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:


martinkou:
** Comments on the flow charts **

1. You should refrain from abusing the "three diamonds with two dashes" sign in flow charts. Conventionally it means concurrency - and we've taken it to model some "for each" loops in the sense that "for each" loops can sometimes be equivalent to map-reduce, and thus can be thought of working like concurrency - even though technically it's processed serially.

    * Say, for example, in the place order flow diagram... that flow is only handling one order at a time. So there's no "for each" loop there.

    * What's more appropriate could be the traditional circle / semicircle flowchart terminal symbols - which you can find us using them in other doc pages. Or, you may just remove them altogether.

2. The order of the diagrams, and thus the descriptions coming with them... seem reversed. It's better to describe the overall flow at first (i.e. what happens at `c_tick()`), and then drill in the specifics (i.e. what happens in `c_place_order()`)... than to describe the specific first and then the general.

martinkou:
**Comments on formatting issues**

The VWAP docs currently look like this when rendered to HTML (via `mkdocs`).

Screen Shot 2019-12-03 at 5 46 19 PM

There're a few formatting problems that should be fixed:

1. All the direct code references, like, `c_process_order()`, `self._is_vwap`, should be quoted as code blocks. We did that in the TWAP documentation before. In Markdown, that quote is indicated by backticks. Just Google for the syntax of it - and use `mkdocs serve` to look at the rendered output in near real time as you edit and save the file. 

2. An indent level in Markdown for bullet lists is 4 spaces. You're currently using 2 spaces, which is why the indent levels are incorrect in the rendered output.

3. I'd recommend you to use code quote for the pricing / sizing formulas as well.

martinkou:
**Comments on contents**

I like the detailed description on how the VWAP logic works. But right now it's lacking a bit on why it's written like this.

It'd be better to include a paragraph like this at the beginning of the "Strategy" section, to let people understand why we're determining the individual order sizes by `percent_slippage`.

--
The VWAP strategy tries to limit the market impact of each individual user order, by calculating the possible order sizes that can be made if we want to guarantee a price slippage below a certain, user-specified level. It would always calculate the price slippage and corresponding order sizes as if the orders to make are market orders, even if the user requested to make limit orders - because the goal is to limit the market impact of orders, not the actual price slippage of the orders.
--

documentation/docs/developers/strategies/vwap.md
```diff
@@ -0,0 +1,53 @@
+# VWAP
+
+The VWAP strategy is a common algorithmic execution strategy which allows traders to account for slippage by splitting up orders over time. Specifically, the VWAP strategy helps traders minimize slippage when buying or selling large orders. The methods utilized in VWAP make the strategy more useful to traders and will help when creating future, more complex strategies.
```
martinkou: This is the old TWAP description... and it doesn't description VWAP well.

Change to this..

> The VWAP strategy is a common algorithm execution strategy which allows traders to execute large orders with controlled and limited market impact. It accomplishes this by measuring the market impact of each order before making new orders, and splitting up the trader's large order into smaller orders that have limited market impact. Unlike the TWAP strategy, VWAP doesn't wait for a specific interval before making new orders - it only waits for the last order to complete before making the next order.

documentation/docs/developers/strategies/vwap.md
```diff
@@ -0,0 +1,53 @@
...
+* `order_percent_of_volume` : percent of open order volume (at the specified price) that user wants to cap the order at (ex. user wants to cap the order volume at 0.1% of total available order volume)
+
+## Strategy
+
+The VWAP strategy attempts to limit the market impact of each individual user order. It accomplishes this by calculating the potential order sizes that can be placed in order to guarantee a user-specified upper bound on price slippage. In this strategy, the price slippage and corresponding order sizes are always calculated as if they are market orders, even if this is not the case. By doing do, the strategy achieves the goal of limiting the market impact of orders rather than limiting the actual price slippage of the orders.
+
+To achieve the goal of limiting the market impact of orders, the VWAP strategy fetches the order book and calculates the total open order volume up to percent_slippage. If no order is outstanding, an order is submitted which is capped at order_percent_of_volume * open order volume up to percent_slippage. The previous order is filled before the next is submitted and if an order is currently outstanding no action occurs.
```
martinkou: Code-quote `order_percent_of_volume * open order volume` and `percent_slippage`.

documentation/docs/developers/strategies/vwap.md
```diff
@@ -0,0 +1,53 @@
...
+
+The flow chart below details the flow of processing orders.
+
+![Figure 1: Processing orders](/assets/img/VWAP2.svg)
+
+Only one order is placed in a clock tick, so a state machine is needed to emit multiple orders over different clock ticks.
```
martinkou: Change the first clause to "At most one order can be placed in each clock tick"

TomasFlow (approved):

PR #1211, opened 2019-12-19
feat / External price sources for pure MM

Nullably:

Related Clubhouse Story: 6571

A description of the changes proposed in the pull request:

- Create an additional set of configuration for external sources: exchange, feed, api
- Reconfigure pure mm strategy to use prices from external sources for orders proposal
- Add a filter to remove any taker orders
- Add custom api data feed
- Unit test most of these new features.
- Docs updated.


hummingbot/strategy/pure_market_making/pure_market_making_v2.pyx
```diff
@@ -295,10 +318,40 @@ cdef class PureMarketMakingStrategyV2(StrategyBase):
...
+    # and sell order price lower than first bid on the order book.
+    cdef object c_filter_orders_proposal_for_takers(self, object market_info, object orders_proposal):
+        cdef :
+            list buy_prices = []
+            list buy_sizes = []
+            list sell_prices = []
+            list sell_sizes = []
+        if orders_proposal.buy_order_sizes[0] > 0:
+            for i in range(0, len(orders_proposal.buy_order_prices)):
+                first_ask = market_info.get_price(True)
```
martinkou: While this works... it's slow performance-wise.

 1. We're in a `c_` function... we should minimize the number of pure Python calls if we can help it. So instead of `market_info.get_price()`, this should be `market.c_get_price()` instead. The `market` object should be stored as a `cdef` allocated type in the beginning so we don't have to go through the tuple attribute traversal.
 2. It doesn't need to be called for each iteration. It should be cached before the `for ...` loop.

hummingbot/strategy/pure_market_making/pure_market_making_v2.pyx
```diff
@@ -295,10 +318,40 @@ cdef class PureMarketMakingStrategyV2(StrategyBase):
...
+    # To filter out any orders that are going to be taker orders, i.e. buy order price higher than first ask
+    # and sell order price lower than first bid on the order book.
+    cdef object c_filter_orders_proposal_for_takers(self, object market_info, object orders_proposal):
+        cdef :
+            list buy_prices = []
+            list buy_sizes = []
+            list sell_prices = []
+            list sell_sizes = []
+        if orders_proposal.buy_order_sizes[0] > 0:
+            for i in range(0, len(orders_proposal.buy_order_prices)):
```
martinkou: This is not necessary here. We should use Python idiomatic loops whenever possible.

e.g.
```
for buy_price in orders_proposal.buy_order_prices:
  ...
```

hummingbot/strategy/pure_market_making/pure_market_making_v2.pyx
```diff
@@ -295,10 +318,40 @@ cdef class PureMarketMakingStrategyV2(StrategyBase):
...
+                first_ask = market_info.get_price(True)
+                if first_ask.is_nan() or (orders_proposal.buy_order_prices[i] < first_ask):
+                    buy_prices.append(orders_proposal.buy_order_prices[i])
+                    buy_sizes.append(orders_proposal.buy_order_sizes[i])
+        if orders_proposal.sell_order_sizes[0] > 0:
+            for i in range(0, len(orders_proposal.sell_order_prices)):
```
martinkou: Ditto. Please use Python idiomatic for loops.

hummingbot/strategy/pure_market_making/pure_market_making_v2.pyx
```diff
@@ -295,10 +318,40 @@ cdef class PureMarketMakingStrategyV2(StrategyBase):
...
+                first_bid = market_info.get_price(False)
```
martinkou: Ditto. Please use the faster way to fetch top prices.

setup.py
```diff
@@ -132,6 +132,7 @@ def main():
           ext_modules=cythonize(["hummingbot/**/*.pyx"], language="c++", language_level=3, nthreads=cpu_count),
           include_dirs=[
               np.get_include(),
+              "D:\\GITHub\\hummingbot\\"
```
martinkou: Remove this line.

What was broken to cause you to add this?

martinkou:
Looks good to me overall. But some minor nitpicks, and an unnecessary line in setup.py.

hummingbot/strategy/pure_market_making/pure_market_making_v2.pyx
Nullably: I understand, but I need the index (i) for later accessing the buy_sizes array.
` buy_sizes.append(orders_proposal.buy_order_sizes[i]) `

hummingbot/strategy/pure_market_making/pure_market_making_v2.pyx
Nullably: Ok, I'll be more careful with cython function calls.

setup.py
Nullably: Sorry, this was from when I was trying to get hummingsim to work on windows, I will remove this.

hummingbot/strategy/pure_market_making/pure_market_making_v2.pyx
Nullably: You mean having market object in OrderBookAssetPriceDelegate instead of market_info and use c_get_price instead of get_price? I think that makes sense, though after I made the change, my unit testing breaks, it complains 
` 'hummingsim.backtest.backtest_market.BacktestMarket' object has no attribute 'c_get_price'`

Nullably:
I will submit a new one.

PR #1237, opened 2019-12-24
Feat/bitfinex exchange market

malinich:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:

- execute buy\sell limit orders in Bitfinex
- execute buy\sell market orders in Bitfinex.
- cancel orders
- get user balances\wallet
- active order tracker
- ws and rest functional for communication with the exchange


malinich:
At times we see the error that came from the Exchange: HTTP status is 500. ['error', 10114, 'nonce: small']. We write so support 27/12/2019 about it.

Garbis (some guy from the bitfinex support) answered me.

> Periodically we face with error 10114 ("nonce: small"). However, we don't see any reason for that: - Nonce is generated by bfxapi library based on a current UNIX timestamp and hence strongly increasing. - We don't send REST requests in parallel. - We use separate key pairs for different. It seems that sometimes REST requests are mixed by Cloudflare and got by Bitfinex API in a wrong order. The most common reason for error 10114 ("nonce: small") is using the same key in different processes or using a nonce value too small or a bugged function to generate nonces.

But this answer no much helps me.
 the request sometimes passes, sometimes it crashes with such an error.
 See screenshot (the statement execute three times, without modification, three times the same statement. first and second-time crash, on third - norm.  )
I want today to write to support officially.
screenshot

malinich:
answer:
```
Thank you for reaching out to us.

This kind of error generally happens in two cases:

A- using a nonce value too small or a bugged function to generate nonces
B- using the same set of api keys with two different bots

In case A you should ensure that you nonce generator always returns an increased number each time it's called.

If it is case B it is likely that you're sending multiple requests in quick succession, and the requests are being processed in a different order to what you sent them. So the request with the higher nonce is causing the previous request to fail. This apparently seems to be with your case.

You can counteract this by creating multiple API keys for different requests or cycling through a pool of API keys to avoid this happening.

Hope this helps. If you have any other issues or questions, please let us know.
```

malinich:
> I move to think (

I reply
```
my function that generate nonce is take from your documentsation. I put 1000, 100000 . No one help

def _make_nonce(self) -> int:
    nonce = int(round(time.time() * 1000))
    return nonce

I use only one bot. 
I already set random sleep to avoid parallel func execution.

await asyncio.sleep(random.random())

no one help me.
I logged my nonce
1) request
'bfx-nonce': '1577713990001'
'bfx-nonce': '1577713995003',
'bfx-nonce': '1577714001001'

nonce generated everytime increase . It no helped whatever
```

malinich:
support:
```
Are you submitting multiple authenticated requests in quick succession? It could be that one is reaching the server before another with a smaller nonce in which case the first request would then be invalidated. It is common to use asynchronous programming in order to await the response of the first request before sending a new one.

This means if you want to use multiple HTTP or WebSocket connections you will need a separate API key for each client. You can generate as many API keys as you would like.

If you have any other issues or questions, please let us know.
```

CrimsonJacket:
@malinich I assume the less ideal solution for this issue is to simply wait for a longer period of time. This should create a long enough delay so that the exchange receive and process the requests in order. Additionally, I do not think that creating multiple API keys is the right approach to solve this issue. 

For now, could you check if waiting longer would solve this?

nionis:
@malinich have you tried @CrimsonJacket suggestion?

malinich:
> @malinich I assume the less ideal solution for this issue is to simply wait for a longer period of time. This should create a long enough delay so that the exchange receive and process the requests in order. Additionally, I do not think that creating multiple API keys is the right approach to solve this issue.
> 
> For now, could you check if waiting longer would solve this?

I disabled ws, I catch the exception and immediately retry wallet-request, the problem the same. If set long sleep, it does not help anyway (I thought I reached limit request: for a wallet is 45 req/min, but I was wrong) (
Screenshot from 2020-01-08 00-57-08

malinich:
I record video (available while MR living)
https://youtu.be/KF0-qTRTJTA

malinich:
I wrote to support about nonce
```
I wrote earlier about this bug, Sadly you documentation is bad. I copy-paste your code, it almost never executed.  
You can see code from documentation (I increase nonce from 1000 to 10_000_000). It helps sometimes, but it through error sometimes. I record a screencast that helps to see what happens.
P.S. nonce every request generates a new value. Requests no parallel, it simple execute in the cycle with sleep 3 sec.
Because I can't attach the video, I put the link here. https://youtu.be/KF0-qTRTJTA
Because I can't attach the code, I put the link here. https://paste.ofcode.org/Y4P2Fc2KymHRzpkKCpg6uR
```

malinich:
> 

I found solution in GitHub  [see](https://github.com/bitfinexcom/bitfinex-api-node/issues/406)

`
@tuloski @vansergen is correct, you need to generate a new API key since the nonce size has been reduced (it was needlessly multiplied by 1000 before). This is the last time the nonce will be reduced in size, we apologize for the inconvenience.
` 
After recreated new api all test pass

CrimsonJacket:
test this please

nionis:
hey guys, yesterday I finished working on a bunch of updates, including:

- client not showing all bitfinex's trading pairs
- tracking user balance, the issue was that `available_balance` is not provided by bitfinex, so I calculate it using our in flight orders, not ideal at all, this fixes PMM, tested on single and multiple mode
- fixed cancel_all
- reduce `nonce` errors by sleeping for 5 seconds instead of 1, again, not ideal, but I found out this is a good sweet spot
- other small updates

overall, our biggest issue is [this](https://docs.bitfinex.com/docs/api-access), I will be committing the updates during the weekend after I do more testing.

test/integration/test_bitfinex_market.py
```diff
@@ -0,0 +1,368 @@
...
+        self.assertEqual("USD", order_completed_event.base_asset)
+        self.assertEqual("ETH", order_completed_event.quote_asset)
+        self.assertAlmostEqual(base_amount_traded,
+                               order_completed_event.base_asset_amount)
+        self.assertAlmostEqual(quote_amount_traded,
+                               order_completed_event.quote_asset_amount)
+        self.assertTrue(any([isinstance(event, BuyOrderCreatedEvent) and event.order_id == order_id
+                             for event in self.market_logger.event_log]))
+        # Reset the logs
+        self.market_logger.clear()
```
CrimsonJacket: I get what you're trying to do here with `test_execute_limit_buy/sell` but there are some instances when the order does not get filled and thus the `Buy/SellOrderCompleteEvent` is not received. 

Afaik, there is currently no way to test this out since most exchanges do implement a mechanism to detect self-trading and this is generally not the right approach. I would take these tests out.

test/integration/test_bitfinex_market.py
```diff
@@ -0,0 +1,368 @@
...
+
+        [order_cancelled_event] = self.run_parallel(
+            self.market_logger.wait_for(OrderCancelledEvent))
+        self.assertEqual(order_cancelled_event.order_id, order_id)
+        # # Reset the logs
+        self.market_logger.clear()
+
+    def test_limit_sell(self):
+        trading_pair = "ETHUSD"
+        amount: Decimal = Decimal("-0.04")
```
CrimsonJacket: Is there a reason why the amount here is `Decimal("-0.04")`?

test/integration/test_bitfinex_market.py
CrimsonJacket: Essentially this is `test_market_buy` and `test_market_sell`. It will be great if you could follow the naming convention. Also, there seems to be times when the order does not get filled and thus the test would fail. It will be good, if you can manipulate the `quantized_bid/ask_price` more to ensure that this passes.

hummingbot/market/bitfinex/__init__.py
```diff
@@ -1,11 +1,22 @@
...
+
+
+MIN_BASE_AMOUNT_INCREMENT = Decimal("0.01")
+
+# this values ​​set by empirically way, because the bitfinex-market does not have
+# these values. maybe later it will be in market-api.
+BITFINEX_QUOTE_INCREMENT = 0.01
+BITFINEX_BASE_INCREMENT = 1e-8
+TAKER_FEE = Decimal("0.002")
+MAKER_FEE = Decimal("0.001")
```
CrimsonJacket: Do keep updated on any change on the API. This might cause some issues especially if BitFinex decides to change their Trading Fee policy.

hummingbot/market/bitfinex/bitfinex_auth.py
```diff
@@ -40,7 +40,7 @@ def generate_api_headers(self, path, body):
 
     # private methods
     def _make_nonce(self) -> int:
-        nonce = int(round(time.time() * 1000000))
+        nonce = int(round(time.time() * 1_000_000))
```
CrimsonJacket: I am wondering if this is the ideal solution for this. From my understanding of the issues mentioned in previous comments, this has to do with the time in which BitFinex receives the requests and simply multiplying the `nonce` with a larger number should not change this behaviour. 🤔

hummingbot/market/bitfinex/bitfinex_market.pxd
```diff
@@ -0,0 +1,38 @@
+from libcpp cimport bool
+from hummingbot.market.market_base cimport MarketBase
+from hummingbot.core.data_type.transaction_tracker cimport TransactionTracker
+
+
+cdef class BitfinexMarket(MarketBase):
+    # TODO: COIN-12, COIN-13, COIN-14
```
CrimsonJacket: May I ask what this is about?

hummingbot/market/bitfinex/bitfinex_market.pyx
```diff
@@ -0,0 +1,1161 @@
...
+        """
+        *required
+        :return: A lowercase name / id for the market. Must stay consistent with market name in global settings.
+        """
+        return "bitfinex"
+
+    @property
+    def bitfinex_auth(self) -> BitfinexAuth:
+        """
+        """
```
CrimsonJacket: Add a proper function description here.

hummingbot/market/bitfinex/bitfinex_market.pyx
```diff
@@ -0,0 +1,1161 @@
...
+
+    async def _api_platform_status(self):
+        path_url = "platform/status"
+        platform_status = await self._api_public("get", path_url=path_url)
+        return platform_status
+
+    async def _api_platform_config_pair_info(self):
+        path_url = "conf/pub:info:pair"
+        info = await self._api_public("get", path_url=path_url)
+        return info[0] if len(info) > 0 else 0
```
CrimsonJacket: I would default to return an empty `List`. Since subsequently `_format_trading_rules` expects a `List`

hummingbot/market/bitfinex/bitfinex_market.pyx
```diff
@@ -0,0 +1,1161 @@
...
+        """
+        Async wrapper for placing orders through the rest API.
+        :returns: json response from the API
+        """
+        path_url = "auth/w/order/submit"
+        data = {
+            "type": {
+                OrderType.LIMIT.name: "EXCHANGE LIMIT",
+                OrderType.MARKET.name: "MARKET",
+            }[order_type.name],  # LIMIT, EXCHANGE
```
CrimsonJacket: Could you use a ternary operator here for improved code readability?

hummingbot/market/bitfinex/bitfinex_market.pyx
```diff
@@ -0,0 +1,1161 @@
...
+
+        decimal_amount = self.quantize_order_amount(trading_pair, abs(amount))
+        decimal_price = self.quantize_order_price(trading_pair, price)
+        if decimal_amount < trading_rule.min_order_size:
+            raise ValueError(f"Sell order amount {decimal_amount} is lower than the minimum order size "
+                             f"{trading_rule.min_order_size}.")
+
+        try:
+            # sell - is negative amount
+            decimal_amount *= -1
```
CrimsonJacket: You might want to document this in the test case for `test_limit_sell` and `test_market_sell`

hummingbot/market/bitfinex/bitfinex_market.pyx
```diff
@@ -0,0 +1,1161 @@
...
+            cancel_result = await self._api_private("post", path_url=path_url, data=data)
+            # return order_id
+            self.logger().info(f"Successfully cancelled order {order_id}.")
+            self.c_stop_tracking_order(order_id)
+            self.c_trigger_event(self.MARKET_ORDER_CANCELLED_EVENT_TAG,
+                                 OrderCancelledEvent(self._current_timestamp, order_id))
+            return order_id
+
+        except IOError as e:
+            if "order not found" in e.message:
```
CrimsonJacket: According to BitFinex API documentation, the response for a bad cancel request would simply be an empty JSON.

> https://docs.bitfinex.com/reference#rest-auth-cancel-order

nionis:
@CrimsonJacket @malinich here is an idea, we just perform everything (post requests) through a websocket connection

nionis:
updates:

- client not showing all bitfinex's trading pairs
- tracking user balance, the issue was that available_balance is not provided by bitfinex, so I calculate it using our in flight orders, not ideal at all, this fixes PMM, tested on single and multiple mode
- fixed cancel_all
- reduce nonce errors by sleeping for 5 seconds instead of 1, again, not ideal, but I found out this is a good sweet spot, we should instead solve this properly using [this](https://github.com/CoinAlpha/hummingbot/pull/1237#issuecomment-578280705) solution
- other small updates

nionis:
@malinich is it possible to try out PMM and confirm that nonce is still an issue even in a new key?, if it ends up being still a problem, you should consider using a websocket connection like suggested above.

gregzaitsev:
> All tests seems to work 
> I left some comments, do have a look a resolve them

Thanks, @CrimsonJacket, we are looking at your comments. Just a heads up: There is a conflicting merge again that breaks our connector, which delays things.

gregzaitsev:
> updates:
> 
> * client not showing all bitfinex's trading pairs
> * tracking user balance, the issue was that available_balance is not provided by bitfinex, so I calculate it using our in flight orders, not ideal at all, this fixes PMM, tested on single and multiple mode
> * fixed cancel_all
> * reduce nonce errors by sleeping for 5 seconds instead of 1, again, not ideal, but I found out this is a good sweet spot, we should instead solve this properly using [this](https://github.com/CoinAlpha/hummingbot/pull/1237#issuecomment-578280705) solution
> * other small updates

Thanks for your help, @nionis! Igor will look at these issues shortly.

PR #1267, opened 2020-01-02
(fix) external pricing source minor issues

Nullably:

After testing the new external pricing source feature, I came across a few issues.
- The normal pure mm strategy without external pricing source is not working. This was my fault as  I didn't fully test the whole features again after making changes to `OrderBookAssetPriceDelegate`. But it also shows that the compiler didn't catch this nor our current unittest cases (as it relies too much on mock objects).
- Update version number on the config template file.


dennisocana:
Tested and confirmed fix.

PR #1271, opened 2020-01-03
Fix for stopping then starting the network

Arctek:

Fix for stopping then starting the network, so it correctly stops and restarts the block watcher as well as the other Ethereum watchers.
At the moment it leaves the web3 wallet backend in a semi-broken state.


martinkou:
ok to test

Issue #1261, opened 2020-01-01
[BUG] [Windows] Windows Time calculation is inaccurate which leads to nonce issues which leads to orders not being tracked properly


Nullably:
**Describe the bug**
Hummingbot keeps on issuing maker orders non stop on Binance pure market making strategy, multiple orders mode. Numbers of open orders reached over 50 in a few minutes. Stop command cancels just one order, left the rest open on the exchange.
I found this bug while investigating another issue #1247 , not sure if they are related, will investigate further.

Here are the steps to reproduce the issue (see attachments in the section below):
Starting balance: 
ETH: 1
LINK: 0

Configure the bot as followings:
```
maker_market: binance
primary_market_trading_pair: LINKETH
mode: multiple
cancel_order_wait_time: 20.0
number_of_orders: 5
order_start_size: 1.0
```
The rest of configs are set to default
Start the bot

**Screenshots**
![image](https://user-images.githubusercontent.com/37262506/71641016-e0d30680-2ccf-11ea-8e74-3352975d8743.png)

Open orders on Binance reached over 50 within a few minutes.
![image](https://user-images.githubusercontent.com/37262506/71641030-124bd200-2cd0-11ea-9cde-edcf93d3cd14.png)

**Release version**
0.21.0
Same problem on Development branch

**Optional: your discord username: Jack | Coinalpha

Nullably:
Here is the log file.
[logs_conf_pure_market_making_strategy_8.log](https://github.com/CoinAlpha/hummingbot/files/4017888/logs_conf_pure_market_making_strategy_8.log)

Nullably:
What happened is the `tracking_nonce used` for `order_id` returns the same value for multiple orders issued almost simultaneously. If this bug is not present on other system, it might be a Windows thing where time.time() is returned within a micro (or lower) second precision.

Once I added a function to check for duplicates as below, the problem solved.
```
cdef int64_t c_last_tracing_nonce(self):
        cdef int64_t tracking_nonce = <int64_t>(time.time() * 1e6)
        self._last_tracking_nonce = tracking_nonce if tracking_nonce > self._last_tracking_nonce else \
            self._last_tracking_nonce + 1
        return self._last_tracking_nonce
```

dennisocana:
@veecos2 confirmed that this happens on Hummingbot installed via binary but not when running in Linux cloud server. Here's a recording of the behavior in Windows.

[...]

PR #1290, opened 2020-01-07
Fix/multiple orders id duplicate

Nullably:

**Optional**:
- Related Github issue:
- Related Clubhouse Story: #1261 

- Create a module for tracking_nonce to ensure its return value is unique, this is necessary due to time.time() resolution on Windows is not enough to create unique order_id 
- Update markets to use this module when creating new order_id
- I have only tested this binance market (as I don't have accounts on other exchanges), the change is minor, I don't expect any issue on other exchanges.


hummingbot/market/bamboo_relay/bamboo_relay_market.pyx
```diff
@@ -1315,7 +1316,7 @@ cdef class BambooRelayMarket(MarketBase):
                    object price=s_decimal_NaN,
                    dict kwargs={}):
         cdef:
-            int64_t tracking_nonce = <int64_t>(time.time() * 1e6)
+            int64_t tracking_nonce = get_tracking_nonce()
```
chowryan: i think`<int64_t>` casts the python `int` to cython `int64_t` so we might want to keep it

hummingbot/core/utils/tracking_nonce.py
```diff
@@ -0,0 +1,12 @@
+import time
+
+_last_tracking_nonce: int = 0
+
+
+# This tracking nonce is needed bc resolution for time.time() on Windows is very low (16ms),
+# not good enough to create unique order_id.
+def get_tracking_nonce() -> int:
+    global _last_tracking_nonce
+    nonce = (time.time() * 1e6)
```
chowryan: this is a float

hummingbot/market/binance/binance_market.pxd
```diff
@@ -1,9 +1,6 @@
-from libc.stdint cimport int64_t
-
 from hummingbot.market.market_base cimport MarketBase
 from hummingbot.core.data_type.transaction_tracker cimport TransactionTracker
 
-
```
chowryan: our pylint rules check for 2 lines before new class.

chowryan:
minor nitpicks, LGTM

hummingbot/core/utils/tracking_nonce.py
Nullably: yep, i'll fix this.

hummingbot/market/bamboo_relay/bamboo_relay_market.pyx
Nullably: i removed the cast as I was worried about performance, after some researching I don't think it's matter much whether to have explicit casting or implicit casting. I'll add it back it.

PR #1296, opened 2020-01-09
feat / first API mock prototype for feedbacks

Nullably:

**This is a prototype for reviewing and getting feedback only, should not  be merged to the development**

**Optional**:
- Related Github issue:
- Related Clubhouse Story: 6494

- Create decorator to mock aiohttp and normal http requests.
- To be used in unittesting which can be switched to either to let requests pass through normally or to mock them with our own responses
- This used external library aresponses, which can be installed by pip install aresponses. aresponses mocks responses by creating response mock server which captures all aiohttp requests.

The limitation of aresponses that I have experienced so far
- It captures all aiohttp requests, thus all responses have to be prepared even ones that might not be needed for the test case (e.g. trading_pair_fetcher makes requests to all exchanges)
- There is pass_through method in the library but it not quite clear on how to use it.
- Responses which have been mapped to the request will be removed from the server, this might not work for market connector which keeps checking network status.


test/integration/test_liquid_market_mock.py
```diff
@@ -0,0 +1,198 @@
+#!/usr/bin/env python
+from hummingbot.market.liquid.liquid_api_order_book_data_source import LiquidAPIOrderBookDataSource
```
martinkou: Don't import anything before the `sys.path.insert()` line... you cannot access the `hummingbot` module here yet.

martinkou:
I tried to run the tests after moving the `liquid_api_order_book_data_source` import line, and found the tests to error out at a `_stop() not found from None` error. But it looks like it's because the `LiquidMarketMockUnitTest` isn't complete and isn't written in the intended way yet.

I was able to run `TestMockAioResponse` and `SimpleTest` successfully.

I'd like to discuss with you on how the `__aenter__()` and `__aexit__()` work in the decorators.

martinkou:
Talked w/ Jack today. The `aresponses` framework seems like to be the right direction to go, but there're some problems.

We should investigate deeper into how `aresponses` manages to redirect requests from both `aiohttp` and `requests` - and see if there're ways we can use the underlying mock mechanism directly, without triggering other problems in `aresponses`.

Nullably:
@martinkou 
I have updated this PR to include new files 
- `response_mock_server.py` based mostly on `aresponses` (though it still has problems when many requests are generated, and the tests on `test_liquid_market_mock_server.py` still not passed).
- `humming_web_app.py` based on your code `JenkinsWebApp`, see `test_liquid_market_mock_web_app.py` for this works

I think `humming_web_app` is the way going forward, here are how it works:
- It creates a local server to handle all requests coming through the available port.
- The response is given based on the host, path, method and query string of the original request.
- The response data can be configured even after the server has started.
- Certain requests are redirected to the local server by mocking `aiohttp.client.URL` (this is where a URL object is created for all aiohttp requests) see `HummingWebApp.reroute_local`
-  `test_liquid_market_mock_web_app.py` is based on 'test_liquid_market.py` with a switch to either test with real api cals or mocked ones. It has 2 tests to demonstrate many api calls including buy and sell tests, all tests here pass successfully on my local.

PR #1323, opened 2020-01-16
Fix / fetching nonexisting orders (coinbase pro and liquid)

kmcgowan2000:

**Optional**:
- Related Github issue: 1281
- Related Clubhouse Story: 6969

Fix the attempted fetching of nonexisting orders in **_coinbase pro_** and **_liquid_**.
- if Hummingbot attempts to update status of a tracked nonexisting order, it will now remove it from tracking
- if Hummingbot attempts to delete a nonexisting order, it will now correctly detect that it already does not exist and consider it a successful delete


martinkou:
Has the patch been tested on both exchanges? Is it true that the "order not found" error message is similar to both exchanges?

kmcgowan2000:
> Has the patch been tested on both exchanges? Is it true that the "order not found" error message is similar to both exchanges?

Yes it has been tested on both exchanges. As it was before last push (fix case of err message), it successfully fixed the bug as tested by duplicating sqlite file with open orders and importing again once closed. Tested again with modified exchange_order_id to force a 404, and now have fixed the error message on Liquid (it is "Order not found" as opposed to CBP's "order not found").

hummingbot/market/coinbase_pro/coinbase_pro_market.pyx
```diff
@@ -447,13 +447,31 @@ cdef class CoinbaseProMarket(MarketBase):
...
+                        self.c_trigger_event(
+                            self.MARKET_ORDER_CANCELLED_EVENT_TAG,
+                            OrderCancelledEvent(self._current_timestamp, client_order_id)
+                        )
+                except asyncio.CancelledError:
+                    raise
+                except Exception as e:
+                    self.logger().network(
+                        f"Error fetching status update for the order {client_order_id}: "
+                        f"{type(e)} {str(e)}.",
```
veecos2: It is better to add exc_info = True here.

hummingbot/market/liquid/liquid_market.pyx
```diff
@@ -607,13 +607,31 @@ cdef class LiquidMarket(MarketBase):
...
+                        self.c_trigger_event(
+                            self.MARKET_ORDER_CANCELLED_EVENT_TAG,
+                            OrderCancelledEvent(self._current_timestamp, client_order_id)
+                        )
+                except asyncio.CancelledError:
+                    raise
+                except Exception as e:
+                    self.logger().network(
+                        f"Error fetching status update for the order {client_order_id}: "
+                        f"{type(e)} {str(e)}.",
```
veecos2: It is better to add exc_info = True here.

hummingbot/market/coinbase_pro/coinbase_pro_market.pyx
```diff
@@ -879,9 +896,14 @@ cdef class CoinbaseProMarket(MarketBase):
                     if type(client_order_id) is str:
                         order_id_set.remove(client_order_id)
                         successful_cancellations.append(CancellationResult(client_order_id, True))
-        except Exception:
+                    else:
+                        self.logger().info(
+                            f"failed to cancel order with error: "
```
veecos2: You are using a different log level for this in liquid and coinbase. Could you use the same level ?. Also not sure if logging the type of client order id is needed here.

kmcgowan2000: Ready to push the other changes but I think logging the type of client_order_id is useful here only when it is an Exception. I'm going to change it to repr(client_order_id) so that if it is an Exception it displays the Exception class as well as the contents.

PR #1337, opened 2020-01-22
Feat/api mock tests

Nullably:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:6494

- Create a local host web server to handle (mock) api requests
- aiohttp requests are redirected to local by mocking URL of yarl 
- Liquid market test cases have been updated to either be tested by api mocking or without
- MOCK_API_ENABLED environment variable needs to be created to configure the mock


test/integration/humming_web_app.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def start(self):
+        if self.started:
+            self.stop()
+        thread = Thread(target=self._start_web_app)
+        thread.daemon = True
+        thread.start()
+
+    def stop(self):
+        asyncio.ensure_future(self._stop())
```
martinkou: `ensure_future()` is no longer sufficient here considering the event loop is now in another thread.

Instead, you should use
```
asyncio.run_coroutine_threadsafe(sel.f_stop(), self._ev_loop)
```

test/integration/humming_web_app.py
```diff
@@ -0,0 +1,226 @@
...
+            await asyncio.sleep(0.1)
+
+    async def _stop(self):
+        if self._runner is None:
+            return
+        await self._runner.cleanup()
+        self._runner = None
+        self._impl = None
+        self._port = None
+        self._started = False
```
martinkou: You should also call `self._ev_loop.stop()` here, to stop the event loop and end the wrapping thread.

test/integration/humming_web_app.py
Nullably: Done.

test/integration/humming_web_app.py
Nullably: Done.

PR #1338, opened 2020-01-23
Feat/handle crossed orders on Bamboo Relay

kmcgowan2000:

**Optional**:
- Related Github issue:
- Related Clubhouse Story: 7060

- Change the way OrderBook deals with crossed orders if dex flag is True.
- Set BambooRelayOrderBook to have dex=True.

When the order book is crossed over, e.g. best bid > best ask, compare the values (amount * price) of the top orders on both sides of the order book. Erase the order that is smaller in value, and continue comparing and erasing until the two sides no longer cross.


martinkou:
Need unit test cases.

kmcgowan2000:
> Need unit test cases.
I did some in a web cpp shell so I could easily share:
http://cpp.sh/6gbws

PR #1346, opened 2020-01-24
Feat/bamboo relay api order updates

Arctek:

Refactoring/updates to how the connector tracks orders and fills.
Especially for orders that may be soft-cancelled - but then also filled within the 90 second / order expiry boundary.

Additionally this adds a new watcher ZeroExFillWatcher, which checks for orders that may have been mined in logs.
This adds some safety in the case the API may go down but orders are still filled.

I've also added the eth-bloom package, and updated the other ethereum watchers. This means they can first check the `logsBloom` bloom filter against each block before speculatively requesting logs.
As bloom filters only have false positives not negatives, so in theory this should reduce some of the RPC demands of hummingbot.


carlolm:
OK to test

yzhang1994:
@Arctek I also don't think `ZeroExFillWatcher` is included in your PR. Did you forget to commit the file?

Arctek:
> 
> 
> @Arctek I also don't think `ZeroExFillWatcher` is included in your PR. Did you forget to commit the file?

Ugh, you're right that's my bad sorry!
Although, how are the tests passing on jenkins without this?

I've added it now anyway, and updated the setup/environment-*.yml files to include the `eth-bloom` library.

yzhang1994:
@Arctek I tested this branch and it seems to be working OK.

I have a question though. In the BambooRelayOrderBookTracker line 235, the tracker applies a diff as a snapshot. Is there a reason why this is happening?

This is important because we just added a handler so that orders on DEXes are never crossed. (https://github.com/CoinAlpha/hummingbot/pull/1354) 

However, if the Bamboo diffs don't go through apply_diff, then the handler is not used. 
image

Arctek:
> 
> 
> @Arctek I tested this branch and it seems to be working OK.
> 
> I have a question though. In the BambooRelayOrderBookTracker line 235, the tracker applies a diff as a snapshot. Is there a reason why this is happening?
> 
> This is important because we just added a handler so that orders on DEXes are never crossed. (#1354)
> 
> However, if the Bamboo diffs don't go through apply_diff, then the handler is not used.
> image

I changed this some time ago, the other month - see https://github.com/CoinAlpha/hummingbot/pull/1183#discussion_r357450121.
Because what was happening - not sure if it is still the case - but the apply_diff was completely clearing the entire book on both sides and only adding the changed order.
Also the API websocket also doesn't really give out diffs as the messages that come through here are single order updates, i.e. FILL, REMOVE etc. (This is the case for Radar API as well)

Is there any reason the code in #1354 doesn't apply to the entire snapshot as well for dexes?
If the book is crossed then it's crossed on the site as well and will be on every snapshot as well, which is refreshed every ~10 seconds anyway?

yzhang1994:
@Arctek So are not supposed to allow orders to cross in Hummingbot, even if it is the case on the site.  When an order book is crossed and running pure market making, the bot will not trade at all when the spread is tight, because it thinks it is creating a taker order. This is more of an issue now since we are going to launch the 0x campaign on Bamboo. The max spread allowed by a market maker is 0.6%. Let me know if that is unclear / can continue chatting on discord.  Right now I'm trying to debug and find why `apply_snapshot` doesn't automatically erase crossed orders, which it is supposed to.

Arctek:
Let me know if it's my PR causing the segfault - I'll need to fix it otherwise when I get back though.

PR #1355, opened 2020-01-28
docs / troubleshooting

dennisocana:

Troubleshooting section
- Added steps how to reset in case of forgotten password
- Added information how to transfer files from/to WSL and local Windows computer
- Added steps how to download a previous version of Hummingbot on Windows via binary
- Minor edits to some existing steps

Cross exchange market making
- Added minimum order size under prerequisites

Liquid
- Added instructions how to get minimum order size of newly listed tokens through public API.

Others
- Started to remove lengthy docker commands from the user manual to encourage users especially non-technical, in using the automated scripts instead.
- Removed note of setting paper trade to disabled when running Discovery strategy after we confirmed that there should be no problems when doing both at the same time.


documentation/docs/connectors/liquid.md
```diff
@@ -36,6 +36,37 @@ This article below in their website under API information shows step-by-step ins
 
 There is no minimum order quantity for trading fiat currency. See [this page](https://help.liquid.com/en/articles/3339119-minimum-order-quantity) for minimum order sizes on crypto trading pairs.
 
+If the token or trading pair isn't listed in the article, you can also get this information through their public API.
+
+```
+https://api.liquid.com/currencies
+```
+
+The minimum order size is the value next to `minimum_order_quantity`. In the example below, the minimum order size of IDRT is `500.0`.
```
carlolm: Can you use another example, ideally any of BTC, ETH, USDC, USDT

dennisocana: Changed to ETH as an example and specified if ETH is not listed in the article for minimum order sizes then will use the public API.

https://github.com/CoinAlpha/hummingbot/pull/1355/commits/356987c85dcb631c5fea3a6e582d15e3239efe75

PR #1373, opened 2020-02-01
refactor/Liquid connector

paleboot:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:

- changed logger info in get_tracking_pairs from {index*1} to {index+1} to avoid user confusion
- changed nonce in Liquid authentication from seconds to milliseconds as recommended by Liquid through their API Docs (https://developers.liquid.com/#authentication)


hummingbot/market/liquid/liquid_auth.py
```diff
@@ -21,7 +21,7 @@ def generate_signature(self, path_url: str) -> (Dict[str, Any]):
         """
         auth_payload = {
             'path': path_url,
-            'nonce': str(time.time()),
+            'nonce': str(time.time() * 1e3),
```
zeitgeistf: just realized what both before and after will have decimals.
e.g. 
old ts: `1581314372.6992035`
new ts: `1581314372699.209`

How about let's remove the decimal points similar to the suggested format from the API doc by casting the expression to int, something along this line? `str(int(time.time() * 1e3))`, what do you think? @paleboot

zeitgeistf: reference: https://developers.liquid.com/#authentication
`str(int(time.time() * 1e3))` => `1581314372699`

hummingbot/market/liquid/liquid_api_order_book_data_source.py
```diff
@@ -233,8 +233,8 @@ def filter_market_data(cls, exchange_markets_data) -> (List[dict]):
 
                     retval[trading_pair] = LiquidOrderBookTrackerEntry(trading_pair, snapshot_timestamp, order_book)
 
-                    self.logger().info(f"Initialized order book for {trading_pair}."
-                                       f"{index*1}/{number_of_pairs} completed")
+                    self.logger().info(f"Initialized order book for {trading_pair}. "
+                                       f"{index+1}/{number_of_pairs} completed")
```
zeitgeistf: good catch!

hummingbot/market/liquid/liquid_auth.py
paleboot: Hi, thanks for checking out my PR! That's actually a great idea, totally missed on that one. I made the change, please do check it out!

hummingbot/market/liquid/liquid_api_order_book_data_source.py
paleboot: Thanks!

carlolm:
OK to test

carlolm:
@Nullably can you approve if this is OK to merge now?

PR #1422, opened 2020-02-14
[Testing] Fix/Seg Fault for Hummingbot running on Bamboo

veecos2:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:


hummingbot/core/cpp/OrderBookEntry.cpp
```diff
@@ -40,16 +40,16 @@ void truncateOverlapEntries(std::set<OrderBookEntry> &bidBook, std::set<OrderBoo
...
         const OrderBookEntry& topBid = *bidIterator;
         const OrderBookEntry& topAsk = *askIterator;
         if (topBid.price >= topAsk.price) {
             if (topBid.amount*topBid.price > topAsk.amount*topAsk.price) {
                 askBook.erase(askIterator++);
             } else {
                 std::set<OrderBookEntry>::iterator eraseIterator = (std::next(bidIterator)).base();
-                bidIterator++;
                 bidBook.erase(eraseIterator);
+                bidIterator++;
```
chowryan: does this same change need to be done to `truncateOverlapEntriesCentralised`?

veecos2: It does not affect, but I changed it to maintain the format across.

hummingbot/core/cpp/OrderBookEntry.cpp
```diff
@@ -60,16 +60,16 @@ void truncateOverlapEntriesDex(std::set<OrderBookEntry> &bidBook, std::set<Order
...
         const OrderBookEntry& topBid = *bidIterator;
         const OrderBookEntry& topAsk = *askIterator;
         if (topBid.price >= topAsk.price) {
             if (topBid.updateId > topAsk.updateId) {
                 askBook.erase(askIterator++);
             } else {
                 std::set<OrderBookEntry>::iterator eraseIterator = (std::next(bidIterator)).base();
-                bidIterator++;
                 bidBook.erase(eraseIterator);
+                bidIterator++;
```
martinkou: Moving the `bidIterator++` after the erase looks wrong to me... basically it means you're advancing an erased iterator - and it should cause even more segfaults than before. Are you sure this is correct?

martinkou:
Is there an explanation on why would the proposed change fix the segfault?

If it's really the order book C++ code causing the issue, e.g. when one side of the order book is empty - it should be possible to produce an isolated test case in pure C++, and catch the memory violation with something like valgrind.

veecos2:
Will test this a bit more to see if this fixes the issue.

veecos2:
On further testing, this does not seem to fix the issue. Closing for now.

martinkou:
I just did some valgrind tests on my branch martin_kou/test_segfault today. I think you actually hit the right fix previously, perhaps by accident.

Turns out reversing `bidBook.erase(eraseIterator);` and `bidIterator++;` was the right fix. It doesn't seem to make sense on the first glance because it looks like you're incrementing something that has been erased. But it actually was correct to erase first and increment next because the reverse iterator holds a base (forward) iterator that's the next element on the reverse order.

Thus, ironically, if we increment the reverse iterator first and then delete the original iterator - the incremented reverse iterator would be holding a deleted iterator and thus will cause invalid memory accesses.

martinkou:
On the other hand, the empty check on the while statement has no effect, because `begin() == end()` and `rbegin() == rend()` when the collection is empty in C++.

veecos2:
I removed the iterator and tested it and that led to segfault so I thought this did not fix it.

martinkou:
Also... for these micro memory violations in C++... the most efficient way is to write micro tests that only stress the functions in question (i.e. the truncate functions), and run them with `valgrind` in Linux.

The micro memory violations often don't cause segfaults on the first few runs, which makes them hard to discover. But with `valgrind`, any out-of-bound access will trigger error logs.

martinkou:
One more change needed... actually the `bidIterator++` is not needed upon further testing. The reason is... the `bidIterator`, being the first element on the reverse iterator sequence... should always have a `base()` equal to `end()`.

Reference: https://en.cppreference.com/w/cpp/iterator/reverse_iterator

Doing an increment to it while deleting the front actually causes the truncate logic to check and truncate the wrong bid entries. The correct thing to do is to actually just keep deleting the forward iterator just before the `end()` for the bids book, and keep the bid iterator constant.

It's counter-intuitive at first, but after reading through the C++ reverse iterator docs and running through test cases, it's correct.

martinkou:
I'm gonna submit my own PR for this, which includes the C++ test cases and comments on why the bid iterator doesn't need to be advanced.

Issue #1388, opened 2020-02-06
[BUG] Binance clock desync in Windows results to problems with orders


Howie620:
**Describe the bug**
// A clear and concise description of what the bug is.
When I was running the bot on Binance yesterday, I saw 2 bids and 1 offer outstanding in single order mode and I have to manually cancel an outstanding order. 

**Steps To Reproduce**
// A concise description of the **steps to reproduce the buggy behavior**:
After I run the bot for a while, I have this issue. 

Here are the steps to reproduce the issue (see attachments in the section below):
1. Set up your configuration file with these parameters: ...
2. Start the bot using this command: ...
... Here is my config file for reference: 

**Screenshots**
// If applicable, add screenshots to help explain your problem.
![IMG_5448](https://user-images.githubusercontent.com/1378301/73900452-c42b7f00-48ca-11ea-8469-1c8d0b94e949.PNG)
![IMG_5447](https://user-images.githubusercontent.com/1378301/73900458-c8f03300-48ca-11ea-9626-efb832807c47.PNG)
As you can see in the screenshot and look closer at the timestamp, I still have this order outstanding. The bot is running correctly as there are other orders executed recently. 

**Release version**
// Include your bot's version number (Can be found at the upper left corner of your CLI).
0.23.0

**Attachments**
// Please attach your config file and log file to help us debug. It would be difficult for us to help you without those! 

[conf_pure_market_making_strategy_zil_eth.yml.txt](https://github.com/CoinAlpha/hummingbot/files/4162913/conf_pure_market_making_strategy_zil_eth.yml.txt)
[logs_conf_pure_market_making_strategy_zil_eth.log.txt](https://github.com/CoinAlpha/hummingbot/files/4162915/logs_conf_pure_market_making_strategy_zil_eth.log.txt)

*WARNING: Do NOT publish any exchange API keys or your wallet's private key. Whoever has access to them may steal your assets!*

**Optional: your discord username:** Howie#7291
**Optional: your ETH address:** (If you want to participate in the bug bounty)

Hashira-Lars:
I've sent you a dm @Howie620 to follow up things faster.we can see from the logs one of the errors your having is with timestamp.

```binance.exceptions.BinanceAPIException: APIError(code=-1021): Timestamp for this request was 1000ms ahead of the server's time.```

Hashira-Lars:
Cannot reproduce behavior in my end running for 2days windows binary.

[...]

PR #1437, opened 2020-02-20
Fix Binance timestamp

Nullably:

**Optional**:
- Related Github issue: #1388
- Related Clubhouse Story: 7455

- Updated `BinanceTime` to no longer store a list of time offsets, only the latest one. 
- Updated `BinanceMarket` to set time offset as soon as the timestamp error happens.

### DIFF
--- a/hummingbot/market/binance/binance_time.py
+++ b/hummingbot/market/binance/binance_time.py
@@ -1,9 +1,7 @@
 import aiohttp
 import asyncio
 import logging
-import statistics
 import time
-from collections import deque
 from hummingbot.logger import HummingbotLogger
 from hummingbot.core.utils.async_utils import safe_ensure_future
 
@@ -30,11 +28,10 @@ class BinanceTime:
         return cls._bt_shared_instance
 
     def __init__(self, check_interval: float = 60.0):
-        self._time_offset_ms = deque([])
+        self._time_offset_ms = 0.0
         self._set_server_time_offset_task = None
         self._started = False
         self.SERVER_TIME_OFFSET_CHECK_INTERVAL = check_interval
-        self.median_window = 100
 
     @property
     def started(self):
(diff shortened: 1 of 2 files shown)


hummingbot/market/binance/binance_time.py
```diff
@@ -30,60 +28,52 @@ def get_instance(cls) -> "BinanceTime":
...
 
     @property
     def started(self):
         return self._started
 
     @property
     def time_offset_ms(self):
-        if not self._time_offset_ms:
-            return 0.0
-        return statistics.median(self._time_offset_ms)
```
martinkou: The statistical error method is still needed, to tackle inconsistent network latencies.

Instead of removing the stats mechanism to tackle system clock getting set forward or backward... We can use monotonic clocks for local clock reference instead.

I.e. instead of `time.time()`, we can use `time.perf_counter()` instead.

martinkou:
As discussed with @Nullably over video conf, here're the changes we agreed upon

 1. Keep the statistics deque, but switch the local reference clock from `time.time()` to `time.perf_counter()` to prevent local clock changes from affecting the Binance timestamps
    * This means a minor change to the `time_offset_ms` property function to account for the case where there isn't any samples in `_time_offset_ms` as well.. since we can't just output `time.perf_counter()` as a timestamp directly.
 2. Decrease the `_time_offset_ms` window size from 100 to 5, s.t. it can respond faster to changes to the Internet connection.
 3. Keep the `set_server_time_offset()` exception mechanism when Binance tells us the timestamp it got from us looks bad.

Nullably:
@martinkou 
time.time() is the time since the epoch whereas time.perf_counter() is probably time since the process starts (actually its reference point is undefined according the docs), e.g.
```
>>> import time
>>> time.time()
1582254419.0332952
>>> time.perf_counter()
15.175606982
```
So they cannot be used interchangeably, we can use the performance counter for when calculating `time_offset` to account for time spent query the Binance time api.

martinkou:
@Nullably they can be used interchangeably in our context - because we just need a local reference time + an offset to get the true time, not the true time.

Let's say Binance's time is 1000, and local time is 999, and the performance counter is 0.

If you use `time.time()`, then you get an offset of 1. So when you call `time()`, it'll calculate local ref + offset = 999 + 1 = 1000.

If you use `time.perf_counter()`, then you get an offset of 1000. So when you call `time()`, it'll calculate local ref + offset = 0 + 1000 = 1000. Same answer.

martinkou:
It doesn't matter where the clock starts from when we get a local reference time, as long as that reference time advanced by 1 for every 1 second in the real world. The local reference can be something strange like how long have passed since you started your computer for example. It doesn't matter at all.

The reason is, we're always calculating an offset from the true Binance time from it. And in case we haven't got a sample from Binance to calculate the diff, then we use `time.time() - local reference` instead.

martinkou:
I've created #1444, which improves upon this one.

Nullably:
@martinkou 
Ok, got it, you're right. The offset will account for difference, so it doesn't matter what perf_counter's reference is (as long as it is a constant). I updated the PR.

Nullably:
Sorry, just saw your previous comment.

PR #1444, opened 2020-02-21
fix / binance timestamp 2

martinkou:

**Optional**:
- Related Github issue: #1388 
- Related Clubhouse Story: 7455

This is an improved version of the Binance timestamp fix, built upon #1437. It has the following improvements

* Changed the time offset semantics to use `time.perf_counter()` as the local reference, instead of `time.time()` - to avoid problems with changing real-time clocks.
* Improved the exception catching in `binance_market.pyx` - apparently Binance has a few similar error messages for invalid timestamps, not all of which involve "1000ms ahead of the server".
* Fixed a systemic error in how the before-request time reference is recorded in Jack's patch... it must be before `session.get()` because the `get()` call is part of the roundtrip.
* Added more type hints and improved function / variable naming to make things easier to understand.
* Updated unit test case.

### DIFF
--- a/hummingbot/market/binance/binance_time.py
+++ b/hummingbot/market/binance/binance_time.py
@@ -1,9 +1,11 @@
 import aiohttp
 import asyncio
+from collections import deque
 import logging
 import statistics
 import time
-from collections import deque
+from typing import Dict, Deque
+
 from hummingbot.logger import HummingbotLogger
 from hummingbot.core.utils.async_utils import safe_ensure_future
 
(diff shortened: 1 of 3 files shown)


hummingbot/market/binance/binance_time.py
```diff
@@ -66,24 +71,25 @@ def stop(self):
...
-                raise
-            except Exception:
-                self.logger().network(f"Error getting Binance server time.", exc_info=True,
-                                      app_warning_msg=f"Could not refresh Binance server time. "
-                                                      f"Check network connection.")
+            await self.update_server_time_offset()
+            await asyncio.sleep(self._server_time_offset_check_interval)
 
-            await asyncio.sleep(self.SERVER_TIME_OFFSET_CHECK_INTERVAL)
+    async def update_server_time_offset(self):
```
Nullably: This function will likely get called too many times within a short window when we get TimeStamp errors burst. You can add last_updated_timestamp and check if it is still within, says, 5 seconds, just return.

test/integration/test_binance_market.py
```diff
@@ -533,23 +533,25 @@ def test_order_price_precision(self):
 
     def test_server_time_offset(self):
```
Nullably: I have rewritten this test case when working on the fix. I find it more relevant and easier to understand.

    def test_server_time_offset(self):
        orig_time = time.time

        def delayed_time():
            return orig_time() - 30.0

        time_obj: BinanceTime = binance_client_module.time
        old_check_interval: float = time_obj.SERVER_TIME_OFFSET_CHECK_INTERVAL
        time_obj.SERVER_TIME_OFFSET_CHECK_INTERVAL = 1.0
        try:
            with patch("time.time", autospec=True, side_effect=delayed_time):
                time_obj.stop()
                # Check to make sure the clock on binance_time is delayed.
                self.assertGreaterEqual(orig_time() - time_obj.time(), 25)
                time_obj.start()
                # There is a check on set_server_time_offset for 5 seconds, got to wait here a bit
                self.run_parallel(asyncio.sleep(6))
                self.run_parallel(time_obj.set_server_time_offset())
                print(f"bin_timeoffset: {time_obj._time_offset_ms}")
                # Assuming local system clock is no more than 2 seconds different to Binance server time.
                self.assertLessEqual(abs(time_obj.time() - orig_time()), 2)
        finally:
            time_obj.SERVER_TIME_OFFSET_CHECK_INTERVAL = old_check_interval
            time_obj.stop()
            time_obj.start()

PR #1495, opened 2020-03-04
fix / coin gecko hotfix 2

Nullably:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:

Another hot fix for Coin Gecko, since they no longer applied the 70 ids restriction on the API, we can use it like how we did before at 500. Not doing so causes some tokens (e.g. ONE) to be missing from the prices. 
This is a temporary fix, a better solution is in development.


mifeng:
LGTM. I don't think we should issue another hotfix since we just did one, but if users complain about liquidity mining for ONE pairs due to this issue, we can point them to the `development` branch.

PR #1503, opened 2020-03-08
fix / pt memory leak

martinkou:

**Optional**:
- Related Github issue:
https://github.com/CoinAlpha/hummingbot/issues/1497
- Related Clubhouse Story:
https://app.clubhouse.io/coinalpha/story/7355/possible-bug-bots-hang-after-a-while-on-smaller-aws-instances

- This patch fixes three major memory leaks: 
  1. Python 3.6.x with aiohttp: https://bugs.python.org/issue34745
  2. prompt-toolkit with async: https://github.com/prompt-toolkit/python-prompt-toolkit/issues/1057
  3. A tasks leak in order book trackers, in which tracker tasks are started with the market connector's `start_network()` but they are not properly stopped with `stop_network()`.
- Upgraded Python to Python 3.8.2 to fix the first leak type
- Upgraded prompt-toolkit to v3.0.3 to fix the second leak type
- Refactored the `start()` and `stop()` functions in all order book trackers back to the old, original semantics, which is free from task leak problems.


setup/environment-linux.yml
```diff
@@ -1,148 +1,157 @@
...
     - pyperclip==1.7.0
-    - pyrsistent==0.15.2
+    - pyrsistent==0.15.7
     - python-binance==0.7.1
-    - python-telegram-bot==11.1.0
-    - regex==2019.2.21
-    - requests==2.21.0
-    - rlp==1.1.0
-    - ruamel-yaml==0.15.89
+    - python-graphviz==0.13.2
```
carlolm: @martinkou 
I think this should be: `graphviz==0.13.2` instead of `python-graphviz==0.13.2`


image

carlolm: That's what's causing the jenkins build to fail

martinkou:
Note: This patch has only reduced two major sources of memory leak, there are still two more leak sources at the moment:

 1. The `EventLogger` in `MarketBase`.  It records all the order created / cancelled / filled events from active trading without bound. We should create an `EventLogger` subclass that logs to SQLite instead. @yzhang1994 
 2. There seems to be a leak in `BinanceAPIOrderBookDataSource.listen_for_order_book_snapshots()`, where old order book snapshots are not freed and accumulate in memory. The cause is unknown. I'm still investigating.

martinkou:
I've fixed the order book tracker task leak problem.

martinkou:
I've also (tentatively) confirmed, that the only leak left is now the `EventLogger` leak.

```
>>> objgraph.growth()
[('OrderCancelledEvent', 180, 12), ('BuyOrderCreatedEvent', 99, 6), ('SellOrderCreatedEvent', 90, 6), ('OrderBookMessage', 66, 1)]
(after half an hour)
>>> objgraph.growth()
[('OrderCancelledEvent', 306, 126), ('BuyOrderCreatedEvent', 162, 63), ('SellOrderCreatedEvent', 153, 63)]
```

The event messages are relatively small compared to all the extra snapshot / diff messages being fetched from the task leak. We may be able to release the fix as-is and leave the `EventLogger` leak for later.

hummingbot/core/network_base.py
```diff
@@ -120,7 +119,7 @@ def check_network_timeout(self, timeout):
                 self.logger().error("Unexpected error starting or stopping network.", exc_info=True)
 
     def start(self):
-        self._check_network_task = safe_ensure_future(self._check_network_loop())
+        self._check_network_task = asyncio.ensure_future(self._check_network_loop())
```
yzhang1994: Why don't want to use `safe_ensure_future` here?

martinkou: Oh that was a mistake. I was trying to do a diff test to see if `safe_ensure_future()` caused memory leak.

martinkou: I'm going to create another PR to fix that.

PR #1510, opened 2020-03-09
fix/vwap formular

zhutianchi-mcarlo:

The formular in order_book.c_get_vwap_for_volume is wrong.
The formular looks like:

vwap
[https://en.wikipedia.org/wiki/Volume-weighted_average_price](https://en.wikipedia.org/wiki/Volume-weighted_average_price)

a counter example of the original code: calling
```
c_get_vwap_for_volume(is_buy=True, volume=10)
```
while the orderbook.asks are
|  price  |  amount  | comment |
|  ----  | ----  | ----  |
| 9  | 9 | take 10 - 7 = 3 |
| 7  | 7 | take 7 |

so that the vwap should be
```
7 * 7 + 9 * 3
------------- = 7.6
     10
```

but the original code returns
```
7 * 7 + 9 * 6
------------- = 7.9
    7 + 6
```


martinkou:
Good catch. Thanks for fixing this for us.

martinkou:
ok to test

PR #1531, opened 2020-03-16
Feat / to track hanging orders and new cancel_hanging_order_pct

Nullably:

**Optional**:
- Related Github issue: 
- Related Clubhouse Story: 7842

- Hanging orders are now tracked (as active orders), they will get cancel when users stop the bot or when its order price spread is greater than `cancel_hanging_order_pct`
- Add a new cancel_hanging_order_pct config
- Show a new Hang column on `status` command (active orders)


martinkou:
Please add unit test case for the hanging order tracking and the `cancel_hanging_order_pct` feature.

test/test_pure_market_making_v2.py
```diff
@@ -905,10 +903,9 @@ def test_prevent_cancellation_feature(self):
...
 
         # Ask is filled and due to delay is not replenished immediately
-        # Bid order is no longer tracked and is not in active bids
         self.clock.backtest_til(self.start_timestamp + 2 * self.clock_tick_size)
         self.assertEqual(1, len(self.maker_order_fill_logger.event_log))
-        self.assertEqual(0, len(self.prevent_cancel_strategy.active_bids))
+        self.assertEqual(1, len(self.prevent_cancel_strategy.active_bids))
         self.assertEqual(0, len(self.prevent_cancel_strategy.active_asks))
 
         # Orders are placed after replenish delay
```
martinkou: Need new test cases for the new behaviors.

mifeng:
Per discussion with @Nullably, limiting the max number of hanging orders would add a fair amount of complexity, so we will wait to get feedback on the hanging orders feature from users before making this additional change.

mifeng:
Tested this extensively with 2, 7, and 10+ hanging orders and didn't have an issue with the bot losing track of those hanging orders. Per discussion with @Nullably, we won't add the `max_hanging_orders` parameter since it adds a fair amount of complexity. Narrowing the `cancel_hanging_orders_pct` param does the same thing.

mifeng:
Looks great overall - please make these changes to the new `status`:
- Change **Order ID** values to last 4 chars only, so `sell-BTCUSDT-1584503410009055` becomes `9055`
- Insert a column after **Order ID** called **Type**: values are either `buy` or `sell`
- Change **Time Placed** to **Age**, and show the order's age as current time minus order creation time. For the displayed value, display the time only, like `03:50:10`

hummingbot/strategy/pure_market_making/pure_market_making_v2.pyx
```diff
@@ -639,39 +649,33 @@ cdef class PureMarketMakingStrategyV2(StrategyBase):
...
+        active_orders = self.market_info_to_active_orders.get(market_info, [])
+        active_buy_orders = [x.client_order_id for x in active_orders if x.is_buy]
+        active_sell_orders = [x.client_order_id for x in active_orders if not x.is_buy]
 
-            # if filled order is buy, adjust the cancel time for sell order
-            # For syncing buy and sell orders during order completed events
-            for other_order_id in active_sell_orders:
-                if other_order_id in self._time_to_cancel:
+        if self._enable_order_filled_stop_cancellation:
+            # If the filled order is a hanging order, do nothing
```
chowryan: not sure what this means? can you clarify this logic?

Nullably: It's a bit hard to read the code comparison here as GitHub outlines it a bit. confusingly. Essentially, it is the logic here is same as existing, if the order that is filled is one of the hanging orders, we don't need to do anything here.

Nullably: It's a bit hard to read the code comparison here as GitHub outlines it a bit. confusingly. Essentially, it is the logic here is same as existing, if the order that is filled is one of the hanging orders, we don't need to do anything here.

PR #1560, opened 2020-03-25
Refactor / all pure mm configs 

Nullably:

**Optional**:
- Related Github issue:
- Related Clubhouse Story: 7953

- Remove single/multiple order mode (multiple is when order_levels > 1)
- Remove `order_start_size` (use order_amount instead)
- Rename `number_of_orders` to `order_levels`, `order_interval_size` to 'order_level_spread`

### DIFF
--- a/hummingbot/strategy/pure_market_making/constant_multiple_spread_pricing_delegate.pxd
+++ b/hummingbot/strategy/pure_market_making/constant_multiple_spread_pricing_delegate.pxd
@@ -5,5 +5,5 @@ cdef class ConstantMultipleSpreadPricingDelegate(OrderPricingDelegate):
     cdef:
         object _bid_spread
         object _ask_spread
-        object _order_interval_size
-        int _number_of_orders
+        object _order_level_spread
+        int _order_levels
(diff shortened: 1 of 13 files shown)


Nullably:
@mifeng 
> Can we delete the code that only applies to the old single order mode, such as the constant_spread_pricing_delegate and constant_size_sizing_delegate files?

This is a very good question, in theory, all pricing and sizing delegate logic for `single` order mode should have already been covered by their multiple counterpart (when order_level = 1). We need to do a fair bit of testing, if we're gonna remove them, but in the long run we have less code to maintain (which is a good thing).

> Are staggered_multiple_size_sizing_delegate and constant_multiple_spread_pricing_delegate the size and price plugins, respectively? 

Yes

> Do you know why they are named differently?

I think that's because one of them need to be constant first before you can calculate the other one. We calculate the prices of orders first, then use them to calculate orders volume (to account for fees and trading rule) especially so for buy orders when fees are in either base asset (Binance) or quote asset (others).

Nullably:
@mifeng Shouldn't default for order_level_spread be 1 (for 1% as existing)?

PR #1580, opened 2020-04-02
Feat / new create and import command

Nullably:

**Optional**:
- Related Github issue:
- Related Clubhouse Story: 8019

- New Create and Import commands to create/import configuration file.
- config command no longer works, config single variable command still works

### DIFF
--- a/hummingbot/client/config/config_helpers.py
+++ b/hummingbot/client/config/config_helpers.py
@@ -161,13 +161,13 @@ def _merge_dicts(*args: Dict[str, ConfigVar]) -> OrderedDict:
     return result
 
 
-def get_strategy_config_map(strategy: str) -> Optional[Dict[str, ConfigVar]]:
+def get_strategy_config_map(strategy: str = None) -> Optional[Dict[str, ConfigVar]]:
     """
     Given the name of a strategy, find and load strategy-specific config map.
     """
-    if strategy is None:
-        return None
     try:
+        if strategy is None:
+            strategy = global_config_map.get("strategy").value
         cm_key = f"{strategy}_config_map"
         strategy_module = __import__(f"hummingbot.strategy.{strategy}.{cm_key}",
                                      fromlist=[f"hummingbot.strategy.{strategy}"])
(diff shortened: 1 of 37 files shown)


mifeng:
`create` command should ask user what strategy and save the strategy in the config file. Currently, it assumes user is running PMM and doesn't ask the strategy question.

Nullably:
@mifeng I have added the strategy question to `create` command. Also removed strategy config variable from config files. The bot knows which strategy the config file is by reading the top 3 lines and scan for the strategy name.

hummingbot/client/config/config_helpers.py
```diff
@@ -191,21 +189,36 @@ def get_strategy_starter_file(strategy: str) -> Callable:
...
+
+
+def strategy_type_from_file(file_path: str) -> str:
+    strategies = get_strategy_list()
+    for strategy in strategies:
+        if strategy_matches_file(file_path, strategy):
+            return strategy
+
+
+def strategy_matches_file(file_path, strategy) -> bool:
```
mifeng, line 206: This method currently works, but in the future we plan to support users creating their own strategies or loading other custom strategies. Would defining a `strategy` param be better?

Nullably: We simply need a kinda hard coded value in the file, it doesn't have to a config variable, just another entry in the template file will do. Though this will not be backward compatible as older version config files will not have this field. 
I'm fine with either approaches.

mifeng: Let's require a `strategy` param in each config file that specifies the strategy name.

Since we're making so many changes to the client, the old config files won't be backward compatible with this new release anyway.

Nullably: Yep, agree, probably will easier for other developers to understand the code base this way too.
I have made the changes to all strategy config files and their templates. Also fixed the problem with creating Arbitrage and XEMM strategy configs.

Nullably:
Run tests

Nullably:
run tests

PR #1591, opened 2020-04-06
Feat / connect and balance command

Nullably:

**Optional**:
- Related Github issue:
- Related Clubhouse Story: 8086, 8125

- Add connect and balance command
- connect without an argument shows all current exchanges connections
- If connection failed, the reason is displayed.

### DIFF
--- a/hummingbot/client/command/__init__.py
+++ b/hummingbot/client/command/__init__.py
@@ -30,7 +30,7 @@ __all__ = [
     StatusCommand,
     StopCommand,
     ConnectCommand,
-    BalanceCommand
+    BalanceCommand,
     CreateCommand,
     ImportCommand
 ]
--- a/hummingbot/client/command/__init__.py
+++ b/hummingbot/client/command/__init__.py
@@ -10,6 +10,8 @@ from .paper_trade_command import PaperTradeCommand
 from .start_command import StartCommand
 from .status_command import StatusCommand
 from .stop_command import StopCommand
+from .connect_command import ConnectCommand
+from .balance_command import BalanceCommand
 
 
 __all__ = [
@@ -25,4 +27,6 @@ __all__ = [
     StartCommand,
     StatusCommand,
     StopCommand,
+    ConnectCommand,
+    BalanceCommand
 ]
--- a/hummingbot/client/command/balance_command.py
+++ b/hummingbot/client/command/balance_command.py
@@ -1,0 +1,57 @@
+from hummingbot.user.user_balances import UserBalances
+from hummingbot.core.utils.async_utils import safe_ensure_future
+from hummingbot.core.utils.exchange_rate_conversion import ExchangeRateConversion as ERC
+import pandas as pd
+from numpy import NaN
+from typing import TYPE_CHECKING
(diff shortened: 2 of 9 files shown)


Nullably:
run tests

Nullably:
run tests

mifeng:
1. `connect`: what is the difference between **Keys Saved** and **Connected**?
Screen Shot 2020-04-06 at 5 39 46 PM

2. `balance`: this is a really useful command. are you using the price feeds to calculate the USD rate? what happens if an asset doesn't have the price feed available?

mifeng:
For `balance`, please sort the dataframe so that it's descending based on USD value.

Nullably:
> 1. `connect`: what is the difference between **Keys Saved** and **Connected**?

Keys Saved is whether the user has provided api key/secret through connect/config command.
Connected is whether the connection can be made from those keys, though connection could fail for other reason, e.g. an exchange down, ip address not white listed...
> 2. `balance`: this is a really useful command. are you using the price feeds to calculate the USD rate? what happens if an asset doesn't have the price feed available?

For the USD value, likely it is from CoinGecko, if the conversion not available it shows zero.

mifeng:
In the medium term, we should add the user's Ethereum wallet as well to `balance` so that if the user trades on DEXs they can see how that compares.

However, currently users can import/create multiple Ethereum wallets into Hummingbot. Before we add wallets to `balance`, I think we should restrict users to only have a single ETH wallet in Hummingbot. This allows us to treat wallets and exchange accounts similarly.

mifeng:
>Keys Saved is whether the user has provided api key/secret through connect/config command.
Connected is whether the connection can be made from those keys, though connection could fail for other reason, e.g. an exchange down, ip address not white listed...

I see - please change the column headings to "Keys Added" and "Keys Confirmed". Also, insert a newline before **Failed connections:** to separate that from the table a bit more.

mifeng:
Remove the following exchanges from `connect` and `balance`, since they are Ethereum DEXs that use the balance in the user's wallet:
`radar_relay`
`dolomite`
`bamboo_relay`

Also, please remove `bitcoin_com` since we're not officially supporting them

Nullably:
> In the medium term, we should add the user's Ethereum wallet as well to balance so that if the user trades on DEXs they can see how that compares.

Looks like it's gonna be a fair bit of work on the design and implementation for Ethereum wallet and DEXs, let's leave them for another PR?

> I see - please change the column headings to "Keys Added" and "Keys Confirmed". Also, insert a newline before Failed connections: to separate that from the table a bit more.

Done

> For balance, please sort the dataframe so that it's descending based on USD value.

Done, both Token and Exchanges are now sorted based on their total USD value.

> Remove the following exchanges from connect and balance, since they are Ethereum DEXs that use the balance in the user's wallet:

Done

mifeng:
>Looks like it's gonna be a fair bit of work on the design and implementation for Ethereum wallet and DEXs, let's leave them for another PR?

Definitely - just wanted to highlight that as something that we'll do in the future.

Nullably:
run tests

PR #1602, opened 2020-04-08
Feat / update config command and config validation

Nullably:

**Optional**:
- Related Github issue:
- Related Clubhouse Story: 8035, 8085, 8016

- `config` command now lists config parameter values, global (subset) and strategy.
- `config` also accepts 2 arguments, config key name and config value. This is to update the config.
- New inventory skew question at the end of `create` process
- Display meaningful error messages when config validation fails. Updated all validation function to return a string (error message) instead of a boolean.


Nullably:
@mifeng 
> Don't show the default value for any configuration question, except the last one with the configuration file name. When answering the questions, it may be confusing which ones have the default value and which ones don't. It's better to not show the default value for all questions.

You mean on the `create` command only or `config [parameter]` as well? If the user just hits enter, should that triggers default value?

Nullably:
@mifeng 
> Add the asset balances to the inventory skew question so that user knows we're fetching their asset balances:

Done.

For the defaults, I removed them for the configs during `create`

Nullably:
run tests

mifeng:
Yes, remove defaults for `create` only is what i meant.

PR #1614, opened 2020-04-13
Feat / connect ethereum command

Nullably:

**Optional**:
- Related Github issue:
- Related Clubhouse Story: 8199

- Added `connect ethereum` command to connect the bot to an Ethereum wallet
- Added `ethereum_wallet` global config to store the connected wallet


mifeng:
Found another bug: when running `connect wallet`, replying either Yes or No to the question "Would you like to replace your existing Ethereum wallet..." just shows the prompt again

Nullably:
@mifeng 
> After adding ETH wallet, connect should show Yes for both Keys Added and Keys Confirmed

Done.

> After adding ETH user, user should be able to create a strategy with Radar Relay. Currently, the inventory pct question is not displayed.

This is quite complicated for ERC20 tokens, I will need to spend more time understanding web3 and contract to get ERC20 balance. We should create a new ticket for this. To get ETH balance though is quite simple, I have added this in `balance` command (if user has added Ethereum wallet).

> I like how we now prompt users with the question to change the wallet if they have already added one? Please add the same question for all exchanges and show only first 4 chars of the API key field (Would you like to replace your existing binance API key abcd...?`)

Done. I opted for showing the last 4 chars (instead of first 4) to keep it consistent with `status` command.

> Also, I found that adding Kraken keys results in the following error below. Not sure if this is related to this change or in the Kraken connector

I have added Kraken to the UserBalances, it should fix the error.

Nullably:
@mifeng 
> Found another bug: when running connect wallet, replying either Yes or No to the question "Would you like to replace your existing Ethereum wallet..." just shows the prompt again

Fixed.

PR #1624, opened 2020-04-15
fix / wrong price precision for some BTC pairs

dennisocana:

- Related Github issue: https://github.com/CoinAlpha/hummingbot/issues/1618, https://github.com/CoinAlpha/hummingbot/issues/1617

Fixed issues with how prices are displayed causing problems when creating orders in paper trade mode.


mifeng:
@dennisocana This only affects the precision in the PaperTrade market class. I believe that the TRX/BTC issue is different since we fetch the precision from the exchange API.

dennisocana:
@mifeng I was looking at this method yesterday and I think we follow the same tick size from the Binance API.

image

We confirmed that the precision works fine when doing live strategy. Here is a screenshot for the same pair using actual assets.

image

However in paper trade mode it cuts down to 7 (as shown in the screenshot in https://github.com/CoinAlpha/hummingbot/issues/1618) because currently the price returns 7 decimal values only in `c_get_order_price_quantum`.

image

martinkou:
retest this please

mifeng:
test this please

PR #1640, opened 2020-04-21
Fix / Export command and minor fixes

Nullably:

**Optional**:
- Related Github issue:
- Related Clubhouse Story: 8342

- Added a new `export` command to replace export_private_keys and export_trades commands.
- Minor fix to notify users if no exchange connection after `create` or `import` strategy config file.
- Removed list command


mifeng:
2. Add performance check text
When user runs `create` or `import`, we perform a check and emit error messages if the strategy is missing values. We should make it clear what's happening. Currently, it's hard to tell the difference between a successful `import` and an unsuccessful one:
Screen Shot 2020-04-21 at 11 56 45 AM

Add the following text output to the end of the `create` and `import` process, similar to the style of the checks output emitted when users run `start` or `status`:

```
Preliminary checks:
- Exchange check: EITHER [All connections confirmed] or [Missing exchange connection...]
- Strategy check: EITHER [All required parameters confirmed] or [Incomplete strategy configuration...]
```

mifeng:
3. Conform the `status` and `start` strategy configuration checks to the new error messages, which are much more descriptive. Currently, we print different error messages but the checks are the same:
Screen Shot 2020-04-21 at 12 10 16 PM

Nullably:
@mifeng I have addressed all the issues here.
Minor changes to your requests:
- I use `Invalid exchange connections` instead of `Missing exchange connection` whereas missing API keys is one reason why connection might be invalid.
- I removed `Please run connect/config` instruction.

mifeng:
LGTM

PR #1650, opened 2020-04-23
Fix / API keys error paper trade

Nullably:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:

- Remove exchange connection check for paper trade mode and DEXs
- Add a new is_api_key property to config_var
- Set inventory_skew_enabled to false, this will ask and set during a create process on the extra question, for some reason if this extra question doesn't happen the default should be false to avoid triggering the other inventory fields to be required.


dennisocana:
Also confirmed fix on `CTRL + S` keyboard shortcut .

PR #1673, opened 2020-04-27
Feat / refresh tolerance

Nullably:

**Optional**:
- Related Github issue:
- Related Clubhouse Story: 8151

- Add a new config parameter `order_refresh_tolerance_spread`, if set to negative value e.g. -1, this means the feature is disable.
- Update all sizing delegates to always return order proposals when requested.
- Add 2 new timers (timestamps) to keep track of when to cancel and to create new orders.
- Add unit testing for the new order refresh tolerance feature


mifeng:
- Tested with 5 order levels in the BTC/USDT market on Binance, with tolerance at `0.01` and `0.1`
- Overall logic seems to be correct. To check, if I set both spread and order_level_spread to `0.2` and order_tolerance_spread to `0.01`, all orders in the screenshot below should be refreshed (assuming no rounding). Is that right?
Screen Shot 2020-04-27 at 1 47 54 PM
- Please add log messages that tell user what happens when there's no order refresh, maybe something like `Not cancelling order [order-ID] since difference between 0.21% current spread and 0.20% target spread is less than 0.01% order_refresh_tolerance_spread`
- I don't think we need a special -1 value to disable this feature. If users can just leave the default value at 0, since that will be the same behavior as if they disabled it.

Nullably:
@mifeng 
> Overall logic seems to be correct. To check, if I set both spread and order_level_spread to 0.2 and order_tolerance_spread to 0.01, all orders in the screenshot below should be refreshed (assuming no rounding). Is that right?

Yes

> Please add log messages that tell user what happens when there's no order refresh, maybe something like Not cancelling order [order-ID] since difference between 0.21% current spread and 0.20% target spread is less than 0.01% order_refresh_tolerance_spread

Do you want this for every order? I imagine it's gonna be lots of messages for users who set multi levels order on short `order_refresh_time`

> If users can just leave the default value at 0, since that will be the same behavior as if they disabled it.

Not the same behaviour, tolerance at 0 means if the market doesn't move at all (mid price remains the same since the orders were created), the orders won't get canceled.

Nullably:
@mifeng 
I have pushed changes as requested. The new `order_refresh_tolerance_enabled` is default to no. 

It is easier and more accurate to compare order new prices with the existing ones (rather than comparing their spreads to mid_price) as the order proposal has done rounding and trading rule check. So the tolerance spread is actually a spread between new proposed prices and the existing orders prices. I changed the log message to accommodate this.

mifeng:
@Nullably 
* Are you comparing order prices to see if they differ by more than `order_refresh_tolerance_spread` to determine if they should be adjusted? If so, that is incorrect and we should compare the new spread vs the old spread, not prices
* We should always compare an order's current to the initial order spread, not to the previous order spread. As a MM, you want the bot to maintain a certain fixed numerical spread. Otherwise, your bot spreads could drift up or down if the change in each refresh is arbitrarily small
* Thought about this some more and I believe a 0 default value for `order_refresh_tolerance_spread` is the same as disabling this feature, and we can remove the boolean enable param. When that param is 0, it should mean that any change in order spread causes it to change, which is basically the same as disabling it.

Nullably:
I have removed the `order_refresh_tolerance_enabled` parameter. `order_refresh_tolerance_spread` is now default to 0. 
As discussed, the order price difference (in percentage) can be used to check whether to refresh the orders.

mifeng:
Tested and the UX looks good. A couple of minor UX changes:
- Since active orders will be all refreshed or not at all, just emit a single log message when orders are not refreshed: "Not cancelling active orders since difference..."
- Change the param name to `order_refresh_tolerance` - I think that would be simpler and less confusing

martinkou:
There's some pretty significant changes to how order proposals are processed in this patch. I'm gonna need another day to review this thoroughly.

martinkou:
How does the refactored sizing delegates handle existing orders? I don't see any logic now that would cause it to skip creating more orders when there are already existing orders.

Nullably:
@martinkou order proposals get created only when `_create_timestamp` is due (in `c_tick`). 
'c_cancel_active_orders()' will check the order proposals against existing active orders, in `c_is_within_tolerance`, and return `True` if cancelling is deferred. If it is deferred, `c_execute_orders_proposal` will not get called. 

Also, order cancelation and creation could happen in the same tick, but I keep the process more of less the same by adding _create_timestamp by 1 so that the creation happens in the next cycle if orders are canceled. This is in `c_cancel_active_orders`

There is a bit of code cleanup, and more unit test cases should be created, but as a POC I didn't pay much attention to these details.

martinkou:
Rather than setting `_create_timestamp` in the order complete handlers and `c_cancel_active_orders()` ... it'd be safer to separate that out to a function that detects whether more active orders can be created or not and set it there.

The reason is twofold.

 1. When you call `c_cancel_active_orders()`... it's not guaranteed that the order would actually be cancelled. e.g. you may get an API call error later. The way it's currently designed would be prone to accumulating open orders after some time. It's safer to depend on the active order tracker to track your open orders and detect whether you should create more orders, rather than assume all cancels would be successful.

 2. The logic flow is simpler. There's just one place to trigger order creation.

Nullably:
@mifeng I have updated the PR to include your change requests, I opted for `order_refresh_tolerance_pct` instead of just `order_refresh_tolerance` to keep it consistent with other params.

Nullably:
@martinkou 
I created `c_to_create_orders` to check for the below conditions 
- No active orders
- Order proposal present
- `create_timestamp` has passed.

I'm not sure about not setting the `create_timestamp` in order complete handlers and in `c_cancel_active_orders.` To me the events that reset the timers (both create and cancel) are
- On orders created - timers are set by `order_refresh_time` amount
- On order cancelation deferred  - timers are set by `order_refresh_time` amount
- On order complete handlers - timers are set by `filled_order_delay` amount

Nullably:
I still leave this below in `c_cancel_active_orders`
```
if self._create_timestamp <= self._current_timestamp:
     self._create_timestamp = self._current_timestamp + 0.1
```
In production, this is not necessary as the system needs to wait for the no active orders condition. But in testing, order cancelation and creation happen in the same tick, and some test cases fail (as they assume otherwise).

Nullably:
@martinkou 
I have cleaned up the code and make it more cython compliant, also add more testing for hanging orders scenario, I think this should cover most of cases. In case, you might want to review one more time as it's no longer POC.

PR #1683, opened 2020-05-01
fix / start stop memory leaks

martinkou:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:
https://app.clubhouse.io/coinalpha/story/8364/investigate-memory-leak-when-stopping-and-restarting-the-strategy

* Fixed start/stop memory leaks, caused by background task leaks in Binance order book tracker and user stream API data source.

### DIFF
--- a/hummingbot/market/binance/binance_api_user_stream_data_source.py
+++ b/hummingbot/market/binance/binance_api_user_stream_data_source.py
@@ -104,32 +104,38 @@ class BinanceAPIUserStreamDataSource(UserStreamTrackerDataSource):
         return websockets.connect(stream_url)
 
     async def listen_for_user_stream(self, ev_loop: asyncio.BaseEventLoop, output: asyncio.Queue):
-        while True:
-            try:
-                if self._current_listen_key is None:
-                    self._current_listen_key = await self.get_listen_key()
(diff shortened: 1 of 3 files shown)


hummingbot/market/binance/binance_api_user_stream_data_source.py
```diff
@@ -104,32 +104,38 @@ async def get_ws_connection(self) -> websockets.WebSocketClientProtocol:
...
-                self.logger().error("Unexpected error while maintaining the user event listen key. Retrying after "
-                                    "5 seconds...", exc_info=True)
-                await asyncio.sleep(5)
+                    await self.wait_til_next_tick(seconds=60.0)
+                except asyncio.CancelledError:
+                    raise
+                except Exception:
+                    self.logger().error("Unexpected error while maintaining the user event listen key. Retrying after "
+                                        "5 seconds...", exc_info=True)
+                    await asyncio.sleep(5)
```
Nullably, line 133: Is it possible for this `except` to cause infinite loop running when exception, other than CancelledError, happens?

martinkou: That is the intended behavior. Background tasks should keep running until cancelled. If it is any errors other than being cancelled, it should print error log instead of exit.

PR #1695, opened 2020-05-05
fix / broken status command when paper trading

dennisocana:

- Related Github issue: https://github.com/CoinAlpha/hummingbot/issues/1666
- Related Clubhouse Story: https://app.clubhouse.io/coinalpha/story/8492/status-doesn-t-work-on-kucoin-in-paper-trading-mode

Fixed `ValueError: invalid literal for int() with base 10: ''` when sending the `status` command on selected exchanges when paper trading. To isolate the error, I checked for common denominators on exchanges that are having this issue (KuCoin, Coinbase Pro, Liquid, Bittrex) and those that are working fine (Binance, Huobi).

Found out that the error happens on exchanges that uses "-" in trading pairs which is part of the order ID. For example, `buy://ETH-BTC/...` results to the error and `buy://ETHBTC/...` returns no error.

Tested after applying fix.

**KuCoin**

pycharm64_pm40ifDOYY

**Coinbase Pro**

pycharm64_H0DDFFHYnn

**Liquid**

pycharm64_awLGrDbY2e

**Bittrex**

pycharm64_JgEUHbVo81


Clint-G13:
Confirm fix issue on PMM, however, status still doesn't work when running cross exchange market making.
broken status

dennisocana:
Thanks @Clint-G13 please test again to confirm using cross exchange strategy.

pycharm64_tgSZSeBUTX

mifeng:
@dennisocana Please think about these changes a bit more. You're using an `and` in one place and `or` in the other. Are you sure that's correct?

Also, think about what the code is trying to do. We put `n/a` there when it's an paper order, so it doesn't have a timestamp that we can use to calculate age. What is the best way to detect via the orderID if it's a paper order or not?

hummingbot/core/data_type/limit_order.pyx
```diff
@@ -26,7 +26,7 @@ cdef class LimitOrder:
                 float(order.price),
                 f"{(0 if mid_price == 0 else abs(float(order.price) - mid_price)/mid_price):.2%}",
                 float(order.quantity),
-                ("n/a" if "-" not in order.client_order_id else
+                ("n/a" if "-" or "//" not in order.client_order_id else
```
mifeng, line 29: why do we nee do to look for both the dash and `//`?  is one of them sufficient to separate real orders from paper orders?

dennisocana: Thanks for the feedback @mifeng you're right `//` would suffice to identify paper trade orders based on the order ID. Just pushed a commit to change it.

Issue #1678, opened 2020-04-29
[BUG] Error when responding to order amount prompt in Huobi


dennisocana:
**Describe the bug**

Same as https://github.com/CoinAlpha/hummingbot/issues/1668 getting an error and can't get past the order amount prompt when setting up pure market making on Huobi.

![image](https://user-images.githubusercontent.com/50150287/80561292-8d22c080-8a16-11ea-9c95-3081480d55c1.png)

Error from logs :

```
2020-04-28 03:50:25,314 - 11 - hummingbot.core.utils.async_utils - ERROR - Unhandled error in background task: type NoneType doesn't define __round__ method
Traceback (most recent call last):
  File "/hummingbot/core/utils/async_utils.py", line 9, in safe_wrapper
    return await c
  File "/hummingbot/client/command/create_command.py", line 64, in prompt_for_configuration
    await self.asset_ratio_maintenance_prompt(config_map)
  File "/hummingbot/client/command/create_command.py", line 130, in asset_ratio_maintenance_prompt
    base_ratio = round(base_ratio, 3)
TypeError: type NoneType doesn't define __round__ method
```

Same workaround :

1. Create a strategy config file first before adding Huobi APi keys
2. If there are API keys in `conf` directory, delete encrypted keys
3. Import an existing strategy config file then start - make sure to use the latest version of the pure market making strategy config file template

```
########################################################
###       Pure market making strategy config         ###
########################################################

template_version: 12
strategy: pure_market_making
```

PtrckM:
Confirmed fix on **dev-0.27.0**

![image](https://user-images.githubusercontent.com/54516858/81273177-18d4c680-9081-11ea-8960-e100aecaa2c0.png)
![image](https://user-images.githubusercontent.com/54516858/81273470-7f59e480-9081-11ea-9988-0679923a649b.png)

Issue #1668, opened 2020-04-26
[BUG] PMM | kraken loop on minimum order


PtrckM:
**Describe the bug**
- unable to pass minimum order using kraken connector

**Steps To Reproduce**
1. create pmm strategy 
2. choose kraken connector
3. choose token pair
4. enter spreads
5. enter order amount (min order) on loop

**Screenshots**
![image](https://user-images.githubusercontent.com/54516858/80307292-b0712400-87fa-11ea-8591-133d76c2489b.png)

**WORKAROUND**
- remove kraken api and keys before creating a pmm strategy conf
```
-rw-rw-r--  1 patrick patrick  441 Apr 26 12:07 encrypted_kraken_api_key.json
-rw-rw-r--  1 patrick patrick  505 Apr 26 12:07 encrypted_kraken_secret_key.json
```
- or copy other pmm strategy conf then edit then use import on bot

![image](https://user-images.githubusercontent.com/54516858/80307385-4efd8500-87fb-11ea-93da-f5f8be4bf88b.png)
![image](https://user-images.githubusercontent.com/54516858/80307387-5329a280-87fb-11ea-87de-33155a3a50fe.png)
![image](https://user-images.githubusercontent.com/54516858/80307398-718f9e00-87fb-11ea-8cad-0d7fb2e61c8a.png)

**Release version**
0.26.0 / Source Build, Docker Build and Binary Installer

dennisocana:
Posting traceback from logs also to be used for referencing future related errors.

```
2020-04-29 07:31:37,136 - 17484 - hummingbot.core.utils.async_utils - ERROR - Unhandled error in background task: Unable to convert 'USDC' to 'XETH'. Aborting.
Traceback (most recent call last):
  File "hummingbot\core\utils\async_utils.py", line 9, in safe_wrapper
  File "hummingbot\client\command\create_command.py", line 64, in prompt_for_configuration
  File "hummingbot\client\command\create_command.py", line 129, in asset_ratio_maintenance_prompt
  File "hummingbot\user\user_balances.py", line 137, in base_amount_ratio
  File "hummingbot\core\utils\exchange_rate_conversion.py", line 179, in convert_token_value_decimal
  File "hummingbot\core\utils\exchange_rate_conversion.py", line 210, in convert_token_value
ValueError: Unable to convert 'USDC' to 'XETH'. Aborting.
```

PtrckM:
Confirmed fix on dev-0.27.0

[...]

PR #1696, opened 2020-05-05
Fix / create command stuck for Huobi and Kraken

Nullably:

**Optional**:
- Related Github issue: #1668 #1678 
- Related Clubhouse Story: 8489

- Allow the create configuration process to complete when exchange conversion or trading pair fetchers fail
- Get balance is now not case sensitive, Huobi connector stores balances in lower case assets, while our UX uses upper cases.

Please note, trading pairs auto-complete still not working for Kraken.

### DIFF
--- a/hummingbot/user/user_balances.py
+++ b/hummingbot/user/user_balances.py
@@ -104,7 +104,12 @@ class UserBalances:
 
     async def balances(self, exchange, *symbols):
         if await self.update_exchange_balance(exchange) is None:
-            return {k: v for k, v in self.all_balances(exchange).items() if k in symbols}
+            results = {}
+            for token, bal in self.all_balances(exchange).items():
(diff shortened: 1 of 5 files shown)


hummingbot/user/user_balances.py
```diff
@@ -131,9 +136,12 @@ def validate_ethereum_wallet() -> Optional[str]:
 
     @staticmethod
     def base_amount_ratio(trading_pair, balances):
```
martinkou, line 138: This file needs some return type annotations in general. e.g. this function needs `-> Optional[float]`. Please also add return types for `ethereum_balance()`, `balances()`, `all_balances_all_exchanges()`.

Nullably: Done.

PR #1740, opened 2020-05-13
fix/version number not shown in welcome screen

vic-en:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:


mifeng:
@vic-en While this change address the issue, it introduces another place in the client that we have to change with each release.

How does the client display the version in the header bar? Does it hard-code the version number or does it import the version from another module?

PR #1786, opened 2020-05-20
fix / performance analysis logger

kmcgowan2000:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:

Previously the logger in PerformanceAnalysis wasn't initialised and so couldn't be used.


Nullably:
@kmcgowan2000 
Doesn't seem like logger is used anywhere in this class.
This class has some old code that is not used anymore, I think only the last 3 functions are relevant. As part of the exchange rate converter clean up task, I'm gonna clean up this class as well, it will no longer be a class, there will be 3-4 functions to calculate performance.

PR #1797, opened 2020-05-21
feat/add_shortcut_to_exit_out_of_config_question

vic-en:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:


vic-en:
The command to use is [Ctrl + x].

hummingbot/client/ui/hummingbot_cli.py
```diff
@@ -74,6 +74,10 @@ def accept(self, buff):
         self.log(output)
         self.input_handler(self.input_field.text)
 
+    def reset(self):
+        self.pending_input = "exiting create command configuration..."
```
mifeng, line 78: I don't think this input is needed since the message in `keybindings.py` is already printed

Screen Shot 2020-05-21 at 5 03 00 PM

hummingbot/client/command/create_command.py
```diff
@@ -41,10 +41,15 @@ async def prompt_for_configuration(self,  # type: HummingbotApplication
         self.placeholder_mode = True
         self.app.hide_input = True
         required_exchanges.clear()
+        # add self.stop_config to know if cancel is triggered
+        self.stop_config = None
```
Nullably, line 45: It's not very clear here if this attribute is a boolean, you can just initiate it to `False` (instead of `None`) if it's gonna be only either True or False. And perhaps use a more indicative name like `to_stop_config` or `to_exit_config`.

hummingbot/client/ui/hummingbot_cli.py
vic-en: CTRL-C is already in use for copy functionality.

mifeng: ok thanks - CTRL-X is fine.

Nullably:
@vic-en 
Currently the shortcut doesn't work on the last question on `create` command (to enter a new file name).

I think we can refactor the code a bit to make it cleaner:
- Change all prompts to use `prompt_a_config`, if the config variable is not required, simply create a temporary config variable. 
- Remove `prompt_new_file_name`, use `prompt_a_config` instead
- Update `prompt_a_config` to return `None` if the exit happens, otherwise return `parse_cvar_value(config, input_value)`
- Update its callers to check for None to exit config loop, e.g.
```
config.value = await self.prompt_a_config()
if config.value is None:
    return 
```

vic-en:
@Nullably  
Note that only the `create` command uses `prompt_new_file_name`. And the function does a few good things to ensure a proper file name is given.
Also, think of the possibility of wanting to accept `None` as a value for a prompt in future. In that case, modifying `config.value` to `None` might result in wasting more time trying to figure out what's wrong. However, using a `True` / `False` FLAG like the `to_stop_config` I implemented would prevent that kind of an issue from arising.
I can just make the configuration stop like I did for the rest if the shortkey is triggered at the `prompt_new_file_name` level as well.
Kindly let me know if you still want me to implement it the way you proposed.

Nullably:
@vic-en I didn't mean to use config.value as a flag. The `to_stop_config` flag is good. What I mean is we shouldn't need to set the flag value on every config. You can set the flag back to `False` in `prompt_a_config` function. 

Another way is to create a wrapper function for `app.prompt`, something like
```python
async def hb_prompt(....) -> Optional[str]:
    input = await self.app.prompt(...)
    if self.app.to_stop_config:
          self.app.to_stop_config = False
          return None
    return input
```
So when we need to prompt user for input, it will have this exit option available to them.
We use `hb_prompt()` instead of `app.prompt`, this is to make it clear to later on that this is how our prompt() should be used, ie. always have exit shortcut. The callers of `hb_prompt` needs to always for None value for when user exits.
```python
input = await hb_prompt()
if input is None:
    return # or whatever it needs to do here
```

vic-en:
@Nullably I get your point. Just afraid it **might** break a lot if no validator is set as the default `config.value` is `None`.

vic-en:
@Nullably 
Just like I predicted, using `if input is None` leads to a hard to track bug in subsequent prompts when there is an error in a previous prompt.

Should I push the commit so you can check it out?

vic-en:
Although it works fine if no validation fails at all. There's a bug that makes `val = None` after a corrected failed validation check.
Error

vic-en:
If not setting the `to_stop_config` flag value on every config is what you want. Then I think a way out should be something like:

```
def stop_conf_triggered():
    if self.app.to_stop_config is True:
        self.app.to_stop_config = False
        return True
    return False

```
Then after every prompt,

```
if stop_conf_triggered():
    return
```

I can implement this if its okay with you.

mifeng:
Currently, hitting CTRL-X when prompted with the filename question at the end of `create`, will create a blank `.yml` file. Ideally, process exits without creating the file, but if it's simpler / more maintainable to keep it this way, it's fine.

Nullably:
@vic-en If there is no easy way to refactor the code, I'm ok with how it is atm.

vic-en:
Shortkey would work properly now when prompted for the filename question at the end

mifeng:
test this please

mifeng:
Unfortunately, the fix for the filename prompt made it worse. Now, trying to exit with CTRL-X at that step causes the user to get stuck in a loop. I suggest reverting the change - it's unlikely that users would exit that far into the process anyway, so I think it's fine to leave it as-is/

Screen Shot 2020-05-28 at 4 12 11 PM

mifeng:
Actually, I realized that the issue I referenced above is due to the fact that I had a `.yml` file in my conf directory. After I removed it, this now works so I wouldn't call it a bug. 

I will approve and merge this PR.

mifeng:
test this please

Issue #1780, opened 2020-05-19
[BUG] Kraken history command error on BTC pairs


dennisocana:
Getting `Error running performance analysis` when running `history` command after executing trades on a BTC pair.

<img width="947" alt="putty_hs18YXAbl1" src="https://user-images.githubusercontent.com/50150287/82292683-e65e8e00-99dd-11ea-967d-2d7e1084062e.png">

Traceback does not always appear in logs. This is the log file of the bot above that has no error in logs - 
[without_errors.log](https://github.com/CoinAlpha/hummingbot/files/4648604/without_errors.log)

The other bot I was running on Kraken has the same issue but the log file contains the traceback - 
[with_errors.log](https://github.com/CoinAlpha/hummingbot/files/4648609/with_errors.log)

```
2020-05-19 04:46:34,557 - 10 - hummingbot.client.hummingbot_application - ERROR - Unexpected error running performance analysis.
Traceback (most recent call last):
  File "/hummingbot/client/command/history_command.py", line 119, in trade_performance_report
    trade_performance_stats, market_trading_pair_stats = self._calculate_trade_performance()
  File "/hummingbot/client/command/history_command.py", line 91, in _calculate_trade_performance
    trade_performance_stats, market_trading_pair_stats = performance_analysis.calculate_trade_performance(
  File "/hummingbot/client/performance_analysis.py", line 244, in calculate_trade_performance
    stats["delta_percentage"] = ((stats["acquired"] / stats["spent"]) - Decimal("1")) * Decimal("100")
decimal.DivisionByZero: [<class 'decimal.DivisionByZero'>
```

Error is not present on **ETH-USDT** pair.

<img width="947" alt="putty_akjiwCPDWp" src="https://user-images.githubusercontent.com/50150287/82292951-540aba00-99de-11ea-8149-f3d9b1c6befc.png">

**Version** : latest 0.27.0
**Environment** : Docker in AWS EC2 instance

mifeng:
@kmcgowan2000 is this bug fixed now?

kmcgowan2000:
@mifeng not yet, prioritising backend but will be fixed this week

dennisocana:
I'm getting this error again after executing some trades on BTC-USDT pair.

```
2020-07-16 07:11:31,502 - 1 - hummingbot.core.utils.kill_switch - ERROR - Error calculating profitability: [<class 'decimal.DivisionByZero'>]
Traceback (most recent call last):
[...]

Issue #1779, opened 2020-05-19
[Refactor] Simplify market connector fee logic


mifeng:
Currently, our market connectors have logic to precisely track the coin used to pay for fees and whether it's taken out of the base or quote token. In addition, there is special logic to handle Binance, which used to handle fees different from other exchanges.

Binance has now adjusted their fee logic to match other exchanges so that the fee is always paid in the quote token (or in BNB if the user turns that on), so we don't need the extra Binance logic now. Also, the exact fee coin doesn't materially impact profitability, which is the most important metric. Adopting a simpler model for calculating fees will make the codebase simpler to maintain.

## Tasks
* Standardize how fees are calculated across all market connectors so that there's a maker fee  (% of order amount), taker fee (% of order amount), and a flat fee for each market (the flat fee will be zero for all centralized exchanges, and is used only for Radar Relay gas currently).
* Assume that both fees are always paid using the quote asset
* Remove the Binance fee handling logic
* In `fee_overrides`, let users override maker, taker, and flat fees for each exchange

Nullably:
We currently assume market order is a taker order and limit order a maker order, this is not always the case for limit order. A limit order can be either maker or taker depending on whether the order matches at submission.
I think this should be fixed in `c_get_fee` method, instead of `OrderType` use `is_maker` argument.

Nullably:
To clarify more on my thoughts on fees.
- Whether an order is a maker or a taker should be considered separately from Limit/Market order types.

- Fees can thought of in 2 stages, before or after a trade happens. Before a trade, this is like an estimation of what the fee is going to be, this is usually needed to whether to place an order and by how much. The strategy knows beforehand whether an order is gonna be a maker or taker. For after a trade happens, we can usually get precisely how much fee is charged for the trade, in case of Binance, this information is available for both REST API and web socket. 
For our Binance connector, I am not sure why we replace this info with `c_get_fee`, as can be seen on line 621 in `binance_market.pyx`  

- As a side, I think we should not use market order type at all, as it is risky and can be taken advantaged of by other types of trading strategies. I suspect that is one of the reasons why users report losing money on xemm strategy. We currently submit the taker side of the trade as a market order. When we fix this, ie. use a limit order instead of a market order, we can no longer consider the fee by order type.

Nullably:
My recommendation:
[...]

PR #1809, opened 2020-05-26
fix / kraken history command error

kmcgowan2000:

**Optional**:
- Related Github issue: #1780 
- Related Clubhouse Story:


kmcgowan2000:
> @kmcgowan2000 Can you add a comment to explain why this change fixed the bug? We're going to refactor how fees work in Hummingbot to make it simpler (#1779) so we will probably need to consider this Kraken issue when we do so

@mifeng I've actually just realised I don't think the existing logic was correct in the first place. I will adjust the PR to change that, and add a comment for reference when refactoring.

PR #1818, opened 2020-06-01
Fix/bamboo relay tests may 2020

Arctek:

- Added a timestamp to the transaction receipt watcher, so it doesn't raise `TransactionNotFound` unless at least 120 seconds has passed and the RPC node can't find the transaction hash.
- Some bug fixes on the market
- Modified zero ex utils to allow for signatures on other chains (for testing)
- Updated Bamboo integration tests


mifeng:
I tested the UX and it worked initially. I did notice this error message in the trade log after running the `stop` command:
Screen Shot 2020-06-05 at 10 39 16 AM When I restarted and ran `stop` again, the error did not re-appear.

After a while, I started to see errors that said "Unknown error occurred while checking for transaction receipts." Note that I am using Infura and they may lowered free tier limits. See the logs attached:
[logs_conf_pure_mm_13.log](https://github.com/CoinAlpha/hummingbot/files/4737710/logs_conf_pure_mm_13.log)

Also, I noticed that the bot was actively cancelling orders instead of using the expiration time by default. This seems to be a regression that we should fix ASAP. I'll check if it's doing that on Radar Relay also.

Arctek:
I'll see if I can track down that co-routine.

I know that error that's rising is from a `TransactionNotFound` so this can happen if hummingbot issues multiple transactions at once, i.e. with increasing nonces.
But the RPC endpoint i.e. infura doesn't necessarily track all of these, as well as some transactions might get dropped from the mempool.

From what I can see the code in `web3_wallet_backend.py` is just for tracking the gas consumption, so maybe if a transaction can't be found after a few minutes it should get dropped from tracking instead of raising an error (which will just keep repeating)?

For active cancels I think you mentioned there is now a flag that can be set to -1 to switch this off now, for the pure strategy and then it would just use the expiry time instead?

mifeng:
>For active cancels I think you mentioned there is now a flag that can be set to -1 to switch this off now, for the pure strategy and then it would just use the expiry time instead?

Yes, @vic-en and I are working on it in https://github.com/CoinAlpha/hummingbot/issues/1833. We'll try to make sure that the PMM strategy works for 0x relayers whether user selects to use expirations or active cancels.

mifeng:
test this please

PR #1824, opened 2020-06-02
Refactor/market connector fee logic

vic-en:

**Optional**:
- Related Github issue: #1779 
- Related Clubhouse Story:

### DIFF
--- a/hummingbot/market/binance/binance_market.pyx
+++ b/hummingbot/market/binance/binance_market.pyx
@@ -149,8 +149,6 @@ cdef class BinanceMarket(MarketBase):
         self._order_not_found_records = {}  # Dict[client_order_id:str, count:int]
         self._tx_tracker = BinanceMarketTransactionTracker(self)
         self._trading_rules = {}  # Dict[trading_pair:str, TradingRule]
-        self._trade_fees = {}  # Dict[trading_pair:str, (maker_fee_percent:Decimal, taken_fee_percent:Decimal)]
-        self._last_update_trade_fees_timestamp = 0
         self._data_source_type = order_book_tracker_data_source_type
         self._status_polling_task = None
         self._user_stream_event_listener_task = None
@@ -298,24 +296,6 @@ cdef class BinanceMarket(MarketBase):
             del self._account_available_balances[asset_name]
             del self._account_balances[asset_name]
 
-    async def _update_trade_fees(self):
-        cdef:
-            double current_timestamp = self._current_timestamp
-
-        if current_timestamp - self._last_update_trade_fees_timestamp > 60.0 * 60.0 or len(self._trade_fees) < 1:
-            try:
-                res = await self.query_api(self._binance_client.get_trade_fee)
-                for fee in res["tradeFee"]:
-                    self._trade_fees[fee["symbol"]] = (Decimal(fee["maker"]), Decimal(fee["taker"]))
-                self._last_update_trade_fees_timestamp = current_timestamp
-            except asyncio.CancelledError:
-                raise
-            except Exception:
-                self.logger().network("Error fetching Binance trade fees.", exc_info=True,
-                                      app_warning_msg=f"Could not fetch Binance trading fees. "
-                                                      f"Check network connection.")
-                raise
-
     cdef object c_get_fee(self,
                           str base_currency,
                           str quote_currency,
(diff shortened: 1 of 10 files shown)


vic-en:
@Nullably 
Kindly check and let me know if refactor is in line with what you want/intended.

vic-en:
Also, 3 connectors(BambooRelay, BitcoinCom & Dolomite) are yet to be refactored.

**Dolomite:**
- I can figure out a way to make Dolomite work.

Both BambooRelay and BitcoinCom implement c_get_fee method in a way that requires their Market classes to be instantiated. This makes it really difficult to make a refactor.

**BambooRelay:**
```
        valid_orders = self.c_get_orders_for_amount_price(trading_pair=base_currency + "-" + quote_currency,
                                                          trade_type=trade_type,
                                                          amount=amount,
                                                          price=Decimal(price))     # market base has this as a `float`

        for order in valid_orders:
            if order["isCoordinated"]:
                is_coordinated = True
                break

        if is_coordinated:
            transaction_cost_eth = self._wallet.gas_price * (len(valid_orders) * order_gas_cost + base_coordinator_cost) / 1e18
        else:
            transaction_cost_eth = self._wallet.gas_price * len(valid_orders) * order_gas_cost / 1e18

        protocol_fee = ZERO_EX_PROTOCOL_FEE_MULTIPLIER * len(valid_orders) * self._wallet.gas_price / 1e18

```

- I believe there shouldn't be need to call `self.c_get_orders_for_amount_price(...)`. That call retrieves all orders from the orderbook. Going to the later part of the code, len(active_orders) is multiplied by the gas fee. That implies that the fee is calculated for all unfilled orders once which should rather be done per order since there's a very low probability that all will be filled at once.
- If the call to ` self.c_get_orders_for_amount_price(...)` is removed, there should be another way to know if an order `is_coordinated` or not.

**BitcoinCom**
- Just figures they use a fixed 0.15%/0.2% maker/taker fee. So, I'll totally discard the existing logic and work with that.
BitcoinCom_Fee

hummingbot/market/radar_relay/radar_relay_market.pyx
```diff
@@ -253,22 +255,22 @@ cdef class RadarRelayMarket(MarketBase):
...
+        gas_estimate: int = 130000  # approximate gas used for 0x market orders
+        transaction_cost_eth: double
 
         # there are no fees for makers on Radar Relay
-        if order_type is OrderType.LIMIT:
+        if is_maker:
             return TradeFee(percent=Decimal(0.0))
         # only fee for takers is gas cost of transaction
-        transaction_cost_eth = self._wallet.gas_price * gas_estimate / 1e18
+        transaction_cost_eth = RadarRelayMarket.class_wallet_object.gas_price * gas_estimate / 1e18
```
Nullably, line 273: `class_wallet_object` will not have value unless at least one of object of the market class is created, and we shouldn't access class member from static method. I think we need to get gas fee in a similar way we get fees for centralised exchanges.

Nullably:
@vic-en 
I think we should leave is_maker consideration for get_fee for now. Sorry for the confusion. As it will further complicate the task, and we can create a new ticket for that. 

@mifeng 
Should we implement a similar fee structure for DEX? i.e. have a fix flat fee/gas for each exchange, and can be overridden by a config (fee_override). At the moment, Bamboo and Radar Relay have quite a complex fee structure, and we cannot have a static method to get them.

hummingbot/market/binance/binance_market.pyx
```diff
@@ -298,47 +296,20 @@ cdef class BinanceMarket(MarketBase):
...
-                          object amount,
-                          object price):
-        cdef:
-            object maker_trade_fee = Decimal("0.001")
-            object taker_trade_fee = Decimal("0.001")
-            str trading_pair = base_currency + quote_currency
-
-        if order_type.is_limit_type() and fee_overrides_config_map["binance_maker_fee"].value is not None:
+    @staticmethod
+    def c_get_fee(base_currency: str,
```
Nullably, line 300: Is it possible to not change this function from cython to python and keep it static? ie.
```python
@staticmethod
cdef object c_get_fee(...
```

we use `c_` prefix only for cython functions.

vic-en: Cython static methods cannot be called by instance name, it wouldn't even compile.
Using `instance.__class__` to access cython static method compiles but produces error at runtime.

I suggest we stick with `c_estimate_fee` as a python static class.
Let me know if you know a way around accessing cython static method from instance names.

mifeng:
> Should we implement a similar fee structure for DEX? i.e. have a fix flat fee/gas for each exchange, and can be overridden by a config (fee_override). At the moment, Bamboo and Radar Relay have quite a complex fee structure, and we cannot have a static method to get them.
Yes, I think we should use the exact same fee setup for DEXs.

mifeng:
On Asia standup, @Nullably and I talked about how to handle fees for the `history` command. 

Currently I believe we use the estimated fee rate to calculate fees for the sqlite db and `history`, but we think the right way is to use the fees returned via API. However, I don't think we should make it part of this PR, since it may introduce additional complexity due to different fee coins and the need to do conversions.

hummingbot/market/binance/binance_market.pyx
vic-en: cython_static

That is what I'm talking about.
Doing `market.__class__.c_estimate_fee(is_maker)` would compile but give error at runtime.

vic-en: should I make it python static again?

Nullably: hmm, perhaps making the method static is not quite feasible.
we might get into a bit over engineering here.
how about, for this task, we simply create a python module called `estimate_fee.py` (this can be in `hummingbot/core/utils`), it has a function called 'estimate_fee' and what it does is returning the fee based on market name and is_maker parameter? 
then all the connectors just call this function to get its fee, the paper trade market also call this function to get the fee. we then don't have to make any changes to strategy_base and market_base classes, no changes to c_get_fee signature. We can then leave out the refactoring for later. 
What do you think?

vic-en: Very good, nice plan.
I'll work with that and delete this branch and make a new PR.

vic-en: For DEX, are we good to go with having a fixed flat fee/gas for each exchange?

Nullably: @vic-en Yes, just a flat fee for each exchange, amount and token name, e.g. 0.00001 ETH. we will need to have fee overrides configuration for this as well so users can update this.

vic-en: Alright. I'll make commits soon.

mifeng:
@vic-en should i close this PR?

PR #1836, opened 2020-06-08
[WIP] feat/added min spread param to pmm

wpenn:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:

- added min_spread parameter to pmm strategy
- added unit test to `test_pmm.py`
- added documentation about param


wpenn:
Documentation In Progress

wpenn:
**Added Min-Spread param explanation** 
Docs: General --> Strategies --> Pure Market Making --> Basic Configuration and Walkthrough
(Section Highlighted in Red)
Untitled1

hummingbot/strategy/pure_market_making/pure_market_making_config_map.py
```diff
@@ -120,6 +120,14 @@ def exchange_on_validated(value: str):
                   type_str="decimal",
                   validator=lambda v: validate_decimal(v, 0, 100, inclusive=False),
                   prompt_on_new=True),
+    "minimum_spread":
+        ConfigVar(key="minimum_spread",
+                  prompt="How close to the mid price do you want to cancel active orders? "
```
mifeng, line 125: "At what minimum spread should the bot automatically cancel orders? (Enter 1 for 1%)

test/test_pmm.py
```diff
@@ -92,7 +92,8 @@ def setUp(self):
             order_amount=Decimal("1"),
             order_refresh_time=5.0,
             filled_order_delay=5.0,
-            order_refresh_tolerance_pct=-1
+            order_refresh_tolerance_pct=-1,
+            minimum_spread=-1,
```
mifeng, line 96: Make the default value 0. I can see a reason why users may want to make this parameter negative, so I think making the lower bound -1 is correct. However, I think users would want to default this at 0.

wpenn: I believe the default value is 0, and in this test case the value is -1 to disable the feature. However, the parameter allows for negative values in the UI, so I may need to consider a different way to disable the feature, or not allow non-negative numbers. A solution may be to add an additional parameter, such as "minimum_spread_enabled".

Issue #1782, opened 2020-05-19
[BUG] Kraken error running performance analysis in paper trade


dennisocana:
Creating a separate ticket since the pair that was working in https://github.com/CoinAlpha/hummingbot/issues/1780 (ETH-USDT) is having an error when in paper trading mode and may not have the same cause of problem.

Error in **ETH-USDT** pair - [paper_trade_eth-usdt.log](https://github.com/CoinAlpha/hummingbot/files/4649175/paper_trade_eth-usdt.log)

<img width="947" alt="putty_1qI9K5oUth" src="https://user-images.githubusercontent.com/50150287/82304341-c7b4c300-99ee-11ea-8f1e-5d1ddbb3d936.png">

Did another test in a different pair - **BCH-EUR** in paper trade and no errors.

<img width="947" alt="putty_uX6qgytmiV" src="https://user-images.githubusercontent.com/50150287/82304392-d7cca280-99ee-11ea-9dbf-4ad9623ca2c3.png">

**Version** : latest 0.27.0
**Environment** : Docker in AWS EC2 instance

mifeng:
@kmcgowan2000 is this bug fixed now?

kmcgowan2000:
@mifeng not yet – will be later this week

kmcgowan2000:
this appears to be happening because starting balances only seem to include the base and quote of the first strategy that is run once the client is opened. will look into the cause of this

PR #1857, opened 2020-06-11
fix / starting_balances not being populated with all assets

kmcgowan2000:

**Optional**:
- Related Github issue: #1782 
- Related Clubhouse Story:

Previously starting_balances would only be populated with the base and quote assets of the first strategy to be started once HB client was opened. Now it is populated with all assets available. This allows for the history command to function as expected even if trading pair is changed without closing the client.

### DIFF
--- a/hummingbot/client/command/history_command.py
+++ b/hummingbot/client/command/history_command.py
@@ -1,4 +1,5 @@
 from decimal import Decimal
+from collections import defaultdict
 
 import pandas as pd
 import threading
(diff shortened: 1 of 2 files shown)


hummingbot/client/command/history_command.py
```diff
@@ -44,16 +45,18 @@ def history(self,  # type: HummingbotApplication
         self.trade_performance_report()
 
     def balance_snapshot(self,  # type: HummingbotApplication
-                         ) -> Dict[str, Dict[str, Decimal]]:
-        snapshot: Dict[str, Any] = {}
+                         starting_balances: bool = False) -> Dict[str, Dict[str, Decimal]]:
```
Nullably, line 48: Would it be simpler for this function to always return the total list? i.e. no `starting_balances` argument required. So it will always return the same number of balances whether it's for starting or current balances.

mifeng:
Screen Shot 2020-06-11 at 9 17 02 PM
Got this error whien compiling

kmcgowan2000:
> Got this error whien compiling

@mifeng I'm not getting that, and that seems unrelated – maybe just something wrong locally for you?

hummingbot/client/command/history_command.py
kmcgowan2000: yes absolutely, wasn't sure if that would work with existing logic hence doing it this way – have now changed to the simpler method

Issue #1864, opened 2020-06-12
[Bug] Celo trades aren't displayed in `history`


mifeng:
<img width="620" alt="Screen Shot 2020-06-12 at 8 37 50 AM" src="https://user-images.githubusercontent.com/1795030/84536077-ef234680-aca1-11ea-9f30-1b39c5b45f1e.png">

Trades for both the Celo leg and the exchange leg should appear in `history` along with performance analysis calculations.

**Version**
`celo`

PR #1867, opened 2020-06-15
Fix / celo orders not showing on history command and other changes 

Nullably:

**Optional**:
- Related Github issue: #1864 
- Related Clubhouse Story:

- Update Bittrex connector when it uses `safe_ensure_future` to use `loop` arg. Note we will need to do this for all connectors later on where `celo_arb` is expected to support.
- Update `celo_arb` to use correct balance when checking whether it has enough for the trade.


martinkou:
You shouldn't force the market connectors to expect `c_buy()` and `c_sell()` to be called from different threads. Instead, you should have tracked the main thread event loop by saving in in `celo_arb.pyx` and use `call_soon_threadsafe()` to ensure the buy and sell ops are executed in main thread.

Nullably:
retest this please

Issue #1669, opened 2020-04-26
remove the check for order_level_amount > 0


bashir1961:
**Feature Suggestion**
In case of multiple orders it is possible to specify a positive order_level_amount, so the orders with highest spread have also the highest order.
I would like to experiment with negative order_level_amount, the idea being of creating a kind of gaussian curve around the middle, the orders with highest probability of being filled should have the highest order, to maximize the gain, at least in theory.

Should be pretty easy to do, just remove the check order_level_amount >0

PR #1868, opened 2020-06-15
(feat) remove the check for order_level_amount > 0

vic-en:

**Optional**:
- Related Github issue: #1669 
- Related Clubhouse Story:

Removed check for minimum order_level_amount in PMM.


hummingbot/strategy/pure_market_making/pure_market_making_config_map.py
```diff
@@ -174,7 +174,7 @@ def exchange_on_validated(value: str):
                          "additional order? >>> ",
```
Nullably, line 174: I think we should change to the prompt message a bit so user knows that decreasing order size is also an option. And we'll need to update doc and the template parameter description as well.

vic-en: Done!

mifeng:
Since this is an enhancement, I think we should merge this fix in after release

vic-en:
> 
> 
> Sorry @vic-en I should have triaged this story more carefully before adding this to the Public Roadmap.
> 
> This story is a good one, but I think the change should be in `order_level_amount`, not in `order_levels`. To allow users to decrease order sizes at higher levels, it's more intuitive to enter negative values for `order_level_amount`, so that the sizes decrease at higher order levels. Negative order levels are confusing.
> 
> Can you change this enhancement so that it adjusts the level amount instead?

@mifeng , The initial change was for `order_level_amount`.

mifeng:
@vic-en OK thanks - I'll review and test properly this time

PR #1876, opened 2020-06-16
Feat / script

Nullably:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:

- Add a new ScriptIterator to execute scripts.
- Add unit test cases to demonstrate how it works.
- Recreate ping pong and price band features using script.
- For demo only for now.

### DIFF
--- a/hummingbot/core/script_iterator.pxd
+++ b/hummingbot/core/script_iterator.pxd
@@ -1,0 +1,17 @@
+# distutils: language=c++
+
+from hummingbot.core.time_iterator cimport TimeIterator
+from hummingbot.core.clock cimport Clock
+
+cdef class ScriptIterator(TimeIterator):
+    cdef:
+        str _start_script
+        str _tick_script
+        str _buy_order_completed_script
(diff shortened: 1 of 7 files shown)


martinkou:
Don't do `exec()`... users can easily input scripts that kill the bot.

You should use `multiprocessing` instead. (https://docs.python.org/3.8/library/multiprocessing.html)

The flow should be like this..

 1. You define an entry function in the user script modules (e.g. `entrypoint()`). This is something the script writer must write.
 2. You define an API interface for communications between the child process and the parent.
 3. You import the file as a module (e.g. https://stackoverflow.com/questions/48841150/python3-importlib-util-spec-from-file-location-with-relative-path)
 4. You create a `Process` and start it from the defined entry function.
 5. You use interprocess queues (given in the `multiprocessing` module) for communication between the child process and the main process
 6. The script communicates with the main process via a pre-defined API interface. The API interface module, which is provided by us, performs the underlying interprocess communications via the multiprocessing queue. The script should not need to use the interprocess queues.

PR #1942, opened 2020-06-25
refactor / Replace inventory skew with ping pong in `create`

vic-en:

**Optional**:
- Related Github issue:
- Related Clubhouse Story: 9243

Replace inventory skew with ping pong in `create` command


hummingbot/client/command/create_command.py
```diff
@@ -158,13 +157,10 @@ async def asset_ratio_maintenance_prompt(self,  # type: HummingbotApplication
                                 f"{balances.get(quote, 0):.4f} {quote}. By market value, "
                                 f"your current inventory split is {base_ratio:.1%} {base} "
                                 f"and {quote_ratio:.1%} {quote}."
-                                f" Would you like to keep this ratio? (Yes/No) >>> ",
+                                f" Would you like to keep this ratio with ping-pong? (Yes/No) >>> ",
```
Nullably, line 160: I think what the story means is to drop this `asset_ratio_maintenance_prompt` entirely. Then on the `pure_market_making_config_map` you can set `prompt_on_new` to true on the ping pong parameter.

vic-en: The instruction in the story is to **replace** inventory skew with ping-pong in `create` command.

The approach of dropping `asset_ratio_maintenance_prompt` entirely too would equally give thesame functionality but would be against the to-do in the story.

Nullably: yeah, it is to replace the entire prompts with ping pong config prompt, not just a part of the prompt. 
@mifeng can give more clarification on this one.

we generally want to make use of `prompt_on_new` for the `create` command to streamline the process (as you can see for all the prompts before these last questions), the `asset_ratio_maintenance_prompt` was created as it wasn't able to fit into that process.

vic-en: I agree with you

vic-en: Done!

Issue #1680, opened 2020-04-29
New PMM param: take_if_crossed


babygiraffe05:
**Feature Suggestion**
The current behavior in Advanced Market Making is as follows:

"In a situation where the calculaton of maker order prices from external sources would result in the order matching any existing orders on the order book, such order will be ignored. For example, if ETH-USDC market is currently displaying 109 bid and 111 ask. A specified external exchange is showing 99 bid and 101 ask on its book (mid price = 100). 2 maker orders will be proposed, a bid maker order at 98 (for 2% bid spread) and an ask maker order at 102 (for 2% ask spread). The 102 ask order will be ignored (as it would match the 109 bid order), only the bid order will be submitted to the exchange."

However, in the above example, assuming the exchange in question is a less liquid market than the external exchange, it is likely that the price will converge toward 100. There would be tremendous value in selling at 109 (highest bid).

Therefore, a desirable setting would be "submit_matching_orders" [Yes/No] which would effectively post limit taker orders that will match existing maker orders. Even if the fee is higher, the behavior may be desirable because of the likely future price migration.

**Impact**
This would effectively allow Advanced Market Making strategy to double as "one-sided" arbitrage strategy on less-liquid exchanges, and lead to much greater liquidity when an exchange's price becomes dislocated from the broader market.

mifeng:
This is a great idea. The #1714 script feature may allow for this, but we may want to add this as an additional parameter regardless.

PR #1943, opened 2020-06-25
feat / Add new PMM param: take_if_crossed

vic-en:

**Optional**:
- Related Github issue:  #1680
- Related Clubhouse Story: 9271

Add param and functionality to submits a limit order on the current exchange with a price that matches that order if any order would result in a crossed market (most likely because you are using an external market as price source).


vic-en:
> 
> 
> When you add a new feature or fix a bug that's hard to replicate, please create a test case for the bug. I think this feature should have one.
> 
> See https://github.com/CoinAlpha/hummingbot/pull/1801/files and https://github.com/CoinAlpha/hummingbot/pull/1736/files for examples.
> 
> @Nullably can show you as well.

Done!

hummingbot/strategy/pure_market_making/pure_market_making.pyx
```diff
@@ -705,10 +707,16 @@ cdef class PureMarketMakingStrategy(StrategyBase):
             list new_sells = []
         top_ask = market.c_get_price(self.trading_pair, True)
         if not top_ask.is_nan():
-            proposal.buys = [buy for buy in proposal.buys if buy.price < top_ask]
+            if not self._take_if_crossed:
```
Nullably, line 710: It'd be simpler to just not call `c_filter_out_takers` if `take_if_crossed` is set.
i.e. c_tick
```
if not self._take_if_crossed:
   self.c_filter_out_takers(proposal)
```

vic-en: The call to `c_filter_out_takers` is still very necessary because there is need for the bot to submit a matching limit taker order on the current exchange with **a price that matches** existing maker orders. This is necessary in order to prevent the price on the current exchange(if it is a less liquid market) from converging to the price of the external source.

Kindly look through #1680 for better explanation.

Nullably: I had taken a look at the requirement before giving my review. The way I see it is if the order price crosses to the other side, let it matches and fills. What you did is to limit the order price to only the top bid/ask, if the top bid/ask does not have enough liquidity, the rest unfilled amount will be left as a maker on the book. 

Note that exchanges match orders by best price possible first. so an order with limit price over top bid/ask still gets filled at the top bid/ask first.

vic-en: Taking the top bid/ask and letting the extra rest as a maker was the way I saw it. I thought that would be the desired effect, just like you interpreted.

However, I get your point.
The reason I didn't implement as you see it is to prevent a great dislocation in a less liquid market and to reduce the rate of converging to the price of the external price source for a high liquid market.

Should I still go on to implement it as you see it or do you have any plans to clarify first?

Nullably: From what I understand, the main reason to set this parameter on is to use PMM also as arbitrage, users believe that the price on the PMM exchange is dislocated, as such turning maker into taker is justified. By how much is the dislocation? The simplest way is to use users bid and ask spread setting (could also be an extra parameter for this, but I think it'll get too complicated). By taking only the top bid/ask, we assume that only the top bid/ask is dislocated, also what happens when those orders are dust orders? 
But I also see your point, by converting the crosses over orders to take on only the top bid/ask makes the bot less intrusive to the market and somewhat safer mode of operation.

mifeng: I think we should go with @Nullably's interpretation of this issue, since I think that's the original intent of the request. It's possible that it might cause more dislocation in the market and weird behavior but we we won't find out unless we allow users to experiment with it.

Issue #1959, opened 2020-06-27
[BUG] Unhandled trading pair error


mifeng:
**Describe the bug**
When running `create`, process is stuck due to trading pair error. Looks like the mid price calculation

**Screenshots**
<img width="2542" alt="Screen Shot 2020-06-27 at 7 45 32 AM" src="https://user-images.githubusercontent.com/1795030/85924996-9b406200-b84a-11ea-9085-6be66e85cd74.png">

**Release version**
0.28.1

**Attachments**
[logs_hummingbot.log](https://github.com/CoinAlpha/hummingbot/files/4840708/logs_hummingbot.log)

mifeng:
Creating a config file and importing it also throws an error with `order_amount`, so there doesn't appear to be a workaround

PR #1905, opened 2020-06-22
(fix) remove order proposals with sizes below the specified order_amount

vic-en:
**Before submitting this PR, please make sure**:
- [x] Your code builds clean without any errors or warnings
- [x] You are using approved title ("feat/", "fix/", "docs/", "refactor/", etc)

**Optional**:
- Related Github issue: #1893
- Related Clubhouse Story:

**A description of the changes proposed in the pull request**:

PR #1960, opened 2020-06-27
fix / Unhandled trading pair error

vic-en:

**Optional**:
- Related Github issue: #1959
- Related Clubhouse Story:

Documentation on Kraken API isn't very detailed on rules for trading pairs.

However, I was able to scan through the results for [tradable asset pairs](https://api.kraken.com/0/public/AssetPairs) and noticed the pattern used for some trading pair.


vic-en:
> 
> 
> Unfortunately, there's still an API error with placing buy orders in the BTC-USD pair. Sell orders work, so it looks like it might be related to the USD symbol:
> Screen Shot 2020-06-27 at 3 37 03 PM
> 
> The BTC-DAI pair works properly:
> Screen Shot 2020-06-27 at 3 39 49 PM
> 
> Logs:
> [logs_conf_kraken.log](https://github.com/CoinAlpha/hummingbot/files/4841355/logs_conf_kraken.log)

The reason for this error is because the min amount for this particular pair ought to be 0.002.

hummingbot/market/kraken/kraken_constants.py
```diff
@@ -24,10 +24,14 @@
...
 ]
 
-FIAT_QUOTES = FIAT_QUOTES + ["Z" + quote for quote in FIAT_QUOTES]
+FIAT_QUOTES = ["Z" + quote for quote in FIAT_QUOTES] + FIAT_QUOTES
 
 QUOTES = CRYPTO_QUOTES + ADDED_CRYPTO_QUOTES + FIAT_QUOTES
 
+SPECIAL_BASES = ["XBT", "ETH", "XRP"]
+
+SPECIAL_QUOTES = ["USD", "USD.d", "EUR", "EUR.d", "JYP", "JYP.d", "CAD", "CAD.d", "GBP", "GBP.d"]
```
dennisocana, line 33: Hi @vic-en the correct fiat token is JPY instead of JYP.

```
2020-06-28 08:04:06,001 - 9594 - hummingbot.core.clock - ERROR - Unexpected error running clock tick.
  File "hummingbot/strategy/pure_market_making/pure_market_making.pyx", line 547, in hummingbot.strategy.pure_market_making.pure_market_making.PureMarketMakingStrategy.c_tick
    proposal =self.c_create_base_proposal()
  File "hummingbot/strategy/pure_market_making/pure_market_making.pyx", line 572, in hummingbot.strategy.pure_market_making.pure_market_making.PureMarketMakingStrategy.c_create_base_proposal
    price = self.c_get_mid_price() * (Decimal("1") - self._bid_spread - (level * self._order_level_spread))
  File "hummingbot/strategy/pure_market_making/pure_market_making.pyx", line 299, in hummingbot.strategy.pure_market_making.pure_market_making.PureMarketMakingStrategy.c_get_mid_price
    mid_price = self._market_info.get_mid_price()
  File "/home/ubuntu/hummingbot/hummingbot/strategy/market_trading_pair_tuple.py", line 33, in get_mid_price
    return self.market.get_mid_price(self.trading_pair)
  File "hummingbot/market/market_base.pyx", line 109, in hummingbot.market.market_base.MarketBase.get_mid_price
    return (self.get_price(trading_pair, True) + self.get_price(trading_pair, False)) / Decimal("2")
  File "hummingbot/market/market_base.pyx", line 316, in hummingbot.market.market_base.MarketBase.get_price
    return self.c_get_price(trading_pair, is_buy)
  File "hummingbot/market/market_base.pyx", line 215, in hummingbot.market.market_base.MarketBase.c_get_price
    return self.c_quantize_order_price(trading_pair, top_price)
  File "hummingbot/market/market_base.pyx", line 190, in hummingbot.market.market_base.MarketBase.c_quantize_order_price
    price_quantum = self.c_get_order_price_quantum(trading_pair, price)
  File "hummingbot/market/kraken/kraken_market.pyx", line 1124, in hummingbot.market.kraken.kraken_market.KrakenMarket.c_get_order_price_quantum
    TradingRule trading_rule = self._trading_rules[trading_pair]
KeyError: 'XBTJPY'
```

dennisocana:
Getting the same issue on other pairs like XMR-BTC, XDG-BTC.

```
2020-06-28 08:19:32,608 - 9823 - hummingbot.client.hummingbot_application - ERROR - 'XMRXBT'
  File "/home/ubuntu/hummingbot/hummingbot/client/hummingbot_application.py", line 148, in _handle_command
    f(**kwargs)
  File "/home/ubuntu/hummingbot/hummingbot/client/command/config_command.py", line 47, in config
    if key not in self.config_able_keys():
  File "/home/ubuntu/hummingbot/hummingbot/client/command/config_command.py", line 77, in config_able_keys
    keys += [c.key for c in self.strategy_config_map.values() if c.prompt is not None]
  File "/home/ubuntu/hummingbot/hummingbot/client/command/config_command.py", line 77, in <listcomp>
    keys += [c.key for c in self.strategy_config_map.values() if c.prompt is not None]
  File "/home/ubuntu/hummingbot/hummingbot/client/config/config_var.py", line 41, in prompt
    return self._prompt()
  File "/home/ubuntu/hummingbot/hummingbot/strategy/pure_market_making/pure_market_making_config_map.py", line 42, in order_amount_prompt
    min_amount = minimum_order_amount(exchange, trading_pair)
  File "/home/ubuntu/hummingbot/hummingbot/client/config/config_helpers.py", line 366, in minimum_order_amount
    mid_price = get_mid_price(exchange, trading_pair)
  File "/home/ubuntu/hummingbot/hummingbot/core/utils/market_mid_price.py", line 27, in get_mid_price
    return kraken_mid_price(trading_pair)
  File "/home/ubuntu/miniconda3/envs/hummingbot/lib/python3.8/site-packages/cachetools/func.py", line 74, in wrapper
    v = func(*args, **kwargs)
  File "/home/ubuntu/hummingbot/hummingbot/core/utils/market_mid_price.py", line 92, in kraken_mid_price
    record = resp_json["result"][k_pair]
KeyError: 'XMRXBT'
```

dennisocana:
Found a bug while testing this patch, not if this is related but the logs are showing a different error related to `.c_get_order_price_quantum` this time. Tested on REP-BTC pair. Still testing which pairs will show errors.

```
2020-06-28 08:04:09,001 - 9594 - hummingbot.core.clock - ERROR - Unexpected error running clock tick.
  File "hummingbot/strategy/pure_market_making/pure_market_making.pyx", line 547, in hummingbot.strategy.pure_market_making.pure_market_making.PureMarketMakingStrategy.c_tick
    proposal =self.c_create_base_proposal()
  File "hummingbot/strategy/pure_market_making/pure_market_making.pyx", line 572, in hummingbot.strategy.pure_market_making.pure_market_making.PureMarketMakingStrategy.c_create_base_proposal
    price = self.c_get_mid_price() * (Decimal("1") - self._bid_spread - (level * self._order_level_spread))
  File "hummingbot/strategy/pure_market_making/pure_market_making.pyx", line 299, in hummingbot.strategy.pure_market_making.pure_market_making.PureMarketMakingStrategy.c_get_mid_price
    mid_price = self._market_info.get_mid_price()
  File "/home/ubuntu/hummingbot/hummingbot/strategy/market_trading_pair_tuple.py", line 33, in get_mid_price
    return self.market.get_mid_price(self.trading_pair)
  File "hummingbot/market/market_base.pyx", line 109, in hummingbot.market.market_base.MarketBase.get_mid_price
    return (self.get_price(trading_pair, True) + self.get_price(trading_pair, False)) / Decimal("2")
  File "hummingbot/market/market_base.pyx", line 316, in hummingbot.market.market_base.MarketBase.get_price
    return self.c_get_price(trading_pair, is_buy)
  File "hummingbot/market/market_base.pyx", line 215, in hummingbot.market.market_base.MarketBase.c_get_price
    return self.c_quantize_order_price(trading_pair, top_price)
  File "hummingbot/market/market_base.pyx", line 190, in hummingbot.market.market_base.MarketBase.c_quantize_order_price
    price_quantum = self.c_get_order_price_quantum(trading_pair, price)
  File "hummingbot/market/kraken/kraken_market.pyx", line 1124, in hummingbot.market.kraken.kraken_market.KrakenMarket.c_get_order_price_quantum
    TradingRule trading_rule = self._trading_rules[trading_pair]
KeyError: 'REPXBT'
```

hummingbot/market/kraken/kraken_constants.py
vic-en: Done!

vic-en:
@dennisocana 
I plan to write a test case to specially test the conversion of all Kraken pairs to and from HB pairs tomorrow.

You can just retry those ones you've reported earlier and let me know if they work now.

vic-en:
> 
> 
> This test fails in Jenkins, but `development` branch passes:
> 
> ```
> test_apply_budget_constraint_to_proposal (test.test_pmm.PMMUnitTest) ... FAIL
> ```

@mifeng , I don't this the failed test is related to this PR or brach.

mifeng:
> > This test fails in Jenkins, but `development` branch passes:
> > ```
> > test_apply_budget_constraint_to_proposal (test.test_pmm.PMMUnitTest) ... FAIL
> > ```
> 
> @mifeng , I don't think the failed test is related to this PR or brach.

Thanks @vic-en. I added that comment to #1905, the right issue.

Regarding this Kraken fix, @kmcgowan2000, who wrote the connector, will review this and get back to you.

hummingbot/market/kraken/kraken_constants.py
```diff
@@ -24,10 +24,16 @@
     "AUD"
 ]
 
-FIAT_QUOTES = FIAT_QUOTES + ["Z" + quote for quote in FIAT_QUOTES]
+FIAT_QUOTES = ["Z" + quote for quote in FIAT_QUOTES] + FIAT_QUOTES
```
kmcgowan2000, line 27: this change is good. however would it mean that `convert_from_exchange_trading_pair("XTZUSD") == "XT-ZUSD"` ?

hummingbot/market/kraken/kraken_market.pyx
```diff
@@ -384,6 +398,7 @@ cdef class KrakenMarket(MarketBase):
         for trading_pair, rule in asset_pairs_dict.items():
             try:
                 base, quote = KrakenMarket.split_to_base_quote(trading_pair)
+                base = KrakenMarket.convert_from_exchange_symbol(base)
```
kmcgowan2000, line 401: this is going to mean that ETH is represented as "ETH" if it's the base asset and as "XETH" if it's the quote asset. I think we would want to convert both base and quote here?

hummingbot/market/kraken/kraken_market.pyx
```diff
@@ -251,6 +251,20 @@ cdef class KrakenMarket(MarketBase):
...
+            return exchange_trading_pair
+        all_special_symbols = constants.SPECIAL_BASES + constants.SPECIAL_QUOTES
+        if base in all_special_symbols and quote in all_special_symbols:
+            if base in constants.SPECIAL_BASES:
+                base = "X" + base
+            if quote in constants.SPECIAL_QUOTES:
+                quote = "Z" + quote
+            elif quote in constants.SPECIAL_BASES:
+                quote = "X" + quote
+
```
kmcgowan2000, line 267: are we really sure any of this is needed? all API endpoints support use of altname instead of name, which in the case of pairs containing assets like XETH, XXBT, etc. is the pair without the extra X (so using ETH, XBT, etc.).

kmcgowan2000: https://api.kraken.com/0/public/AssetPairs for reference

kmcgowan2000: ah I see, we're doing this because our internal data structures (order book) are using exchange asset names (XETH, XXBT, etc.) due to the API responses. I would prefer we change it there (so that they use ETH, XBT, etc.) rather than here, but after looking at the endpoints it seems like that may not be as simple as it should be. If this works then it's fine.

hummingbot/market/kraken/kraken_market.pyx
vic-en: Only the base is needed in this section. The base is required to get the `min_order_size` that is needed by `TradingRule`

hummingbot/market/kraken/kraken_constants.py
vic-en: I see, that is one exception. But it is best to logically let the quotes with the `Z` prefix come first so majority of all pairs can work properly with `split_to_base_quote(exchange_trading_pair)`

vic-en:
I will create a test case to process all trading pairs from and to Kraken pairs.

hummingbot/market/kraken/kraken_constants.py
kmcgowan2000: Yes definitely this change (`Z` prefix quotes first) is an improvement. We do need to have a more complete solution though.

kmcgowan2000: An ideal solution would allow us to verify that the base is actually an asset once we've taken away the possible quote. e.g. we test the split XTZUSD->XT+ZUSD, we check XT and find it isn't an asset, we continue, then we check XTZUSD->XTZ+USD, we check XTZ and find it is an asset. Perhaps caching the results of https://api.kraken.com/0/public/Assets including altnames would be one way?

hummingbot/market/kraken/kraken_market.pyx
kmcgowan2000: yes you're right

hummingbot/market/kraken/kraken_constants.py
vic-en: Exactly the main reason I added something like `if len(quote) > 2 or quote == "SC"` in `split_to_base_quote(exchange_trading_pair)`. Because `SC` is the only Asset with 2 letter.

vic-en:
I really tried to factor out all that can be factored to reduce the number of manual cases. But there's so much inconsistency with a few pairs and I tried simplifying them as best as possible. Generally, all most manually handled pairs can be found in:

- `def convert_to_exchange_trading_pair(hb_trading_pair: str, delimiter: str = "")` in the kraken_market.pyx file
```

        if quote == "USD":
            if base == "USDT":
                return "USDTZUSD"
            elif base == "EUR" or base == "GBP":
                return "Z" + base + "Z" + quote
        if base == "USD":
            if quote == "CAD" or quote == "JPY":
                return "ZUSDZ" + quote
```

- And in the kraken_constants.py file
`SPECIAL_PAIRS = ["XDGEUR", "XDGUSD", "XRPETH", "XRPGBP", "LTCETH", "LTCGBP"]`

Those are basically the pairs that are really different from the general pattern.

hummingbot/market/kraken/kraken_market.pyx
vic-en: I guess this can be marked as resolved now?

hummingbot/market/kraken/kraken_constants.py
vic-en: > 
> 
> Exactly the main reason I added something like `if len(quote) > 2 or quote == "SC"` in `split_to_base_quote(exchange_trading_pair)`. Because `SC` is the only Asset with 2 letter.

I believe just checking if the length of base is greater than 2 with the exception of "SC" asset is a better alternative than caching the result of  https://api.kraken.com/0/public/Assets and running a check on every if it exists on every available asset.

vic-en:
Removed logic in `convert_to_exchange_trading_pair` according to Kieren's suggestion as they're not really necessary when making requests. But results of the methods shouldn't be used to parse responses as trading pairs in responses are formatted by Kraken.

Issue #1917, opened 2020-06-24
[Extend price source param] Change price source logic (1/4)


mifeng:
This is part 1 of https://github.com/CoinAlpha/hummingbot/issues/1715:

## Remove `exchange` option in `price_source_type` and add `mid_price`

Currently, the `price_source_type` parameter allows users to select whether they want to use an external exchange (`exchange`) or a custom API (`custom`) to generate the mid price used for algo trading. It is only used when `price_source_enabled` is true.

Since we're going to change `price_source_type` so that it sets the **type** of price used rather than its **source**, we don't need the exchange option anymore. Make `mid_price` the default value instead of `None`.

## Adjust `price_source_exchange` and `price_source_market`

Adjust the external price source logic so that the `price_source_exchange` and `price_source_market` params determines if we use the current order book or a different order book. If either of those params is `None`, the current order book is used. If both parameters are not `None`, then the bot

## Remove `price_source_enabled`

Given the change above, this boolean is not needed anymore.

petioptrv:
Currently working on this.

Nullably:
The changes here are a little confusing to me. `price_source_type` has both source of the price (custom API) and the type of the price (mid_price or last_trade) implied in the same parameter. Would it be easier to understand if we separate the parameter into 2 parameters as below 
* price_source
  This is the source of price, the choices for this are `current_market` (default), `external_market`, `custom_api`
* price_type
  The type of price used, this can be either `mid_price` (default) or `last_trade`   

If `external_market` price source is selected, `price_source_exchange` and `price_source_market` are required. The `price_source` takes a higher priority, if `current_market` or `custom_api` is selected, ignore `price_source_exchange` and `price_source_market`.
If `custom_api` is selected `price_source_custom` is required and ignore `price_type`.
We can remove `price_source_enabled`.

petioptrv:
> @mifeng @petioptrv
> The changes here are a little confusing to me. `price_source_type` has both source of the price (custom API) and the type of the price (mid_price or last_trade) implied in the same parameter. Would it be easier to understand if we separate the parameter into 2 parameters as below
> 
> * price_source
>   This is the source of price, the choices for this are `current_market` (default), `external_market` and `custom_api`
> * price_type
>   The type of price used, this can be either `mid_price` (default) or `last_trade`
> 
> Tasks:
> 
> * Rename `price_source_custom` to `price_source_custom_api`
[...]

PR #1961, opened 2020-06-27
[v0.30] Removes price_source_enabled and adds mid_price to price_source_type

petioptrv:

**Optional**:
- Related Github issue: #1917 
- Related Clubhouse Story:


petioptrv:
@Nullably, I'd like to make it possible to disable external source after it has been configured, but couldn't find a clean way to do it. Specifically, could not set price_source_market to None if price_source_exchange was set to None through the config command. Is there a parameter combo that I can take a look at and model to do this?

Nullably:
> @Nullably, I'd like to make it possible to disable external source after it has been configured, but couldn't find a clean way to do it. Specifically, could not set price_source_market to None if price_source_exchange was set to None through the config command. Is there a parameter combo that I can take a look at and model to do this?

@petioptrv 
Sorry, I don't quite understand, can you elaborate your questions a bit more?
The  `price_source_market` and `price_source_exchange` should both have values, or both are None. 
If you mean to set `price_source_market` to None, when `price_source_exchange` is set to None, then you can do use `on_validated` e.g.:
```
"price_source_exchange":
        ConfigVar(key="price_source_exchange",
                  prompt="Enter external price source exchange name (empty to disable) >>> ",
                  required_if=lambda: pure_market_making_config_map.get("price_source_type") == "mid_price",
                  type_str="str",
                  validator=validate_price_source_exchange,
                  on_validated=lambda v: pure_market_making_config_map["price_source_market"].value = None if v is None else pure_market_making_config_map["price_source_market"].value
)
```

mifeng:
Since this is part 1 of a series of enhancements to the Price Source feature, I think we should wait until it's complete before merging this branch to `development`. It may not be done until v0.30 release.

mifeng:
Here's the branch I mentioned: https://github.com/CoinAlpha/hummingbot/tree/price-source

petioptrv:
@Nullably, @mifeng, I attempted to fix the problem of disabling external exchange, but I can see why we had a "switch" for it before. The switch made it so that when it was set to `False`, it would set to `None` all concerned config vars. Now, if the user passes an empty string to the `price_source_exchange` prompt, there is no elegant way of setting its value to `None`. 

I propose we bring back the switch as `external_price_source_enabled` and require it only if `price_source_type` is set to `mid_price`. Switching it to `False` will set `price_source_exchange`, `price_source_market` and `take_if_crossed` to `None`.

petioptrv:
> Here's the branch I mentioned: https://github.com/CoinAlpha/hummingbot/tree/price-source

Just got to that. No problem, I will roll-back merging `development` into my branch and switch the PR to `price-source`.

Nullably:
@petioptrv 
> @Nullably, @mifeng, I attempted to fix the problem of disabling external exchange, but I can see why we had a "switch" for it before. The switch made it so that when it was set to False, it would set to None all concerned config vars. Now, if the user passes an empty string to the price_source_exchange prompt, there is no elegant way of setting its value to None.

Ok, thanks, I now understand the issue. I am proposing a small change in the requirement, let me know what you think.
https://github.com/CoinAlpha/hummingbot/issues/1917#issuecomment-654936952
This will help solve problem above, we only need to validate `price_source_exchange` input if `price_source` is set to `external_market`.

petioptrv:
@Nullably, @mifeng, the changes as discussed in #1917. I have not included `price_type` as that is part of a subsequent issue.

Issue #1970, opened 2020-06-30
[Feature] Misc UX fixes


mifeng:
Per comment in #1942, changing the create command to use `ping_pong` instead of `inventory_skew` is a good change, but I think we should make some misc UX changes along with it:

1. Use the logic in `asset_ratio_maintenance_prompt()` to make `config inventory_target_base_pct` question better. Seeing the current ratio helps users answer the question

When users config `inventory_target_base_pct`, the question should be the same as what it was when it was in the `create` walkthrough.

2. Currently `inventory_target_base_pct` defaults to None. Set default value of  `inventory_target_base_pct` to 50

3. Add more helpful text to `ping_pong` question since it's now part of basic `create` flow: `"Would you like to use the ping pong feature and alternate between buy and sell orders after fills?"`

mifeng:
To clarify, this doesn't add inventory skew back to `create`. Instead, change #1 is intended to add more descriptive detail to the `config` command when it is used to set the target base pct.

Right now, the question just asks "What is your target base pct?" By adding more info that tells the user the current percent, it makes it easier to set the right target percent.

PR #1986, opened 2020-07-03
feat / miscellaneous ux fixes

vic-en:

**Optional**:
- Related Github issue: #1970 
- Related Clubhouse Story:


vic-en:
> 
> 
> Tested and did not run into issues during `create`. Balances are also displayed correctly during `inventory_target_base_pct` prompt.
> 
> One minor note @vic-en , we used to be able to reconfigure this parameter by doing `config inventory_target_base_pct [value]` in the current development branch. See screenshot below.
> 
> image
> 
> In this patch, this behavior is changed - is this expected?
> 
> image
> 
> Reference : https://docs.hummingbot.io/operation/commands/config/#config-key-value

@dennisocana  Thank you for calling my attention to this.
I also see in the initial code that the notification message `"Please follow the prompt to complete configurations:"` is printed when a value is supplied. I believe it should only be printed when no value is supplied. I'll fix that too.

Issue #1848, opened 2020-06-10
[Research] Assess Infura usage when using Hummingbot on DEXs


mifeng:
Due to recent changes in rate limits by Infura, user report that Hummingbot clients running PMM with basic settings now hit the daily 100k limit in approx 21 hours.

Assess how severe this issue is by running various bot configurations for 1 hour each and assessing how many Infura calls are made over that time. The main params that affect number of Infura calls should be:
- number of order levels
- order refresh time
- other params that increase number of markets tracked or API calls made

wpenn:
For Infura's free "Core" account,  they restrict the number of Requests/Day to 100,000. Otherwise, you have to pay at least $50/month. 

We are therefore restricted to about 4,166 requests per hour. After some testing with our PMM strategy on Radar Relay, our client calls Infura ranging from 4000 - 6000 times per hour, depending on the configuration. `order_refresh_time` of less than 60 seconds and an `order_level` greater than 1 easily results in a rate higher than 4166. 

**Top Infura Method Calls:**
- `eth_getLogs`: 39.38%
- `eth_getBlockByNumber`: 19.47%
- `eth_call`: 14.91%
- `eth_getBlockByHash`: 12.71%
- `eth_getBalance`: 5.58%
- `eth_gasPrice`: 5.50%
- `eth_accounts`, `eth_getTransactionCount`, `eth_blockNumber`: < 2.00%

**Infura Usage Data from Past 24-hours**
(Only ran bot for 7 Trials [19 hours total Maximum])
<img width="1091" alt="Screen Shot 2020-06-12 at 4 55 37 PM" src="https://user-images.githubusercontent.com/20906874/84545607-8e116800-accd-11ea-895f-cdba7c091028.png">

gabririgo:
a hint for reducing smart contract calls is using helper contracts like the following any time multiple requests are sent for the same operation:
[https://github.com/RigoBlock/rigoblock-monorepo/blob/master/packages/contracts/src/examples/helpers/HGetMultipleBalances.sol](url)

Ofc logs, which are the biggest number of calls here, can be reduced only by substituting log queries with transaction queries and filters, or, as already discussed with @mifeng , by properly optimizing the code.

PR #1995, opened 2020-07-06
fix/reduced infura consumption with websocket connection

wpenn:

Reduced infura consumption below 100,000 per day limit with WebSocket connection to replace watcher for new blocks
- Clubhouse Ticket: https://app.clubhouse.io/coinalpha/story/9071/reduce-infura-api-calls-on-hb-client
- Github: #1848 

Need someone to also to test if the consumption is indeed below the rate (4166/hr) for them


dennisocana:
Ran a bot overnight with 3 orders per side. Total method calls are at 39,856 for 10 hours.

image
image

I will stop the instance for now since I ran into errors due to Radar Relay's request limit.

[logs_rr_weth-dai.log](https://github.com/CoinAlpha/hummingbot/files/4887617/logs_rr_weth-dai.log)

dennisocana:
OK to test

hummingbot/templates/conf_global_TEMPLATE.yml
```diff
@@ -45,6 +45,7 @@ celo_password: null
 # Ethereum wallet address: required for trading on a DEX
 ethereum_wallet: null
 ethereum_rpc_url: null
+ethereum_rpc_ws_url: null
```
Nullably, line 48: We need to increase the template_version by one every time we add a new param.

hummingbot/wallet/ethereum/watcher/websocket_watcher.py
```diff
@@ -0,0 +1,169 @@
...
+                        if incoming_block is not None:
+                            new_block: AttributeDict = await self.call_async(self._w3.eth.getBlock,
+                                                                             incoming_block.get("hash"), True)
+                            self._current_block_number = new_block.get("number")
+                            self._block_cache[new_block.get("hash")] = new_block
+                            self.trigger_event(NewBlocksWatcherEvent.NewBlocks, [new_block])
+                except asyncio.CancelledError:
+                    if self._network_on:
+                        self.logger().network("Network closed abnormally, reconnecting ...")
+                        self.connect()
```
Nullably, line 140: Not sure if we should `connect()` here on this exception, seems to me the task is cancelled and exit from here. To connect here might be unnecessary.

hummingbot/wallet/ethereum/watcher/websocket_watcher.py
```diff
@@ -0,0 +1,169 @@
...
+        nonce = await self._send(emit_data)
+        raw_message = await self._client.recv()
+        if raw_message is not None:
+            resp = ujson.loads(raw_message)
+            if resp.get("id", None) == nonce:
+                self._node_address = resp.get("result")
+                return True
+        return False
+
+    async def unsubscribe(self, params) -> bool:
```
Nullably, line 109: Do we need this? Is it used anywhere?

hummingbot/wallet/ethereum/watcher/websocket_watcher.py
wpenn: The Websocket sometimes "abnormally" disconnects, for example when my computer sleeps with the bot running; this accounts for the "abnormal disconnections" when there's a CancelledError but the WebSocket is still intended to be connected; this replaces the self._client with a connected WebSocket. Let me know if I should figure out a workaround or just remove it

hummingbot/wallet/ethereum/watcher/websocket_watcher.py
wpenn: It's not used, but would be necessary if this became generalized for handling WebSockets. I can remove it

hummingbot/wallet/ethereum/watcher/websocket_watcher.py
Nullably: Does the `CancelledError` raise on those situations, like when the websocket abnormally disconnects? From what I understand the CancelledError is raised only when `cancel()` is called, and it is a way to wrap up and finish the task. If you need to reconnect, it's fine but I don't think this is the right place for it.

hummingbot/wallet/ethereum/watcher/websocket_watcher.py
```diff
@@ -0,0 +1,169 @@
...
+                                                          "Check wallet network connection")
+                except BlockNotFound:
+                    pass
+                except ConnectionClosedOK:
+                    raise
+                except Exception as e:
+                    self.logger().network(f"Error fetching new block: {e}", exc_info=True,
+                                          app_warning_msg="Error fetching new block. "
+                                                          "Check wallet network connection")
+        except asyncio.CancelledError:
```
Nullably, line 154: This except and raise here is unnecessary, you need to do this only when you also catch other kinds of Exceptions.

hummingbot/templates/conf_global_TEMPLATE.yml
Nullably: What I mean is you need to update `template_version` (line 6 of this file) to 9.

hummingbot/wallet/ethereum/watcher/websocket_watcher.py
wpenn: Made a mistake. Wrong location; I put it in ConnectionClosedError from WebSockets instead. This is for when connectionclosed errors codes are not 1000 (OK) or 1001 ("Going Away"). Trying to figure out if there's a way to check connection another way in the fetch loop

wpenn: @Nullably pushed up changes, what do you think?

hummingbot/wallet/ethereum/watcher/websocket_watcher.py
Nullably: You don't need to except asyncio.CancelledError here, but you do need one in your inner try-except block.

Issue #2006, opened 2020-07-08
[Bug] orders are rejected when take_if_crossed is set on Binance PMM


Nullably:
**Describe the bug**
The crossed orders are rejected with the below message
```
APIError(code=-2010): Order would immediately match and take.
```

**Steps To Reproduce**
- Start a PMM strategy on Binance that uses external price source and has `take_if_crossed` parameter switched on.

**Desired Effect**
- When the parameter is switched on, the strategy uses LIMIT order type instead of LIMIT_MAKER.
- Additionally, this parameter should only be used when external price source is specified.

PR #2009, opened 2020-07-08
fix / orders are rejected when take_if_crossed is set on Binance PMM

vic-en:

**Optional**:
- Related Github issue: #2006 
- Related Clubhouse Story:


Nullably:
> Additionally, this parameter should only be used when external price source is specified.
What I thought was take_if_crossed can only be set to True if external price source is specified. You can achieve this doing configuration (when validating take_if_crossed value).

vic-en:
> 
> 
> > Additionally, this parameter should only be used when external price source is specified.
> 
> What I thought was take_if_crossed can only be set to True if external price source is specified. You can achieve this during configuration (when validating take_if_crossed value).

I thought you wanted it in the strategy itself as I saw that was already in place already in the configuration.
I would go on to remove the extra check I just added then.

PR #2087, opened 2020-07-22
Feat / last trade price

Nullably:

**Optional**:
- Related Github issue: https://github.com/CoinAlpha/hummingbot/issues/2036
- Related Clubhouse Story: 

- Add last_trade_price to order_book
- Last trade price is updated from both web socket and REST API


hummingbot/core/data_type/order_book_tracker.py
```diff
@@ -136,11 +139,34 @@ def stop(self):
...
+    async def _update_last_trade_prices_loop(self):
+        '''
+        Updates last trade price for all order books, this is fall-back mechanism for when the web socket
+        has not yet started receiving messages or fails for some reasons.
+        '''
+        await wait_til(lambda: len(self.data_source._trading_pairs) == len(self._order_books.keys()))
+        while True:
+            try:
+                for trading_pair, order_book in self._order_books.items():
+                    if order_book.last_trade_price_updated >= time.perf_counter() - (60. * 3):
```
vic-en, line 159: 1. I believe you want this to only run if `last_trade_price` isn't updated within 3 minutes. I don't see anywhere else `last_trade_price_updated` is updated aside in `c_apply_trade()`(which is only used for simulating trades). I believe `last_trade_price_updated` should be updated later in this block. It should also be updated after `last_trade_price` is also updated over websocket.

2. Also, I can't see where `last_trade_price ` is updated over websocket. I guess you plan to do that in subsequent commit for this PR.

hummingbot/core/data_type/order_book_tracker.py
```diff
@@ -136,11 +139,34 @@ def stop(self):
...
+                        continue
+                    last_price = await self.data_source.get_last_traded_prices([trading_pair])
```
vic-en, line 161: Also, this will result in unnecessary multiple call to the ticker endpoint for connectors like huobi and eterbase. I see those connectors supply last price of all available pairs in a single call. Since by design, `get_last_traded_prices()` takes in a list of pairs, why not just do away with the for-loop on line 158 above and simply pass all trading pairs at once?

hummingbot/core/data_type/order_book_tracker.py
Nullably: `c_apply_trade()` is used for when the bot is listening to trades updates through web socket channel, if the web socket is configured and connected properly, there is no additional work required for this. So the `last_trade_price` is simply whatever the last trade update. I will rename `last_trade_price_updated` to `last_applied_trade` to avoid confusion. `_update_last_trade_prices_loop` is a fall back mechanism for when web_socket is failing, it checks every 3 seconds if `last_trade_price_updated` is over 3 minutes it will use REST API to update the last trade price. The reason `last_trade_price_updated` is not updated here is because it will update the price every 3 seconds til the web socket mechanism kicks in. Also note that REST API is a bit slower than ws API, to avoid concurrency problem, it is designed to be used only as a fall back mechanism.

hummingbot/core/data_type/order_book_tracker.py
Nullably: `last_trade_price_updated` is per order book, hence I used in the for-loop, and so far we don't have any strategy/configuration that uses multiple order book from the same exchange yet.
Though, your comment is valid, I will update this part to check for the oldest `last_trade_price_updated` and update them all.

hummingbot/core/data_type/order_book_tracker.py
vic-en: I see, nice job!

hummingbot/core/data_type/order_book_tracker.py
vic-en: Nice work!

PR #2090, opened 2020-07-22
docs/ add net delta explanation

sigit2306:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:

Add explanation about Net Delta and Trade Delta on History page https://docs.hummingbot.io/operation/commands/history


documentation/docs/operation/commands/history.md
```diff
@@ -81,4 +81,29 @@ Using the same example, get the starting quote value. Convert the starting ETH a
 ```
 Return % = -0.562 USDT / 2539.13 USDT
 Return % = -0.0222 %
+```
+
+
+## Net Delta
+
+Net Delta is calculated based on the difference of starting balance and current balance, while Trade Delta is tokens acquired after executing trades. Both Net Delta or Trade Delta can be positive or negative (profit or loss).
```
dennisocana, line 89: Net Delta is the difference of your starting balance and current balance i.e. it takes into calculation assets spent & acquired during trades, deposits and withdrawals.

Trade Delta does not take into account deposits and withdrawals, only balance changes after executing trades in Hummingbot.

documentation/docs/operation/commands/history.md
```diff
@@ -81,4 +81,29 @@ Using the same example, get the starting quote value. Convert the starting ETH a
...
+
+$Net Delta = Starting - Current$
```
dennisocana, line 91: To make it consistent across this page, use lower case just like the rest.

```suggestion
$net_delta = $starting_balance - $current_balance
```

documentation/docs/operation/commands/history.md
```diff
@@ -81,4 +81,29 @@ Using the same example, get the starting quote value. Convert the starting ETH a
...
+
+```
+Inventory:                                                                 
+      Market Asset  Starting   Current Net Delta Trade Delta               
+  0  binance   ETH   10.0000   10.0000    0.0000      2.0000               
+  1  binance  USDT  500.0000  499.9840   -0.0160   -407.7920               
+                                                                           
+```
+
+Using the same example, after executing these trades we can calculate Net Delta for base and quote assets, so Net Delta for base asset is **0.0000** because we started with `10.0000 ETH` and current also `10.000 ETH`. Net Delta for quote asset is **-0.0160 USDT** because we started with `500.0000 USDT` and current is `499.9840 USDT`, it means we obtained lost of **-0.0160 USDT** on quote asset.
```
dennisocana, line 101: "it means we ~~obtained~~ lost ~~of~~ 0.0160 USDT on quote asset."

documentation/docs/operation/commands/history.md
```diff
@@ -81,4 +81,29 @@ Using the same example, get the starting quote value. Convert the starting ETH a
...
+
+```
+Base asset Net Delta = 10.000 ETH - 10.0000 ETH
+Base asset Net Delta = 0.0000 ETH
+
+Quote asset Net Delta = 50.000 USDT - 499.9840 USDT
+Quote asset Net Delta = -0.0160 USDT
```
dennisocana, line 108: These are not needed since the examples and explanation above will suffice.

documentation/docs/operation/commands/history.md
sigit2306: Noted, thanks.

documentation/docs/operation/commands/history.md
sigit2306: Noted, thanks.

documentation/docs/operation/commands/history.md
sigit2306: Noted, thanks.

documentation/docs/operation/commands/history.md
sigit2306: Noted, will revise, thanks.

Issue #2049, opened 2020-07-15
[BUG] high cpu usage after upgrade to v0.29.0 (public beta) using a DEX


gabririgo:
**Describe the bug**
After upgrading to 0.29.0 CPU usage for 1 hummingbot instance run inside a docker container goes from 2% to 50%. This happens both when upgrading old configuration files and with a fresh configuration of a new bot.

**Steps To Reproduce**
upgrade hummingbot with ./update.sh script, run the bot and you will soon see the performance issue.
The process that gives problems seems to be python3.

Here are the steps to reproduce the issue (see attachments in the section below):
1. Set up your configuration file with these parameters: wethdai, 1bps spread, 0.4WETH, ping-pong true, ethereum https and wss
2. Start the bot using this command: any
3. check CPU usage with top -i command
4. after a few minutes you'll be able to see the higher CPU usage in aws if you're using it

...

**Screenshots**
![Screenshot from 2020-07-15 09-38-19](https://user-images.githubusercontent.com/12066256/87517720-8e5fa300-c67f-11ea-95b6-64a258a3f09d.png)
![Screenshot from 2020-07-15 09-42-41](https://user-images.githubusercontent.com/12066256/87517745-9586b100-c67f-11ea-87b3-c1df2215a94b.png)

**Release version**
0.29.0

**Attachments**
[logs_conf_pure_mm_1.log](https://github.com/CoinAlpha/hummingbot/files/4924669/logs_conf_pure_mm_1.log)
[logs_conf_pure_mm_1.log](https://github.com/CoinAlpha/hummingbot/files/4924690/logs_conf_pure_mm_1.log)
logs_hummingbot.log file is empty as this is a fresh bot and does not crash. Same happens also when upgrading an existing bot with different config file.

*WARNING: Do NOT publish any exchange API keys or your wallet's private key. Whoever has access to them may steal your assets!*

**Optional: your discord username:**
gabrigo | RigoBlock

mifeng:
@gabririgo thanks for raising this - a couple of questions that may help us figure out what happened: 
* did you try this on both DEX and CEX or only on DEX?
* was the CPU level stable at 50% or does it fluctuate as in the image above?

gabririgo:
> @gabririgo thanks for raising this - a couple of questions that may help us figure out what happened:
> 
> * did you try this on both DEX and CEX or only on DEX?
> * was the CPU level stable at 50% or does it fluctuate as in the image above?

- DEX only so far
[...]

PR #2098, opened 2020-07-23
fix / high cpu usage with DEXes

vic-en:
…likely disconnection

**Optional**:
- Related Github issue: #2049 
- Related Clubhouse Story: 


gabririgo:
not sure if this is already in the code, given the goal of sticking into an infura limit and preventing the bot from ddossing itself, is it possible we have an edge case when the client continuously disconnects and tries to reconnect? Would it be a good idea to set a timestamp counter and sleep for something like 10 seconds in this remote case?
Btw thank you for fixing this!

PR #2127, opened 2020-07-30
[v0.31] feat/ maintain user balance in market

### LINKED ISSUE #2012: [BUG] Orders not created caused by broken balance update
**Describe the bug**

So far I only tested this while running on Liquid and Coinbase Pro but not happening on Binance. Some orders are not created during certain cycles because it thinks I don't have enough balance to place the order.

For example, in the first cycle all orders are created.

putty_5hjUOIFtV0

In the next order refresh, it says I have insufficient balance - which is incorrect.

jKZR4eATte

This is running on Liquid exchange, same bug is happening. First cycle all orders are created.

putty_GWFgSnOsZf

In the next order refresh, some orders are not created.

CUPWgFrZQI

**Steps To Reproduce**

1. Run a pure market making bot on Coinbase Pro or Liquid
2. Set order amount to just 60% or above of your total balance on each side
3. Wait for orders to refresh

**Release version**

dev-0.29.0

```
"branch": "development",
"commit": "9dfdc1d5dc18e09364e8868dded42d9f8421fb90",
"date": "2020-07-09T06:40:46Z",
"maintainer": "CoinAlpha, Inc. <dev@coinalpha.com>"
```

**Attachments**

[logs_cbp_link-eth.log](https://github.com/CoinAlpha/hummingbot/files/4896543/logs_cbp_link-eth.log)
[logs_liquid_qash-eth.log](https://github.com/CoinAlpha/hummingbot/files/4896545/logs_liquid_qash-eth.log)
[logs_binance_mft-eth.log](https://github.com/CoinAlpha/hummingbot/files/4896654/logs_binance_mft-eth.log)

### LINKED ISSUE #2017: [Feature] Maintain user account balances in market connectors
**Feature Suggestion**
Currently the bot relies solely on the exchange for current state of user balances. For most exchanges, we use REST API periodically to get the information, so during the gap between each interval the current state of balances can be out of synced. This causes bugs like https://github.com/CoinAlpha/hummingbot/issues/2012

The feature is for the bot to maintain user account balances itself as below.
1. Get current available balances, apply a limit on how much it can see and use #1443
2. Maintain the balances by 
    - Subtract order amount from the available balance when creating a new order.
    - Add an amount back to the balance when order is filled, canceled or failed.
3. Periodically sync the balances to the current balance on the exchange. 

For exchanges that offer web socket API for user's balances, this feature is not needed.

We can create a generalised solution that can be applied to all exchange connectors,, ie. each connector shouldn't have to implement this.

CrimsonJacket:

**Optional**:
- Related Github issue:
Closes #2017 
Fixes #2012 
- Related Clubhouse Story:
9541


hummingbot/market/bamboo_relay/bamboo_relay_market.pyx
```diff
@@ -475,8 +487,17 @@ cdef class BambooRelayMarket(MarketBase):
                     locked_balances[currency] = locked_balances.get(currency, s_decimal_0) + amount
 
                 for currency, balance in total_balances.items():
-                    self._account_available_balances[currency] = \
-                        Decimal(total_balances[currency]) - locked_balances.get(currency, s_decimal_0)
+                    asset_limit = exchange_limits.get(currency.upper(), None)
```
Nullably, line 490: It'd be better if this part is generalised, ie. have a function called `apply_balance_restriction` in `market_base` class. For all subclasses that need to do this can call the function after update its balance (from API).

CrimsonJacket: Will do

Issue #1911, opened 2020-06-23
[Feature] Deprecate Market Order Type


Nullably:
**Feature Suggestion**
Simply put, there is no use for Market order type in automated trading system.
Limit order will cover most cases, if you want your order to be filled immediately, simply set the price which will likely be filled.

Using market order type has following disadvantages:
- It makes strategy vulnerable to attack, especially for open source project like ours. For example, a counter trading strategy to Arbitrage strategy can be achieved by flashing a maker order (placing order and immediately canceling it) and having another order waiting to be filled at advantaged price.
- Market order has no price, in our code base we have to cater for 2 different data structures for orders.

**Tasks / Implications:**
- All current market connectors, `order_book`, `strategy_base` and their unit tests need refactoring 
- XEMM is still currently using Market order type on the taker side, we'll need to update this to use Limit order - similar to how we update Arbitrage strategy
- Update `order_book.c_get_fee` to remove order_type argument and added in `is_maker` argument, this will require update to `Hummingsim` as well

Nullably:
After discussion with @vic-en DEXs e.g. radar relay or bamboo relay use Market order type to submit a taker order, and in order to keep terminology consistent with how DEXs use them (Limit order means putting order on the order book - maker order, and Market order means taker order), I think it's better that we still keep Market order type around.

- Add `supported_order_types` function on each exchange connector, generally for CEXs the function should return `LIMIT` and `LIMIT_MAKER` (if the exchange supports it), and for DEXs - LIMIT and MARKET

- Remove MARKET order type support on CEXs and remove related unit test cases

- Keep strategy order tracker the same, but add comments/doc string explaining tracked_market_orders are used by DEXs

PR #2167, opened 2020-08-06
[v0.30] Feat / Outstanding part for Deprecating MARKET OrderType in CEXs

vic-en:
…fillied LIMIT_MAKER orders

- [x] Fix all mock tests
- [x] Modify XMM to work with recent changes
- [x] Update documentation

**Optional**:
- Related Github issue:  #1911
- Related Clubhouse Story: [ch9539]

Continuation of https://github.com/CoinAlpha/hummingbot/pull/2133


hummingbot/market/market_base.pyx
```diff
@@ -389,7 +389,7 @@ cdef class MarketBase(NetworkIterator):
         return self.c_quantize_order_amount(trading_pair, amount)
 
     def supported_order_types(self):
-        return [OrderType.LIMIT]
```
vic-en, line 392: @Nullably 
I notice you changed this to just return OrderType.LIMIT in a previous PR.
I changed it to return both OrderType.LIMIT and OrderType.MARKET.
The reason is that returning just OrderType.LIMIT causes DEXs or PaperTrade to misbehave in XEMM because it can't differentiate maker from taker. But adding OrderType.MARKET makes it use LIMIT as maker and MARKET as taker for those connectors. Let me know if you think this change would break any part of the code so we can agree on a lasting solution for all cases.

Nullably: since you're making PR for the hummingsim, i think it's best this function raises NotImplemented in the base class, s.t. all sub classes need to implement this, including the `market` class in the hummingsim. And there, you can return all the 3 types of the order.

Nullably: The main reason that I removed the LIMIT_MAKER from this is to get tests passed, hummingsim doesn't understand LIMIT_MAKER, it will just raise error.

PR #2262, opened 2020-08-27
[v0.33] OKEx connector

martinvol:
Added OKEX/OKCoin support

TODO:
* Currently supports OKEx, changing a fews endpoints it should support OKCOIN
* Needs testing
* Type and style checks had been omitted during development, so those checks fail


martinvol:
closed in favor of  #2462

Issue #2314, opened 2020-09-11
[Feature] Pure Market Maker strategy using the inventory cost as mid price for new orders


phbrgnomo:
**Feature Suggestion**
Implement a simpler and less-risk strategy that considers the average inventory buying cost as the reference for new orders.

Every time a bid type order is filled, the bot updates a new variable called `inventory_price`, which will store the volume-weighted average price of the inventory, and the strategy will use this value as the reference to define the ask and bid price of new orders, instead of the market mid-price.

Example:

First, the bot would have to set a price for the present inventory, and we have two options for that:

1 - The user defines the inventory cost starting price
2 - If there is no price entered, the bot could use the mid market price when `start` is entered

Either way, two values should be stored:

`inventory_size` - the number of assets on the account at the start of the bot run;
`inventory_price` - the price used on one of the two cases above

Then calculate:

`inventory_cost = inventory_size * inventory price`

The bot proceeds to send the bid and ask orders to the exchange using inventory_price as the mid-price, where:

`ask_price = inventory_price + inventory_price * ask_spread`
`bid_price = inventory_price - inventory_price * bid_spread`

When a bid order is filled:

`inventory_size = inventory_size + order_amount`
`inventory_cost = inventory_cost + (order_amount * bid_price)`

`inventory_price = inventory_cost / inventory_size`

When a ask order is filled:

`inventory_size = inventory_size - order_amount`
`inventory_cost = inventory_cost - (order_amount * ask_price)`

`inventory_price = inventory_cost / inventory_size`

When the new order cycle starts, repeat that calculation of `ask_price` and `bid_price` with the new value of `inventory_size`.

There is also the need for a check to compare the inventory_price with the market mid-price, to avoid sending orders as market taker:

`IF inventory_price > market_mid_price:

`ask_price = inventory_price + inventory_price * ask_spread`
`bid_price = market_mid_price - market_mid_price * bid_spread`

`IF inventory_price < market_mid_price:

`ask_price = market_mid_price + market_mid_price * ask_spread`
`bid_price = inventory_price - inventory_price * bid_spread`

**Impact**
This strategy is easier to understand, especially by those new to market making, since the bot will only create orders that will be profitable relative to the buying cost.

**Additional context**
It doesn't have to be a new strategy, but a trigger inside the pure_mm strategy, like ping_pong.

bitphage:
@phbrgnomo is there a typo in section "When a ask order is filled:", should be

[...]

PR #2332, opened 2020-09-15
feat / mid price from own last trade

bitphage:

**Optional**:
- Related Github issue:
- Related Clubhouse Story:

Adds a new mid price option to use last own trade price as a midprice.

## Notes

* This is a simple implementation uses a variable in PMM strategy. I also tried to add last own trade price tracking to `ExchangeBase` class where other mid prices are, but it appeared that it requires addition of market subscription events etc. 
* Another more proper place where last own trade price could be kept is `StrategyBase` class as it has `c_did_complete_xxx_order` methods supposed to be overridden. `XxxOrderCompletedListener` may call a method which stores last own trade price.
* I added a separate testcase because adding new price option to `test_basic_one_level_price_type()` was too messy and `self.simulate_maker_market_trade()` method failed to fill all eligible orders at once.


bitphage:
One caveat: in `status` report `Orders:` are uses `price = self.get_price()` call to get price for spread calculation, it means for any price source / price type it uses referenced price and not actual market `mid_price`. Maybe this should be corrected.

Nullably (approved):
The code looks good to me.
I agree with having last_own_trade_price in the PMM strategy (as opposed to in the exchange connector) for now, it seems to be strategy specific and we might not need it for other strategy.
Pending testing by the QA team, this one can be merged.

dennisocana (commented):
The `order_refresh_tolerance_pct` is not working when used with `last_own_trade_price`. After a fill, subsequent orders are not cancelled when they breach the tolerable change in spreads. This behavior is not present when used with other price reference points such as mid price, last trade, best bid and best ask.

Below is an order created after a fill at 0.56% spread.

image

With `order_refresh_tolerance_pct` set to `0.01`, the order should have been refreshed when the spread went above 0.57% or below 0.55%

In the image below the spread of the active order is already at 0.63%

image
image

Log file: [last_own_trade_price.log](https://github.com/CoinAlpha/hummingbot/files/5264679/last_own_trade_price.log)

Below is the log file when using other price types that shows subsequent orders are cancelled as expected.

[logs_test_binance.log](https://github.com/CoinAlpha/hummingbot/files/5264776/logs_test_binance.log)

Nullably:
The PR changed how spreads are calculated on the status command. It used to be spreads from the reference price (one of the price type) and now are spreads from the market mid price. 
There are pros and cons to this, e.g. it is harder to see if spreads adjusted correctly to the ref price but the user can see clearer how their prices are to the mid price.

Nullably:
@dennisocana Is the bug on orders not refreshing something to do with the change in spread display?
The spreads you now see are from mid price, not last own trade price.

bitphage:
@dennisocana oh I think `order_refresh_tolerance` doesn't make sense with `price_type: last_own_trade_price` because refreshed orders will be placed at the same prices anyway (based on last trade price).
So, the logic flow is:
1. After `start`, `mid_price` will be a market mid price, `order_refresh_tolerance` should work as intended
2. An order is filled
3. `mid_price` becomes fixed and is equal to filled order price. `order_refresh_tolerance`  has no effect now.

Another thing what's concerns me is that last trade price is not being kept across restarts, while it's actually should be.

dennisocana:
@Nullably yes, thank you very much for explaining. I get it now, since the price reference is the last trade price in my case it's 344.01 then the spread of subsequent order at price 345.73 won't change since the last trade price is constant. The confusion is how the `status` output displays the spreads of the orders because it's the spread to mid price as you mentioned.

dennisocana:
Maybe it's a good idea or feature to display in the `status` output whichever price reference the strategy is using.

Nullably:
As discussed during stand-up
- The spreads should be from the reference price (whatever price type user selected), not from mid price
- The mid price column on Markets section of `status` should be the selected price type

Nullably (changes requested):
I'll add in the 2 changes

bitphage:
Hmm with these 2 changes `status` output will be pretty confusing at least for `last_own_trade_price`.

Here Markets and Orders both using market mid price, not reference last own trade price:
Selection_042

This reflects current market situation very well.

Nullably:
I updated the PR, for when last trade is selected, the Markets section now shows an extra column `Last Trade`, likewise for last own trade, a `Last Own Trade` extra column is added.
The spread is now calculated from the ref price.

Nullably:
@bitphage `spread` from our definition is a spread from reference price, i agree it's more useful for Last Own Trade to show spread from mid price rather than last own trade, but this changes what spread is. 
Maybe we could change the column name to `spread from mid` for last own trade?

phbrgnomo:
I finished running the strategy for a few days, and I am processing the data to track how the orders are being placed and the interaction with `hanging_orders`.

Found some issues to be checked:

- It seems that when a sell order is filled, the `last_own_price` isn't being used.
I was using 0.5% on both sides, and every time a bid order was filled, it was easy to see using `status` that the total spread were >1%
But when an ask order was filled, the next pair of orders the total was =1%.
Will have more information after processing the data.

- The same situation seems (ignoring the last `last_own_price` to place new orders) seems to happen when a full cycle of `order_refresh_time` passes with no orders filled:

1 - First aks and bid orders, uses market mid price
2 - one of the orders is filled
3 - after `filled_order_delay`, the next bid and ask orders price uses the `last_own_price`
4- If there is no fill, after `order_refresh_time` the next 2 orders get back to use market mid price

- There were times that orders were one added on side (ask OR bid)

Haven't fully checked yet, but it can be related to the interaction with `order_refresh_tolerance` mentioned by @dennisocana 

- The bot is only using the price of the last trade, but in the long run, this kind of invalidates the "risk protection", because it isn't considering the accumulated cost of the asset on the trading session.

Like i wrote on Issue #2314 , i think the accumulated inventory cost is what should be tracked:

`inventory_price = inventory_cost / inventory_size`

The price of each order should be calculated based on `inventory_price` that would change over time as orders get filled, what would be changing `inventory_cost` and `inventory_size`

mifeng:
@Nullably I tested this branch. To avoid confusing users, I propose changing the `status` display a bit:
* Rather than both `Mid Price` and `Last Trade` / `Last Own Trade`, I think we should just list a single `Ref Price (TYPE)` field in the the **Markets** section
* If `price_source` is `external_market`, I think we should display the external market as well
* Only display the `Ref Price` field for the market used to generate the price
* To save space for smaller terminal windows, remove the word `Price` from `Best Bid Price` and `Best Ask Price`

@phbrgnomo I think your inventory cost option for `price source` is a good one. Generally, giving users more ways to construct reference price is important. In addition, I think we should allow using an imputed price from 2 external markets, which would allow users to more easily market make for pairs like MFT/BNB which are very toxic if you just use the mid price. Since we're approaching code freeze, I think those stories should be in future releases.

Nullably:
@mifeng I have updated the PR as per requested.
image

Nullably (approved):
LGTM, pending testing by the QA

mifeng (approved):
LGTM

rxlxrxsx:
No other issues found during testing by QA.

bitphage:
@phbrgnomo I'm not sure why you got such results. For me, it works as expected - `last_own_price` used all the time after first fill for both buy/sell orders, and no price reset to mid price happening after `order_refresh_time`. Maybe you had a feeling of wrong behavior  because of a way that `status` command displayed spreads etc, so if you haven't, I would advise to try building from recent commit of this PR and see how it goes.

To confirm what price is being used, a `self.logger().info()` calls could be added in a custom build, for example, in `hummingbot/strategy/pure_market_making/pure_market_making.pyx` `get_price()`

#2314 is pretty cool btw.

PR #2354, opened 2020-09-21
(doc) add a note to the balancelimit to warn user not to set value to zero

thomasyit:
Screenshot 2020-09-21 at 11 09 10 AM


dennisocana (dismissed):
LGTM

dennisocana (approved):

dennisocana (changes requested):
Actually some final thoughts while the PR is updating to base branch:

- Change note title to "Note: Disabling balance limit"
- Change description to:

To disable balance limit, you need to remove the balance limit from the **conf_global.yml file**. Setting it to 0 from the Hummingbot client e.g. `balance limit binance USDT 0` will not place orders for that particular asset.

We've had user requests in the past to allow placing of orders on 1 side only which this feature will address.

dennisocana (approved):
LGTM

PR #2502, opened 2020-10-22
[v0.33] feat / connector status indicator

dennisocana:

- Add a Connector Status column in `connect` command output
- Add a warning message on `start` if the exchange connector used is in YELLOW or RED status
- Add status to README.md with a section "Connector status" that indicates what the colors mean


dennisocana:
To-do:

1. Use actual status values in `get_connector_status`
2. Add in README.md file

The formatted text styling doesn't seem to be working in the output pane as it always displays the ANSI or hex codes as a string output. Will look into this further.

hummingbot/connector/connector_status.py
```diff
@@ -0,0 +1,32 @@
...
+        'ethereum': 'ok',
+        'huobi': 'unavailable',
+        'kraken': 'warning',
+        'kucoin': 'unavailable',
+        'liquid': 'warning',
+        'loopring': 'ok'
+    }
+    if connector_name not in connector_status.keys():
+        status = "Unknown"
+    elif connector_status[connector_name] == 'unavailable':
```
Nullably, line 26: This can be done with 
```python
else:
    return connector_status[connector_name].capitalize()
```

hummingbot/connector/connector_status.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python
+
+def get_connector_status(connector_name: str) -> str:
+    """
+    Indicator whether a connector is working properly or not. Unknown means the connector is not in the dict.
+    Unavailable means a connector doesn't work, Warning means the connector works but has issues.
+    Ok means a connector is working properly. The connector_status dict is updated every new release.
+    """
+    connector_status = {
```
Nullably, line 9: This list should be outside of the function, simply as a member of this module (file).
The reason is the list will get updated often, if it is in the function, technically we'll need to retest the function and it's easier to maintain and update.

dennisocana: Thanks Jack, I just pushed a commit with this change.

hummingbot/connector/connector_status.py
dennisocana: Thanks Jack, I just pushed a commit with this change.

Nullably:
I just realised one problem with displaying connector status under `connect` command - there are many connectors under `ethereum` and each should displays status separately.

mifeng (changes requested):
Rather than Ok / Warning / Unavailable, use GREEN / YELLOW / RED in the client, so that the user doesn't have to translate colors in their head. 

> I just realised one problem with displaying connector status under connect command - there are many connectors under ethereum and each should displays status separately.

I agree with Jack that we need to think through this a bit more - I'll propose a design for next release. For now, I think it's fine since people will use this status primarily for centralized exchange connectors.

hummingbot/connector/connector_status.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python
+
+connector_status = {
+    'binance': 'ok',
+    'bitfinex': 'ok',
```
Nullably, line 5: Please add `binance_us` status.

mifeng:
@dennisocana We discussed this at Asia standup and decided to add a "message" field in addition to status. When we mark the status for each connector, we will also add a message for any YELLOW and RED connectors. 

I think we should show this message in the README and when the user starts a yellow/red connector.

For instance, OKEx would be RED with a message like: "OKEx is reportedly being investigated by Chinese authorities and has stopped withdrawals."

hummingbot/connector/connector_status.py
dennisocana: Pushed a new commit to add `binance_us`

hummingbot/connector/connector_status.py
```diff
@@ -0,0 +1,31 @@
...
+    'kucoin': 'green',
+    'liquid': 'green',
+    'loopring': 'green'
+}
+
+
+def get_connector_status(connector_name: str) -> str:
+    """
+    Indicator whether a connector is working properly or not. Unknown means the connector is not in the dict.
+    Unavailable means a connector doesn't work, Warning means the connector works but has issues.
```
Nullably, line 24: Forgot to update this?

dennisocana: Thanks Jack, I updated it just now.

Nullably (dismissed):
LGTM

mifeng:
@dennisocana Looks from the code that we show the same message for each connector that is not green.

> For instance, OKEx would be RED with a message like: "OKEx is reportedly being investigated by Chinese authorities and has stopped withdrawals."

I meant that we should have a custom status message for each connector that is YELLOW or RED that explains the status. During the review process, I think we should write/edit these messages, which are listed in the README table and the warning message when user runs `start`.

dennisocana:
@mifeng I pushed requested changes and added custom message for okex and eterbase connectors.

hummingbot/client/command/start_command.py
```diff
@@ -76,6 +78,25 @@ async def start_check(self,  # type: HummingbotApplication
         self._notify(f"\nStatus check complete. Starting '{self.strategy_name}' strategy...")
         if global_config_map.get("paper_trade_enabled").value:
             self._notify("\nPaper Trading ON: All orders are simulated, and no real orders are placed.")
+
+        # Show warning message if the exchange connector has outstanding issues or not working
+        connector = get_strategy_config_map(self.strategy_name).get("exchange").value
+        status = get_connector_status(get_strategy_config_map(self.strategy_name).get("exchange").value)
+
+        # Display custom message for connectors on RED or YELLOW status
+        if connector == "okex":
```
Nullably, line 87: A better way of doing this is to create another dictionary in `connector_status.py`, and retrieve and display messages here.
For example:
```python
warning_msg = get_connector_warning(connector)
if warning_msg is not None:
    self._notify(....
```

hummingbot/connector/connector_status.py
```diff
@@ -0,0 +1,32 @@
...
+    'eterbase': 'red',
+    'ethereum': 'red',
+    'huobi': 'green',
+    'kraken': 'yellow',
+    'kucoin': 'green',
+    'liquid': 'green',
+    'loopring': 'green',
+    'okex': 'red'
+}
+
```
Nullably, line 20: You can create another list here similar to `connector_status` but for warning messages, and create another function to retrieve it similar to `get_connector_status`, the only different is it returns None if the connector is not on the list to indicate no warning.

Nullably (changes requested):

hummingbot/client/command/start_command.py
dennisocana: Thanks @Nullably! I just pushed a new commit with these changes.

hummingbot/connector/connector_status.py
dennisocana: Thanks @Nullably! I just pushed a new commit with these changes.

Nullably (approved):

hummingbot/client/command/start_command.py
```diff
@@ -76,6 +78,21 @@ async def start_check(self,  # type: HummingbotApplication
         self._notify(f"\nStatus check complete. Starting '{self.strategy_name}' strategy...")
         if global_config_map.get("paper_trade_enabled").value:
             self._notify("\nPaper Trading ON: All orders are simulated, and no real orders are placed.")
+
+        connector = get_strategy_config_map(self.strategy_name).get("exchange").value
```
Nullably, line 82: This will only work for PMM strategy, as other strategies might not have connector config key as "exchange". A better way to get configured exchanges is to use `required_exchanges` from `setting`, this is a list so you'll have to loop through them.

Nullably (changes requested):

hummingbot/client/command/start_command.py
dennisocana: Thanks @Nullably for the catch. I pushed a new commit, tested on cross exchange and arbitrage strategy.

image

Nullably (approved):
LGTM

PR #2508, opened 2020-10-23
[v0.33] Feat / update client UI as proposed

vic-en:

hb-interface

**Note:** CPU and Mem usage may not always be thesame with report from process viewer on different platform.


rxlxrxsx:
After producing 2 trades, it still displays 0 on trades and performance.

trades

dennisocana:
Actually I just realized the feature request says "duration of the bot". In this branch, the duration in navbar is different than the duration in the `history` output.

The duration in navbar shows how long the instance is running while the duration in the `history` output shows how long the strategy is running. We can either:

1. Decide which duration we want to display (instance or strategy) and use the same for both
2. We can keep them separate but we need to explicitly indicate what each of them represent i.e. renaming to bot duration and strategy duration

vic-en:
> After producing 2 trades, it still displays 0 on trades and performance on the bottom bar.
> 
> trades

@rxlxrxsx Can you restart the bot or strategy and confirm it it still doesn't update?

vic-en:
> Actually I just realized the feature request says "duration of the bot". In this branch, the duration in navbar is different than the duration in the `history` output.
> 
> The duration in navbar shows how long the instance is running while the duration in the `history` output shows how long the strategy is running. We can either:
> 
>     1. Decide which duration we want to display (instance or strategy) and use the same for both
> 
>     2. We can keep them separate but we need to explicitly indicate what each of them represent i.e. renaming to bot duration and strategy duration

@dennisocana  I think option 2 is better. In my opinion, I think it's better for the duration at the bottom nav bar to show duration of instance. So I'll adjust that of the history command to "Strategy Duration" to make it less confusing.

dennisocana:
@vic-en @rxlxrxsx the trade count works fine on my end.

image

dennisocana (approved):
LGTM

dennisocana:
Will also test on other platforms like Windows to make sure the navbars are working correctly.

vic-en:
@dennisocana  I resolved conficts. Thanks.

dennisocana:
@vic-en we're unable to test this on Windows because of this issue https://github.com/CoinAlpha/hummingbot/issues/2518

mifeng (changes requested):
This feature is a big improvement over what we had before!

I'd like to request 2 changes:

1. Replace **Markets** and **Script_file** in the top bar with **Strategy file**. This should be more informative than Market because many users use different configs depending on market state. Also, I think Script file may not be future-proof we will allow users to load multiple scripts and we're going to steer people to write custom strategies rather than scripts.

2. In bottom bar, add **Total P&L** and change **Performance** to **Return %**. This uses the same wording as the new `history` command: 
Screen Shot 2020-10-27 at 12 58 13 PM

hummingbot/client/performance_analysis.py
```diff
@@ -201,5 +201,6 @@ def calculate_trade_performance(current_strategy_name: str,
     trade_performance_stats["portfolio_spent_quote_value"] = portfolio_spent_quote_value
     trade_performance_stats["portfolio_delta"] = portfolio_acquired_quote_value - portfolio_spent_quote_value
     trade_performance_stats["portfolio_delta_percentage"] = portfolio_delta_percentage
+    trade_performance_stats["no_of_trades"] = len(raw_queried_trades)
```
Nullably, line 204: In the #/2512 PR, I created a new module for performance calculation, it's `performance.py` and we'll not be using `performance_analysis` anymore. Though I believe you can get the same information from there.

rxlxrxsx:
Trades are not showing after successfully getting a trade from the session. I tried restarting the session but got the same result. This happened both on cloud and binary. 

Test on cloud
trades not showing 2

Test on Windows binary
trades not showing binary

Also found this on the logs after starting the strategy that's possibly related as to why trades are not logged on the ``Trades`` and ``Perfomance`` bottom navigation. 

image

```
04:57:27,426 - 32589 - hummingbot.client.ui.hummingbot_cli - ERROR - Unhandled error in prompt_toolkit: 'HummingbotApplication' object has no attribute '_calculate_trade_performance'
NoneType: None
```

[logs_binance-test.log](https://github.com/CoinAlpha/hummingbot/files/5457330/logs_binance-test.log)

vic-en:
@rxlxrxsx It is due to the merged history redesign PR. I'm yet to push new commit.

vic-en:
@rxlxrxsx I pushed all commits. Test can be done now. Thank you.

mifeng (approved):
LGTM

PR #2513, opened 2020-10-26
feat / Add inventory cost price type

bitphage:
Closes: #2314


TomasFlow (approved):

TomasFlow (approved):

TomasFlow (dismissed):

bitphage:
Notes:

1. Inventory price is being computed only when `price_type: inventory_cost` is set. I'm not sure is this fully right approach. Maybe it's better to always track inventory cost even when user is switching `price_type` to different options. When he/she will come back to `price_type: inventory_cost`, it will be accurate.
2. This implementation is using own database table to track inventory cost.

rxlxrxsx:
-`inventory_price` doesn't have a default value or is not prompted when creating a pure market making strategy thus showing as incomplete on preliminary checks after being created.

Screen Shot 2020-12-22 at 6 25 42 PM

-Randomly getting this warning upon starting the strategy although no impact on bot's behavior

```
08:32:09 - binance_api_user_stream_data_source - Reconnecting to wss://stream.binance.com:9
443/ws/gQhbiNE8Dk7W3xumgELTjl29oRrCtedFL3pfPfcidUCitTDQt4vUDxngo7S1.

/home/ubuntu/miniconda3/envs/hummingbot/lib/python3.8/site-packages/sqlalchemy/sql/sqltypes
.py:659: SAWarning: Dialect sqlite+pysqlite does *not* support Decimal objects natively, an
d SQLAlchemy must convert from floating point - rounding errors and other issues may occur.
 Please consider storing Decimal numbers as strings or integers on this platform for lossle
ss storage.
  util.warn(
  ```

-Exiting config (Ctrl + X) while configuring `inventory_price` results to this error:

```
2020-12-22 21:00:28,922 - 4186 - hummingbot.client.hummingbot_application - ERROR - unsupported operand type(s) for *: 'decimal.Decimal' and 'NoneType'
  File "/Users/reiniersalazar/hummingbot/hummingbot/client/command/config_command.py", line 147, in _config_single_key
    await self.inventory_price_prompt(config_map, input_value)
  File "/Users/reiniersalazar/hummingbot/hummingbot/client/command/config_command.py", line 263, in inventory_price_prompt
    quote_volume=balances[base_asset] * cvar.value,
TypeError: unsupported operand type(s) for *: 'decimal.Decimal' and 'NoneType'
```

**Comment by bitphage** [comment]
Addressed issues found by @rxlxrxsx, except reconnection and sqlalchemy warning.

**Comment by rxlxrxsx** [comment]
Confirmed fix for issues listed above except for the SQLAlchemy warning.

**Review by RC-13 — DISMISSED** [review]

File: hummingbot/strategy/pure_market_making/inventory_cost_price_delegate.py [hunk 1]
```diff
@@ -0,0 +1,68 @@
...
+                    # Ok, some other asset used (like BNB), assume that we paid in quote asset for simplicity
+                    quote_volume /= 1 + fill_event.trade_fee.percent
+
+        if fill_event.trade_type == TradeType.SELL:
+            base_volume = -base_volume
+            quote_volume = -quote_volume
+
+            # Make sure we're not going to create negative inventory cost here. If user didn't properly set cost,
+            # just assume 0 cost, so consequent buys will set cost correctly
+            record = InventoryCost.get_record(self._session, base_asset, quote_asset)
```
Nullably, line 58: I think it's better to store data as how the orders are filled, this is for easier logging and finding out what happened later. The business logic on how the data is used (e.g. negative volume means no price) should be applied later.

hummingbot/strategy/pure_market_making/inventory_cost_price_delegate.py
```diff
@@ -0,0 +1,68 @@
...
+        return True
+
+    def get_price(self) -> Decimal:
+        record = InventoryCost.get_record(
+            self._session, self.base_asset, self.quote_asset
+        )
+        try:
+            price = record.quote_volume / record.base_volume
+        except (ZeroDivisionError, InvalidOperation, AttributeError):
+            raise NoPrice("Inventory cost delegate does not have price cost data yet")
```
Nullably, line 25: I think it'd be easier to know which conditions exactly causing NoPrice, and it'd be better to use None to handle this rather than exception. e.g.
```python
if record.base_volume <= s_decimal_0 or record.base_volume is None:
    return None
```
InvalidOperation or AttributeError will hide other problem and make it. difficult to debug later on.

hummingbot/strategy/pure_market_making/pure_market_making_config_map.py
```diff
@@ -331,6 +341,14 @@ def exchange_on_validated(value: str):
                   prompt="Enter pricing API URL >>> ",
                   required_if=lambda: pure_market_making_config_map.get("price_source").value == "custom_api",
                   type_str="str"),
+    "inventory_cost_allow_higher_bids":
```
Nullably, line 344: Why do we need this config variable? And why only for bids?
From what I understand the bid price will always be below inventory price (if inventory price is known), the bid price is inventory price minus spread.

Nullably (changes requested):

hummingbot/model/inventory_cost.py
```diff
@@ -0,0 +1,75 @@
...
+    Integer,
+    UniqueConstraint,
+)
+from sqlalchemy.orm import Session
+
+from . import HummingbotBase
+
+
+class InventoryCost(HummingbotBase):
+    __tablename__ = "InventoryCost"
```
Nullably, line 17: Do we actually need this new table in the sqlite? 
Since we already store trade fills, we can use that to calculate inventory cost, all we need is just a start time which can be config map variable (ie. when inventory_cost price type is set, it will also set inventory_cost_start_time).

bitphage: I'm not exactly remember why I choose to use separate table. Let's imagine user has some base asset amount which wasn't obtained via trading in the bot, how it should set the cost for this amount / how we can store it properly?

hummingbot/strategy/pure_market_making/inventory_cost_price_delegate.py
bitphage: In general, a function must return only one type. It's a common anti-pattern described here:  https://docs.quantifiedcode.com/python-anti-patterns/maintainability/returning_more_than_one_variable_type_from_function_call.html

For these exceptions, all of them are appear when price data is not ready yet (for different reasons). For debug purposes, we can add info on what exact exception happened:

```python
except (ZeroDivisionError, InvalidOperation, AttributeError) as e:
    raise NoPrice(f"Inventory cost delegate does not have price cost data yet: {str(e)}")
```

hummingbot/strategy/pure_market_making/pure_market_making_config_map.py
bitphage: There are 2 reasons:
1. Over time with profit accumulation inventory price goes lower and lower, so bids can become too far from the mid price, see example here https://github.com/CoinAlpha/hummingbot/issues/2314#issuecomment-716409652
2. When market is in uptrend, you may still want to perform occasional buys.

hummingbot/strategy/pure_market_making/inventory_cost_price_delegate.py
bitphage: Hmm, removing these checks will influence logic a lot. Generally, we should not allow to sell anything without knowing it's previous cost (user didn't set existing inventory price). The current logic is a workaround for this. So, not doing such a workaround could render the whole feature not usable. It would be better to return an error to the user at the strategy start if price is not set while having some amount + "inventory cost" feature enabled. Sounds good?

hummingbot/strategy/pure_market_making/pure_market_making_config_map.py
Nullably: Personally, i don't think we should crater for exceptions like this. Reasons:
* Your reasons can be argued for `inventory_cost_allow_lower_asks` as well which we don't have, just the opposite direction.
* Should you consider to apply this only when the user is in profit? also, is it realised or unrealised profit?
*  If there is `inventory_cost_allow_lower_asks`, turning both it and `inventory_cost_allow_higher_bids` on wouldn't make any sense, since it simply neglects the feature entirely.

hummingbot/model/inventory_cost.py
Nullably: > Let's imagine user has some base asset amount which wasn't obtained via trading in the bot, how it should set the cost for this amount / how we can store it properly?

@phbrgnomo suggested below in his feature request.

> Example:
First, the bot would have to set a price for the present inventory, and we have two options for that:
1 - The user defines the inventory cost starting price

Generally, I don't think we should store data which can be calculated/derived from existing data. This will create more maintenance down the road.

hummingbot/strategy/pure_market_making/inventory_cost_price_delegate.py
Nullably: I, personally, don't like catching exceptions for expected None value, it makes code harder to read (many try-excepts and nested ones). Moreover, our practice so far has been returning None is acceptable (with hint `-> Optional[]`), it'd be easy to maintain if we follow this (but sure when there is clear benefit in accepting new coding practice we should consider).

hummingbot/model/inventory_cost.py
Nullably: @bitphage Sorry, you already store the inventory starting price when user already has the asset, can ignore my previous comment. Also, having the db table is fine for now, making changes to that logic now will be a lot of work.

bitphage: Ok. Generally, I agree that it would be better to automatically calculate inventory cost from trade fill history.

hummingbot/strategy/pure_market_making/pure_market_making_config_map.py
bitphage: Ok, I'll revert this option.

hummingbot/strategy/pure_market_making/inventory_cost_price_delegate.py
bitphage: Refactored to return None.

hummingbot/strategy/pure_market_making/inventory_cost_price_delegate.py
bitphage: @Nullably This one still needs input.

Nullably: Ok, I'm fine with that.

florinteodorescu:
Hi, I`m interested in this feature and since I saw it didn`t catch the 0.34 build I wanted to test it to see how it works so far and maybe help out by testing the changes.
Please see below my observations, I hope they are useful for you and help you fix and eventually submit this feature in a stable build:

- the script keeps getting itself in situations in which it gets an error saying "min_price_violated"
- if you set the start inventory cost as "0.0078" for example you get the above error instantly :) 
- I had to sell all of the units I had just so I can start from inventorycost 0 (I had something like 0.0093 units before this)
- when the price of the asset goes down, the script keeps buying and the total inventory cost goes down as well (this is expected and correct) but when the asset price goes up again I`ve noticed that the InventoryCost from the status screen kept going down as well after each sell order filled. The buy price kept going down as well, although the sell price stayed at a correct value.
To be more explicit: 
- I started the bot at a price of the asset of around 18.4. Bid_spread set to 0.6 and ask_spread set to 0.5
- The asset price than started to go down until 17.4 and then it went up again to around 18.9
- When the asset price went down, the bot executed more buy commands and the inventorycost went down as well
- When the price of the asset went up again, the bot started to fill sell orders and the buy price started to get lower and lower. At one point when I had 2 units of the asset left, I had the sell price at: 18.7, the buy price at around: 6.7 and the InventoryCost at around 6.8
- After all the units of the asset had been completly sold, the bot just stopped and it started to throw the "min_price_violated" warning.
- I managed to restart the bot by stopping it, manually setting the inventory_cost to 0 and then starting it again

I`m not sure what the issue is and I`m not familiar with the code to do the modifications myself, but please let me know if I can help with validating any changes or if you need more information about the reported issues above.

I`ve also attached a screenshot of the price variation during my test. Sorry I couldn`t attach a screenshot of the bot running, but its running on a difference machine :)
crv

bitphage:
@florinteodorescu I've refactored logic a bit to reflect @phbrgnomo comments on discord. Inventory price is now only limit sell orders, so buy orders are placed everytime. To limit buy prices, just use `price_ceiling`. Also sell trades now aren't influencing inventory price.

hummingbot/strategy/pure_market_making/inventory_cost_price_delegate.py
bitphage: Not sure how to properly implement start-time check, so I just added an exception for such situation.

hummingbot/model/inventory_cost.py
bitphage: I've also looked into `history` command and `calculate_performance_metrics`, not sure that using this solution to obtain average buy price is performance-perfect, as it's doing same calculations on the supplied trades every time.

rxlxrxsx:
Getting an `Unexpected error running clock tick` immediately after starting the bot. This happens when `inventory_price` parameter is not modified after changing the `price_type` to `inventory_cost`. Switching `inventory_price` to a different value or re-entering the default value (1) however doesn’t produce the same error.

```
2021-01-27 13:03:43,010 - 2593 - hummingbot.core.clock - ERROR - Unexpected error running clock tick.
  File "hummingbot/strategy/pure_market_making/pure_market_making.pyx", line 672, in hummingbot.strategy.pure_market_making.pure_market_making.PureMarketMakingStrategy.c_tick
    proposal = self.c_create_base_proposal()
  File "hummingbot/strategy/pure_market_making/pure_market_making.pyx", line 712, in hummingbot.strategy.pure_market_making.pure_market_making.PureMarketMakingStrategy.c_create_base_proposal
    raise RuntimeError("Initial inventory price is not set while inventory_cost feature is active.")
RuntimeError: Initial inventory price is not set while inventory_cost feature is active.
```

Logs: [logs_cost_based_pmm.log](https://github.com/CoinAlpha/hummingbot/files/5879989/logs_cost_based_pmm.log)
Config: [cost_based_pmm.txt](https://github.com/CoinAlpha/hummingbot/files/5879993/cost_based_pmm.txt)

bitphage:
@rxlxrxsx yes this is expected behavior because I didn't found a more proper way to validate that initial inventory is set prior to strategy start.

Nullably (approved):
LGTM

rxlxrxsx (approved):

PR #2546, opened 2020-11-02
fix / kucoin data source refactor

martinkou:

This patch fixes a subscription update bug in the last version of Kucoin order book data source - in which the old and "markets" sets being compared in the message loop had the same reference, which means they're always the same and no subscription adjustments would ever be made.

This patch also refactors the auto-ping and subscription management logic out from the main order book data source code, to make it more maintainable going forward.


Nullably (approved):
LGTM

Issue #2538, opened 2020-10-31
[BUG] XEMM/ARB Incorrect P&L and return 


RC-13:
**Describe the bug**
Similar to #2537, after getting a filled trade using cross exchange market making or arbitrage strategy. The total P&L and return doesn't matched with what's displayed on the navbar/UI.

**Steps To Reproduce**
// A concise description of the **steps to reproduce the buggy behavior**:

Here are the steps to reproduce the issue (see attachments in the section below):
1. Create a cross exchange market making strategy
2. Set a low min_profitability to speed up getting a fill
3. Once orders are filled, run `history` or `history --verbose`
3. Compare the averaged return and return % on the navbar, the result will be different
5. Same goes with the Total P&L, compare the P&L for both exchange or even get the average, it won't match the Total P&L on the UI.
...

**Screenshots**
Test done between Huobi and Binance for ZRX-ETH market pair
![xemm3](https://user-images.githubusercontent.com/57189990/97767176-63795c00-1b55-11eb-8ff5-08d8e20bc7c3.PNG)

**Release version**
0.33.0
```
branch: development
commit: 90de286843f3a8b881303f6e98c537e42ce9493e 
Date:   Fri Oct 30 17:36:59 2020 +0800
```
**Attachments**
// Please attach your config file and log file to help us debug. It would be difficult for us to help you without those! 

*WARNING: Do NOT publish any exchange API keys or your wallet's private key. Whoever has access to them may steal your assets!*

**Optional: your discord username:**

vic-en:
Similar to #2537, the values at the bottom nav bar should be treated as estimate.

This issue can also be closed.

Nullably:
Also please see my comment on #2537 before closing this.

RC-13:
Re-test and confirm return% base on the last trade.  Average return and return under the navbar match, can confirm that issue can now be closed.
![history xemm](https://user-images.githubusercontent.com/57189990/98491999-83a1cd00-2271-11eb-8656-c2bb78d5bc74.PNG)
![history xemm1](https://user-images.githubusercontent.com/57189990/98492002-86042700-2271-11eb-9028-71d570621721.PNG)

dennisocana:
Reopening issue. Actually we should explicitly indicate first in our docs why there will be a difference on the performance in `history` output and the navbar when the price changes. This will definitely be a common question from users. Will close when the docs are ready.

Issue #2537, opened 2020-10-31
[BUG] P&L and Return mismatched between history and navbar


RC-13:
**Describe the bug**
After getting a filled trade, when you run `history` or `history --verbose` and compare the `Total PnL` output and  and `Return %` with what's displayed on the navbar, most of the time they mismatched or output are delayed on the UI.

**Steps To Reproduce**
// A concise description of the **steps to reproduce the buggy behavior**:

Here are the steps to reproduce the issue (see attachments in the section below):
1. Create a pure market making strategy on any market pair
2. Set a tight spread to speed up getting a fill
3. Once you have filled order, run `history` or `history --verbose` then compare the result with what's displayed on the navbar/UI
...

**Screenshots**
Binance test bot:
![pnl](https://user-images.githubusercontent.com/57189990/97766666-6d4d9000-1b52-11eb-9105-7da53582c5f6.png)

Kucoin test bot:
![image](https://user-images.githubusercontent.com/57189990/97766851-578c9a80-1b53-11eb-9709-3e50eb17c001.png)

**Release version**
0.33.0
```
branch: development
commit: 90de286843f3a8b881303f6e98c537e42ce9493e 
Date:   Fri Oct 30 17:36:59 2020 +0800
```

**Attachments**
// Please attach your config file and log file to help us debug. It would be difficult for us to help you without those! 

*WARNING: Do NOT publish any exchange API keys or your wallet's private key. Whoever has access to them may steal your assets!*

**Optional: your discord username:**

vic-en:
The values from both history and nav bar may not always be thesame.
The values from the bottom nav bar should be treated as estimate due to design.

This issue can be closed.

Nullably:
The reason Return % are not the same is the current market price used for Return % calculation moves during the time navbar reports the return and the time `history` command is run. 
To verify if this is the case, please test if Return % values are different during the period where last traded price doesn't move.

RC-13:
Re-test and confirmed that return% on the navbar matches with the calculated return on history after the last trade. If the current price move, return on history will change but navbar will retain the calculated return base on the last price.

Base on the last trade:
![last trade](https://user-images.githubusercontent.com/57189990/98421341-1bfa4f00-20c4-11eb-9a8e-4efbd6f367f2.PNG)

After a change in the current price:
![move price](https://user-images.githubusercontent.com/57189990/98421656-ca9e8f80-20c4-11eb-81c7-663da15ed0c0.PNG)

```
[...]

Issue #2535, opened 2020-10-30
[BUG] Cosmetic bug with trade volume quantization on redesigned history output


RC-13:
**Describe the bug**

Currently trade volume precision is set to 6 decimal, some token with more than 6 precision may tend to have slight different result once performance is computed, for example, QASH-ETH market on liquid, Trade P&L and return are slightly off when calculated manually.
![history-output](https://user-images.githubusercontent.com/57189990/97651441-00c18b00-1a97-11eb-80ed-7f5c0047abfa.PNG)

**Steps To Reproduce**
// A concise description of the **steps to reproduce the buggy behavior**:

Here are the steps to reproduce the issue (see attachments in the section below):
1. Create a pure market making strategy and use a market pair with more than 6 precision
2. Set a thigh spread to easily get a fill
3. Once you have a filled order, run history command and try to manually compute the results of performance
using the formula from the code base
```
    perf.hold_value = (perf.start_base_bal * perf.cur_price) + perf.start_quote_bal
    perf.cur_value = (perf.cur_base_bal * perf.cur_price) + perf.cur_quote_bal
    perf.trade_pnl = perf.cur_value - perf.hold_value
    fee_paid = 0
    perf.fee_token = quote
    if trades[0].trade_fee.get("percent", None) is not None and trades[0].trade_fee["percent"] > 0:
        fee_paid = sum(t.price * t.amount * t.trade_fee["percent"] for t in trades)
    elif trades[0].trade_fee.get("flat_fees", []):
        perf.fee_token = trades[0].trade_fee["flat_fees"][0]["asset"]
        fee_paid = sum(f["amount"] for t in trades for f in t.trade_fee.get("flat_fees", []))
    perf.fee_paid = Decimal(str(fee_paid))
    perf.total_pnl = perf.trade_pnl - perf.fee_paid if perf.fee_token == quote else perf.trade_pnl
    perf.return_pct = divide(perf.total_pnl, perf.hold_value)
    return perf
```
...

**Screenshots**
// If applicable, add screenshots to help explain your problem.

**Release version**
0.33.0
```
branch: development
commit: b4b11fa913f080a2ae58885e62abd03c27aaa294
Date:   Thu Oct 29 10:24:37 2020 -0700
```

**Attachments**
// Please attach your config file and log file to help us debug. It would be difficult for us to help you without those! 

*WARNING: Do NOT publish any exchange API keys or your wallet's private key. Whoever has access to them may steal your assets!*

[...]

PR #2560, opened 2020-11-04
[v0.33] Fix / fix all cosmetic bug in history command by adding the precision option

vic-en:

### DIFF
--- a/hummingbot/client/performance.py
+++ b/hummingbot/client/performance.py
@@ -111,6 +111,19 @@ def calculate_performance_metrics(trading_pair: str,
     return perf
 
 
-def smart_round(value: Decimal) -> Decimal:
-    # Simply quantize values to 8 decimal places
-    return Decimal(str(value)).quantize(Decimal("0.00000001"))
+def smart_round(value: Decimal, precision: int = 0) -> Decimal:
+    if precision > 0:
+        precision = 1 / (10 ** precision)
+        return Decimal(str(value)).quantize(Decimal(str(precision)))
+    step = Decimal("1")
+    if Decimal("10000") > abs(value) > Decimal("100"):
+        step = Decimal("0.1")
+    elif Decimal("100") > abs(value) > Decimal("1"):
+        step = Decimal("0.01")
+    elif Decimal("1") > abs(value) > Decimal("0.01"):
+        step = Decimal("0.0001")
+    elif Decimal("0.01") > abs(value) > Decimal("0.0001"):
+        step = Decimal("0.00001")
+    elif Decimal("0.0001") > abs(value) > s_decimal_0:
+        step = Decimal("0.000001")
+    return (value // step) * step
--- a/hummingbot/client/performance.py
+++ b/hummingbot/client/performance.py
@@ -112,15 +112,5 @@ def calculate_performance_metrics(trading_pair: str,
 
 
 def smart_round(value: Decimal) -> Decimal:
-    step = Decimal("1")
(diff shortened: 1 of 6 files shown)


RC-13:
Test and confirm that PR address issue #2535 and #2537

RC-13:
Re-test issue #2538, confirm that return (navbar) matches now the average return under history, however,  it won't reflect the change if there will be a price movement. Also the total P&L doesn't still match with the average P&L under history status.

I tried manually getting the P&L average base on the P&L for each exchange then convert it to USDT and it did not match what's displayed on the navbar.
binance performance
huobi performance

hummingbot/client/performance.py
```diff
@@ -112,15 +112,5 @@ def divide(value, divisor):
...
-        step = Decimal("0.01")
-    elif Decimal("1") > abs(value) > Decimal("0.01"):
-        step = Decimal("0.0001")
-    elif Decimal("0.01") > abs(value) > Decimal("0.0001"):
-        step = Decimal("0.00001")
-    elif Decimal("0.0001") > abs(value) > s_decimal_0:
-        step = Decimal("0.000001")
-    return (value // step) * step
+    # Simply quantize values to 8 decimal places
+    return Decimal(str(value)).quantize(Decimal("0.00000001"))
```
Nullably, line 116: if we do this, it will no longer be a smart round, also what happens when this fn is used on large value e.g. BTC-USDT of 14,000?

vic-en:
@RC-13 
Kindly note that the values shown at the bottom bar wouldn't always match the one in history for several reason.
One should just take those values as estimate and not always expect them to be thesame.
Also, one reason why your manual calculation wouldn't always give exact value as the one shown is that the one shown gets the conversion rate to USDT and does that conversion **only when the number of trade increases**. So you manual result may always be slight different.

hummingbot/client/ui/interface_utils.py
```diff
@@ -63,16 +62,32 @@ async def start_trade_monitor(trade_monitor):
...
                         cur_balances = await hb.get_current_balances(market)
                         cur_price = await get_last_price(market.replace("_PaperTrade", ""), symbol)
                         perf = calculate_performance_metrics(symbol, cur_trades, cur_balances, cur_price)
                         return_pcts.append(perf.return_pct)
-                        pnls.append(perf.total_pnl)
+                        # convert to USDT equivalent
+                        usdt_price = None
+                        if quote_asset not in ["USDT", "USDC", "USD"]:
+                            try:
+                                usdt_price = Decimal(str(perf.total_pnl)) * await get_last_price(market.replace("_PaperTrade", ""), f"{quote_asset}-USDT")
```
Nullably, line 75: I don't think this is a good approach as USDT trading pair may not always be available, especially for connectors like Balancer. It's probably safer to assume the quote assets of different trading pairs are of the same value (it doesn't make sense in the first place to do strategy like Arb or XEMM on different quote values pairs).

Nullably (changes requested):

hummingbot/client/performance.py
vic-en: Yeah, I agree it may no longer be a smart function. If that's the case I can do a rename to `report_round`, just felt that wasn't necessary. A value of 14,000 would be 14000.00000000.

hummingbot/client/ui/interface_utils.py
vic-en: okay, I agree. I will revert the change.

Nullably (approved):
LGTM

Issue #2672, opened 2020-11-24
Dynamically calculate gas price for Balancer trades


mifeng:
# Why

Setting the correct gas limit used for Balancer trades is very important, since transaction costs can be $20 or more at higher gas prices. The initial implementation of Balancer assumes a 1.2m gasLimit in `gateway-api`, which was too conservative.

In this draft PR for Balancer, I have changed how gas is calculated for Balancer transactions. In the `buy-price` and `sell-price` transactions, the user can pass in a `maxSwaps` param that affects the gasLimit:
```
const MAX_SWAPS = 4;
const GAS_BASE = 200688;
const GAS_PER_SWAP = 100000;
...
gasLimit: GAS_BASE + swaps.length * GAS_PER_SWAP
```

Since Balancer transactions are expensive, setting this gasLimit correctly ensures that the user doesn't pay too much gas.

# What

In the client, use the logic above to set the gasLimit used for Balancer transactions:

* Add a global parameter called `balancer_max_swaps`, default = 4 (int)
* Pass this parameter as `maxSwaps` into calls to `buy-price` and `sell-price` so that it's used to set the gasLimit
* In the arbitrage profitability calculation, use the logic above to estimate `gasLimit`

mifeng:
@sdgoh Here's the Gateway PR for this change: https://github.com/CoinAlpha/gateway-api/pull/15

mifeng:
@sdgoh Are you working on this ticket? I haven't seen any updates on this over past few days

PR #2699, opened 2020-12-08
Feat/dynamic balancer gas price

sdgoh:

**Description**:

- Add global config for balancer max swap setting. 
- Get quote/order price api call with max swap param per Balancer transaction
- Get estimate gas limit based-on max swap only when global config balancer_max_swaps changes.

Issue #2672


hummingbot/core/utils/eth_gas_station_lookup.py
```diff
@@ -28,6 +30,50 @@ def get_gas_price(in_gwei: bool = True) -> Decimal:
     return gas_price if in_gwei else gas_price / Decimal("1e9")
 
 
+def get_gas_limit(connector_name: str) -> int:
+    gas_limit = None
+    Lookup = EthGasStationLookup.get_instance()
```
Nullably, line 35: object instance name should start with lower case.

hummingbot/core/utils/eth_gas_station_lookup.py
```diff
@@ -72,6 +120,23 @@ def refresh_time(self):
     def gas_price(self):
         return self._gas_prices[self.gas_level]
 
+    @property
+    def gas_limit(self):
```
Nullably, line 124: if this is a dictionary, might be better to name it with plural, e.g. gas_limits

Nullably (approved):
I left a few comments on naming convention. The rest looks good to me.

sdgoh:
naming convention updated in a different branch.

Issue #2670, opened 2020-11-23
Poll for confirmation from client


mifeng:
# Why

Currently for Ethereum transactions executed by `gateway-api`, the `buy` and `sell` functions return failure or success after the transaction is confirmed on the blockchain. This may take a long time if the chain is congested.

Ethereum blockchain transactions can be split into 2 parts:

* sending the transaction to the chain and getting a `txHash`
* polling the chain with `txHash` and awaiting success/failure of the transaction

By splitting ETH transactions up into these 2 steps, we can get better resolution on transaction status.

# What

- Gateway (Uniswap + Balancer): Split the buy and sell transactions into 2 steps
- Client (Uniswap + Balancer): Afterwards, handle each step separately: Emit log message when `txHash` is available, and poll for transaction confirmation - I believe Ryan may have added logic to the Radar Relay connector for this already.

PR #2703, opened 2020-12-09
Feat/tx polling for confirmation

sdgoh:

#2670

- Update Balancer & Uniswap connectors to add tx entry to in-flight order based-on return tx hash from Gateway. 
- Tx polling will access Gateway endpoint to get up-to-date tx confirmation & mark as complete or fail accordingly.


hummingbot/core/utils/eth_gas_station_lookup.py
```diff
@@ -28,6 +30,50 @@ def get_gas_price(in_gwei: bool = True) -> Decimal:
...
+            global_config_max_swaps = global_config_map["balancer_max_swaps"].value
+            if current_max_swaps == global_config_max_swaps:
+                gas_limit = lookup.gas_limits[connector_name]
+            else:
+                gas_limit = request_gas_limit(connector_name)
+                # update new max swap & gas limit due to config change
+                lookup.balancer_max_swaps = global_config_max_swaps
+                lookup.gas_limits = {connector_name: gas_limit}
+        elif connector_name == "uniswap":
+            gas_limit = lookup.gas_limits[connector_name]
```
vic-en, line 50: I think it's okay if we want to handle balancer specially for now. If that's the case, we may want to try something like:
```suggestion
        else:
            gas_limit = lookup.gas_limits[connector_name]
```
That way, it would be able to handle other ethereum connectors we would add to gateway later.
Also, we would certainly refactor this part of the code and `EthGasStationLookup` class later to be non-balancer specific. For instance, there's a possibility we would enhance Uniswap to go through a number of hops when it's routing swaps soon.Also, we can assume other connector have a max_swaps of 2.

vic-en (changes requested):

hummingbot/core/utils/eth_gas_station_lookup.py
mifeng: @sdgoh Let's talk through the changes to `eth_gas_station_lookup.py` after Asia standup. I think the current logic is close but could use a few tweaks.

hummingbot/client/config/config_helpers.py
```diff
@@ -186,7 +186,7 @@ def get_erc20_token_addresses() -> Dict[str, List]:
                 override_list.append(token)
 
             if len(override_list) > 0:
-                print(f"Applied address overrides from /conf/erc20_tokens.json to symbols {override_list}.")
+                print(f"Applied address overrides from {TOKEN_ADDRESSES_FILE_PATH} to symbols {override_list}.")
```
Nullably, line 189: It'd be better to log this, st. we have the message printed on the screen as well as logged into a file (for later analysis if needed).

hummingbot/client/config/global_config_map.py
```diff
@@ -121,6 +121,13 @@ def connector_keys():
                   required_if=lambda: global_config_map["celo_address"].value is not None,
                   is_secure=True,
                   is_connect_key=True),
+    "balancer_max_swaps":
+        ConfigVar(key="balancer_max_swaps",
+                  prompt="Enter the maximum swap pool in Balancer >>> ",
+                  required_if=lambda: False,
+                  type_str="int",
+                  validator=lambda v: validate_decimal(v, Decimal(-100), Decimal(100)),
```
Nullably, line 129: It'd be better to validate_int for this.

hummingbot/connector/connector/balancer/balancer_connector.py
```diff
@@ -48,6 +48,7 @@ class BalancerConnector(ConnectorBase):
     """
     API_CALL_TIMEOUT = 10.0
     POLL_INTERVAL = 60.0
+    UPDATE_ORDER_STATUS_MIN_INTERVAL = 10.0
```
Nullably, line 51: I don't think we need this min interval, it's there in other connectors which have web socket updates, so the _update_order_status is used as a backup. So for this connector, the _update_order_status should be called every poll_interval.
Also the poll_interval of 60 seconds looks a bit too long to me.

hummingbot/connector/connector/balancer/balancer_connector.py
```diff
@@ -345,6 +319,79 @@ def stop_tracking_order(self, order_id: str):
...
+                if isinstance(update_result, Exception):
+                    raise update_result
+                if "txHash" not in update_result:
+                    self.logger().info(f"_update_order_status txHash not in resp: {update_result}")
+                    continue
+                if update_result["confirmed"] is True:
+                    if update_result["receipt"]["status"] == 1:
+                        gas_used = update_result["receipt"]["gasUsed"]
+                        gas_price = get_gas_price()
+                        fee = Decimal(str(gas_used)) * Decimal(str(gas_price)) / Decimal(str(1e9))
```
Nullably, line 350: Shouldn't the gas_used be the total amount of gas used for the transaction, ie. gas_used is the fee?

hummingbot/connector/connector/balancer/balancer_connector.py
```diff
@@ -492,7 +540,7 @@ async def _api_request(self,
             err_msg = ""
             if "error" in parsed_response:
                 err_msg = f" Message: {parsed_response['error']}"
-            raise IOError(f"Error fetching data from {url}. HTTP status is {response.status}.{err_msg}")
+                raise IOError(f"Error fetching data from {url}. HTTP status is {response.status}.{err_msg}")
```
Nullably, line 543: It'd be safer to always raise exception when the response is not 200 regardless.

hummingbot/connector/connector/uniswap/uniswap_connector.py
```diff
@@ -492,7 +537,7 @@ async def _api_request(self,
             err_msg = ""
             if "error" in parsed_response:
                 err_msg = f" Message: {parsed_response['error']}"
-            raise IOError(f"Error fetching data from {url}. HTTP status is {response.status}.{err_msg}")
+                raise IOError(f"Error fetching data from {url}. HTTP status is {response.status}.{err_msg}")
```
Nullably, line 540: It'd be safer to always raise exception when the response is not 200 regardless.

hummingbot/connector/connector/uniswap/uniswap_connector.py
```diff
@@ -48,6 +48,7 @@ class UniswapConnector(ConnectorBase):
     """
     API_CALL_TIMEOUT = 10.0
     POLL_INTERVAL = 60.0
+    UPDATE_ORDER_STATUS_MIN_INTERVAL = 10.0
```
Nullably, line 51: same, we don't need this.

hummingbot/connector/connector/uniswap/uniswap_connector.py
```diff
@@ -345,6 +316,79 @@ def stop_tracking_order(self, order_id: str):
...
             del self._in_flight_orders[order_id]
 
+    async def _update_order_status(self):
+        """
+        Calls REST API to get status update for each in-flight order.
+        """
+        last_tick = self._last_poll_timestamp / self.UPDATE_ORDER_STATUS_MIN_INTERVAL
+        current_tick = self.current_timestamp / self.UPDATE_ORDER_STATUS_MIN_INTERVAL
+
+        if current_tick > last_tick and len(self._in_flight_orders) > 0:
```
Nullably, line 326: no need to check for the time here.

hummingbot/strategy/amm_arb/amm_arb_config_map.py
```diff
@@ -88,7 +88,7 @@ def order_amount_prompt() -> str:
         prompt="How much buffer do you want to add to the price to account for slippage for orders on the first market "
                "(Enter 1 for 1%)? >>> ",
         prompt_on_new=True,
-        default=Decimal("0"),
+        default=Decimal("0.05"),
```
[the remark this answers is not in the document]
Nullably: There are a few reasons for this:
- Hummingbot always round the price down, for buy order this causes the order to be below the quote price and the transaction might get rejected.
- Ethereum transactions are costly, it'd be safer to always provide some margin for price.

Nullably (changes requested):

hummingbot/strategy/amm_arb/amm_arb_config_map.py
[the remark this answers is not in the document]
sdgoh: The quantize function always rounds down and the order price is less in few decimal points in value than the actual quote price. This causes the 'Price too high' on Balancer trade submission. So we set the default to the minimum of 0.05. This also gets applied to Uniswap since the config map is not connector specific.

hummingbot/connector/connector/uniswap/uniswap_connector.py
sdgoh: Fixed the Gateway return code and capture any non 200 error.

hummingbot/connector/connector/balancer/balancer_connector.py
sdgoh: I separated the tx poll order update from the main _status_polling_loop() into _tx_polling_loop(). This let us keep 2 time intervals (1 for balance, another for faster tx polling).

hummingbot/connector/connector/balancer/balancer_connector.py
sdgoh: gas_used is the actual amount used which should be lower than our gas limit estimation (i.e. number of swaps + base gas). The `get_gas_price()` returns the estimated (fast, fastest, etc) gas price. The fee (Transaction Fee) is the final cost used to complete the transaction.

hummingbot/connector/connector/balancer/balancer_connector.py
sdgoh: I refactored this further to use the single loop.

hummingbot/connector/connector/uniswap/uniswap_connector.py
sdgoh: Changed to set balance polling interval.

hummingbot/connector/connector/balancer/balancer_connector.py
```diff
@@ -257,53 +259,24 @@ async def _create_order(self,
...
-            status = order_result["status"]
             tracked_order = self._in_flight_orders.get(order_id)
             if tracked_order is not None:
                 self.logger().info(f"Created {trade_type.name} order {order_id} txHash: {hash} "
                                    f"for {amount} {trading_pair}.")
-                tracked_order.exchange_order_id = hash
-            if int(status) == 1:
+                tracked_order.update_exchange_order_id(hash)
+                tracked_order.gas_price = gas_price
+            if hash:
```
Nullably, line 272: better to be more specific here, doesn't look like hash is `bool` type. 
does this mean `hash is not None`?

hummingbot/connector/connector/uniswap/uniswap_connector.py
```diff
@@ -254,53 +256,24 @@ async def _create_order(self,
...
-            status = order_result["status"]
             tracked_order = self._in_flight_orders.get(order_id)
             if tracked_order is not None:
                 self.logger().info(f"Created {trade_type.name} order {order_id} txHash: {hash} "
                                    f"for {amount} {trading_pair}.")
-                tracked_order.exchange_order_id = hash
-            if int(status) == 1:
+                tracked_order.update_exchange_order_id(hash)
+                tracked_order.gas_price = gas_price
+            if hash:
```
Nullably, line 269: same here, is it `hash is not None`?

Nullably (dismissed):
A few minor clarification, the rest looks good to me.

Nullably (approved):
LGTM

PR #2717, opened 2020-12-14
(fix) Terra order not submitted. Gateway install script update

sdgoh:

- Fix Terra connector where it was calling Balancer InFlighOrder class. 
- Update Gateway install script to read global config from HB yaml file. 
- Gateway install script prompts user for Balancer/Uniswap or Terra and exit if 'No' to both.


sdgoh:
The TerraInFlightOrder was using same logic for BalancerInFlightOrder. It was working for Terra until BalancerInFlightOrder was updated to include new property in the class, therefore breaking Terra's connector.

dennisocana (approved):
LGTM

PR #2867, opened 2021-01-24
feat/Add relative-to-start profitability and equity

ladismrkolj:
In an effort to better asses the market-making profitability of my pmm bot I added relative-to-start profitability (as in the hummingbot blog: [here](https://hummingbot.io/blog/2019-07-measure-performance-crypto-trading/), the fixed method) and equity (to status screen).

Equity represents the value that would be returned to a company's shareholders if all of the assets were liquidated and all of the company's debts were paid off. We can also think of equity as a degree of residual ownership in a firm or asset after subtracting all debts associated with that asset. (Wikipedia)

Basically equity is the balance you would have if all the orders were instantly closed, this solves the hanging orders dragging down your profitability, when they could close soon/or never (so take some precaution).

In global config I added a variable `advanced_stats`, which enables the display of the mentioned statistics.

_Minor issues that I didn't know how to solve but don't affect the functionality of the bot:_

> There is only one problem I didn't know how to tackle. The advanced_stats variable isn't persistant. Every time I restart the hummingbot it has empty value and you have to set it again.
> 
> Also: I also found increased RAM usage in my docker image, but haven't found the reason why or if it even is a problem.

The build passes on docker hub and I am sucessfully running it.

image
image

### DIFF
--- a/hummingbot/client/command/history_command.py
+++ b/hummingbot/client/command/history_command.py
@@ -164,8 +164,7 @@ class HistoryCommand:
         perf_data = [
             ["Hold portfolio value    ", f"{smart_round(perf.hold_value, precision)} {quote}"],
             ["Current portfolio value ", f"{smart_round(perf.cur_value, precision)} {quote}"],
-            ["Trade P&L               ", f"{smart_round(perf.trade_pnl, precision)} {quote}"],
-            
+            ["Trade P&L               ", f"{smart_round(perf.trade_pnl, precision)} {quote}"]
         ]
         if global_config_map.get("advanced_stats").value:
             perf_data.extend(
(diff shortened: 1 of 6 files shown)


hummingbot/client/command/history_command.py
```diff
@@ -150,9 +150,9 @@ def report_performance_by_market(self,  # type: HummingbotApplication
              smart_round(perf.cur_quote_bal, precision),
              smart_round(perf.tot_vol_quote, precision)],
             [f"{trading_pair + ' price':<17}",
-             smart_round(perf.start_price),
```
ladismrkolj, line 153: this is optional

dennisocana:
@phbrgnomo can you please review or test this PR since it's more of a product change. Because it's a request to merge an additional feature, we should try adding this to our plans for v0.38 release

dennisocana:
Hi @ladismrkolj thanks for your contribution. We're planning to include this PR in the upcoming release this May. Please resolve the conflicting files and update your branch with the most recent changes in `development`.

dennisocana:
Pull request has gone stale and outdated, no response from the contributor.

Issue #2760, opened 2020-12-20
Bug on Binance Perpetual Market Making Take Profit Mechanism


billcrypto97:
**Describe the bug**
Hummingbot places extra but unnecessary Take Profit order when multiple levels of orders have been filled. As can be seen in Screenshots: Examples of what happened. My bot had 3 buy order filled, for a total amount of 0.3 LTC and average price of 117.94. Therefore, according to my config, the bot should place Take Profit order amounting to 0.3 LTC at 0.15% profit (1.0015*117.94=118.12). As can be seen in my screenshot, the bot does have this take profit sell order of 0.3 LTC at 118.12, but the bot also has 2 other sell orders which are unnecessary to be placed there. From what I understand, the bot should only place the Take Profit Sell Order of 0.3 LTC at 118.12, and once it gets filled, the bot will return to No Position, and can start with a new order cycle.

**Steps To Reproduce**

Here are the steps to reproduce the issue (see attachments in the section below):
1. Set up your configuration file with these parameters:
Strategy: Perpetual Market Making
Maker Exchange: Binance Perpetual
Order Levels: 3
Position Mode: One-way 
Position Management: Profit Taking

**Screenshots**
Config
![image](https://user-images.githubusercontent.com/76391386/102711812-f1202f00-42ee-11eb-9f8a-a1d6206114ef.png)

Examples of what happened:
![image](https://user-images.githubusercontent.com/76391386/102711863-42c8b980-42ef-11eb-98ed-5219bf68c048.png)

**Release version**
v0.34.0

**Attachments**
[logs_LTC-USDT_PERP_Test.log](https://github.com/CoinAlpha/hummingbot/files/5720828/logs_LTC-USDT_PERP_Test.log)
[LTC-USDT_PERP_Test.txt](https://github.com/CoinAlpha/hummingbot/files/5720829/LTC-USDT_PERP_Test.txt)

**Updated Log**
Contains errors where during a sudden move in the market the unnecessary take profit orders are filled and might disrupt the bot flows. Extremely dangerous. Look at the latest trades.

[logs_LTC-USDT_PERP.log](https://github.com/CoinAlpha/hummingbot/files/5724548/logs_LTC-USDT_PERP.log)

*WARNING: Do NOT publish any exchange API keys or your wallet's private key. Whoever has access to them may steal your assets!*

**Optional: your discord username:billcrypto97**

PR #2880, opened 2021-01-27
Feat/add stop_loss to perpetual mm strategy

vic-en:

**With profit taking feature:**
- A profit taking order would be placed immediately if the position has no opening loss, otherwise it'd monitor position and place a stop-loss order if loss increases to the set stop-loss spread.
- Existing take profit order will be cancelled and stop-loss order will be placed if loss increases to the set stop-loss spread.
-  If stop-loss spread is set too low and LIMIT order is specified to close position in an active market, there's a high probability that the take profit feature would cancel and replace a stop-loss order. Both features might even compete every tick and cancel order previously set by the other.

**With trailing stop feature:**
- A log has been added to this feature to include the price at which the feature would close an active position. Price "NILL" means the market is yet to move enough to have a peak price that'll give a profitable exit price according to the configured parameters for the feature.
- Once a profitable exit price is detected, no stop-loss order should be placed(except for weird cases where the market swings swiftly to an unprofitable price within ticks or configured ts parameters are too low.)
- In the event that no exit price has been detected, an order from the stop-loss feature may be placed if loss increases to the set stop-loss spread.

### DIFF
--- a/hummingbot/strategy/perpetual_market_making/perpetual_market_making.pxd
+++ b/hummingbot/strategy/perpetual_market_making/perpetual_market_making.pxd
@@ -21,6 +21,7 @@ cdef class PerpetualMarketMakingStrategy(StrategyBase):
         object _stop_loss_spread
         object _close_position_order_type
         object _close_order_type
+        double _market_position_close_timestamp
         int _order_levels
         int _buy_levels
         int _sell_levels
@@ -52,7 +53,6 @@ cdef class PerpetualMarketMakingStrategy(StrategyBase):
         int _filled_buys_balance
         int _filled_sells_balance
         list _hanging_order_ids
-        list _exit_order_ids
         double _last_timestamp
         double _status_report_interval
         int64_t _logging_options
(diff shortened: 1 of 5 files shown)


RC-13:
Confirmed and tested by QA, also reported issue before (#2760) did not persist during testing.

Test summary
  - setup a pure market making strategy with the following configuration:
   > bid and ask spreads: 0.01 
   > profit taking spread set to 0.1 
   > stop-loss spread set to 0.2
  - when entry price is opened and no loss is detected, profit taking order is created 
  - when loss is detected, stop loss order will be created instead
  - set order level to 3 with order level spread of 0.001
   >  - when one side got filled at the same time, the entry price will be the average of the filled side
   >  - when order gets filled, one at a time, same behavior from above, bot recalculate entry price and order amount is adjusted
  - change position management to trailing stop with ts_activation_spread 
   >  - price NILL shown on the logs if bot did not detect profitable exit price : Ok
   >  - stop-loss order created when exit price is NILL : Ok
  - when bot detects profitability, trailing start and logs will display possible peak price and exit price
   > successfully close position base on the last peak and ts call back rate

RC-13 (approved):

keithbaum (approved):
LGTM

PR #2895, opened 2021-01-30
fix/ [script-base] Set maximum length of mid_prices list

nickfraser:

If a script is run indefinitely, the mid_prices list will continue to accumulate entries, eventually resulting in an OOM crash, as noted by [this issue](https://github.com/CoinAlpha/hummingbot/issues/2872). This PR sets a maximum value of the mid_prices script, with the length set to a time window of 1 day.

The limit is arbitrary in this case, feel free to suggest how long the time window should be and I can adjust accordingly.


Nullably (approved):
LGTM

PR #2910, opened 2021-02-02
fix/Fixed bug in trade history retrieval for history reconciliation

keithbaum:

Changed way order is searched in in_flight_orders when history reconciliating. Also fixed bug when retrieving trade history

### DIFF
--- a/hummingbot/connector/exchange/binance/binance_exchange.pyx
+++ b/hummingbot/connector/exchange/binance/binance_exchange.pyx
@@ -447,8 +447,8 @@ cdef class BinanceExchange(ExchangeBase):
             tasks=[]
             for trading_pair in self._order_book_tracker._trading_pairs:
                 my_trades_query_args = {'symbol': convert_to_exchange_trading_pair(trading_pair)}
-            tasks.append(self.query_api(self._binance_client.get_my_trades,
-                                        **my_trades_query_args))
+                tasks.append(self.query_api(self._binance_client.get_my_trades,
+                                            **my_trades_query_args))
             self.logger().debug("Polling for order fills of %d trading pairs.", len(tasks))
             exchange_history = await safe_gather(*tasks, return_exceptions=True)
         for trades, trading_pair in zip(exchange_history, self._order_book_tracker._trading_pairs):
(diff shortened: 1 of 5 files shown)


hummingbot/connector/exchange/binance/binance_exchange.pyx
```diff
@@ -460,14 +460,13 @@ cdef class BinanceExchange(ExchangeBase):
...
-                    client_order_id = get_client_order_id("buy" if trade["isBuyer"] else "sell", trading_pair)
                     # Should check if this is a partial filling of a in_flight order.
                     # In that case, user_stream or _update_order_fills_from_trades will take care when fully filled.
-                    if client_order_id not in self.in_flight_orders:
+                    if not any(trade["id"] in in_flight_order.trade_id_set for in_flight_order in self._in_flight_orders.values()):
                         self.c_trigger_event(self.MARKET_ORDER_FILLED_EVENT_TAG,
                                              OrderFilledEvent(
                                                  trade["time"],
-                                                 client_order_id,
+                                                 get_client_order_id("buy" if trade["isBuyer"] else "sell", trading_pair),
```
vic-en, line 469: I don't think we should generate a new client id at this point. We use the client id to identify orders that were created by the bot. Order created manually shouldn't need a client id attached to them. Also, orders created by the bot should have same client id that was generated when the order was created by the bot. 
Lastly, I think we should use https://binance-docs.github.io/apidocs/spot/en/#all-orders-user_data endpoint for history reconciliation as it include bot client and exchange order id.
Then

vic-en (changes requested):

hummingbot/connector/exchange/binance/binance_exchange.pyx
keithbaum: I've tried using that endpoint instead of the currently being used (myTrades) but unfortunately output info lacks the commission paid in the trade. That's important to create the order filled event. I couldn't find an way to get the commission other than calling other endpoints too.
image

On the other hand, think it's a good idea and will implement your suggestion to reuse the original local order id instead of generating a new one. This requires further changes which will be proposed in this PR.

vic-en (approved):
LGTM

PR #2919, opened 2021-02-04
Fix / race condition in binance status polling loop

vic-en:

This would prevent a race condition that results in a bug where some where an completion event is triggered before necessary filled events are triggered.


CrimsonJacket (approved):
LGTM 🍵

PR #2924, opened 2021-02-04
(fix) history reconciliation trading_pairs list

keithbaum:
hist recon now performs its own request to get_my_trades instead of reusing result from _update_order_fills_from_trades.
In that way it doesn't rely on the list of in_flight_orders being unchanged between latest call and the history recon.


CrimsonJacket (approved):
LGTM

Nullably:
This branch has been included in #2871 so I am closing this PR.

PR #3031, opened 2021-03-08
fix/ spot perp finance history error

Nullably:

- check for xdai balance when connector is perpetual finance, it's not a clean solution. Perp Fi is Ethereum connector but balances are queried on layer 2.


RC-13 (approved):
Confirmed PR addresses the reported issue, there is no longer an error when running the `history` command if the strategy is stopped or restarted.

vic-en (approved):
LGTM

PR #3059, opened 2021-03-10
[0.37.1] hotfix / added check to replace timeout errors with empty lists

keithbaum:
Added check to replace timeout errors with empty lists when fetching trading pairs in slow connections


dennisocana:
Changing branch to `master` since we're pushing this as a hotfix for the issue when entering a trading pair in any connectors. Checking in with @CoinAlpha/qa-team from time to time about testing and target 0.37.1 release is before the end of today.

hummingbot/VERSION
```diff
@@ -1 +1 @@
-0.37.0
+dev-0.38.0
```
CrimsonJacket, line 1: Considering that this is a hotfix, should we change this to `0.37.1` instead.

CrimsonJacket (approved):

RC-13 (approved):
Tested and confirmed by QA,
Here's the test case that we did (all tests are done in different environments):
1. Disconnect from the internet and start the bot
   - Create a strategy and re-connect to the internet when choosing a market pair
       -> Autocomplete won't be available however as long as valid pair is entered, there won't be any issue starting the strategy
2. Start bot, disconnect from the internet, then create a strategy
    - Re-connect to the internet when asked to enter market pair
       -> Autocomplete list would be available
3. Testing during a slow connection
    - Some connectors won't display autocomplete lists ( KuCoin, Probit, Balancer, and Uniswap) however as long as the correct market is entered, the strategy will start without issue.

PR #3063, opened 2021-03-10
feat/ rate oracle

Nullably:

- See linked issues.

**For testers**
- You won't be able to see how it works with strategies yet, that will be on another story/PR.
- Features proposed here have been implemented. https://github.com/CoinAlpha/hummingbot/issues/2939#issuecomment-794966621


rxlxrxsx:
- Setting the `rate_oracle_source` to `coingecko` and entering a non existing token on `global_token` results to this error when trying the `rate` command and it's arguments. This is not the case when replicating the behavior above but with `rate_oracle_source` as `binance` and instead it shows that `Rate is not available`.

With coingecko as rate_oracle_source
Screen Shot 2021-03-17 at 2 47 49 PM

With binance as rate_oracle_source
Screen Shot 2021-03-17 at 3 00 02 PM

Log file for the error: [logs_hummingbot.log](https://github.com/CoinAlpha/hummingbot/files/6154300/logs_hummingbot.log)

Nullably:
@rxlxrxsx Thanks, good catch.
I have updated the branch with the followings:
* Catch and report error better while rate command still shows rate unavailable
* Add auto completer for rate oracle sources.

There is still a slight difference between `binance` and `coingecko` when `global_token` is specified. `global_token` is needed for coingecko API to get rates while not so for `binance`.

PtrckM (dismissed):
All issue above has been addressed. Tested and confirmed working by QA.

Nullably:
I just realised CoinGecko has an API end point for their supported vs currency. So I have made a small adjustment to our `global_token` works. It will no longer give error if the token is not on the supported list, and will give `rate not available` message on `rate --token`.

PtrckM:
Tested and confirmed working.
- when using not supported (not on list) global_token
     - bot no longer throws error on right pane
     - will display `Rate is not available` with the current rate source
     - works for both --token and --pair

PtrckM (dismissed):

test/test_async_ttl_cache.py
```diff
@@ -9,10 +9,14 @@ class AsyncTTLCacheUnitTest(unittest.TestCase):
 
     @async_ttl_cache(ttl=3, maxsize=1)
     async def get_timestamp(self):
-        return time.time()
+        timestamp = time.time()
+        print(f"timestamp: {timestamp}")
+        return timestamp
```
CrimsonJacket, line 14: I think this inclusion wasn't intended.

Nullably: Oh, yep, you're right, this wasn't intentional.

CrimsonJacket (approved):
LGTM 🍵

rxlxrxsx (approved):
Re-tested and confirmed no other issues found by QA.

PR #3102, opened 2021-03-20
Fix: code quality issues

withshubh:

### Description
Hi :wave: I work at [DeepSource](https://deepsource.io), I ran DeepSource analysis on the forked copy of this repo and found some interesting [code quality issues](https://deepsource.io/gh/withshubh/hummingbot/issues/) in the codebase, opening this PR so you can assess if our platform is right and helpful for you.

### Summary of changes
 

- Remove length check in favour of truthiness of the object
- Remove unnecessary use of comprehension
- Convert `async` magic method into a regular method
- Replace `assertEquals` with `assertEqual`
- Use literal syntax to create data structure
- Simplify the `if` statement
- Raise `NotImplementedError` in abstract method
- Access attribute directly 
- Added .deepsource.toml config to continuously analyze the repo


TheHolyRoger:
I haven't tested this but I think those getattr changes might be breaking depending on python version? This does seem a bit overkill... flake does the job well and in fact more rules have been disabled over time for the extra headache it adds when using third party code

withshubh:
Hi @TheHolyRoger :wave: 
> I haven't tested this but I think those getattr changes might be breaking depending on python version? This does seem a bit overkill... flake does the job well and in fact more rules have been disabled over time for the extra headache it adds when using third party code

Here's the [description](https://deepsource.io/gh/withshubh/hummingbot/issue/PTC-W0034/description/) about getattr changes. Do you want me to revert these changes?

It's not overkilling, it's worth integrating DeepSource. After integrating DeepSource, you do not have to worry about flake8 and disabling its rules as DeepSource internally runs black, isort, mypy, and flake8. In addition to that: if your codebase is strictly following the conventions from these tools, instead of failing the checks and depending on the contributors to fix it, DeepSource Transformers can do this for you (commit to the same PR with the fixes). Also, you won't need to look after the version upgrades for these tools.

Some other features which you may like:
- Fix the issues. DeepSource can automatically fix some of the issues it detects (also includes some flake8 issues) with just a click. Read more about this [here](https://deepsource.io/blog/code-formatting-on-autopilot/).
- DeepSource's own code quality checks.
- Option to analyze only modified code: DeepSource will show you only newly introduced code quality issues for a changeset. Read more about it [here](https://deepsource.io/blog/release-granular-diffs/).

keithbaum:
Hi @withshubh, the point of decorating with @abstractmethod apply method in hummingbot/model/db_migration/base_transformation.py is to raise an exception if not implemented, but at the same time using pass allows a default behavior to be called by child classes using super.
The ABC class with its decorator comes to solve the need to use NotImplementedError because it fails at instantiation time instead of execution of the method.
I personally liked the way it's explained here https://dbader.org/blog/abstract-base-classes-in-python

withshubh:
> Hi @withshubh, the point of decorating with @AbstractMethod apply method in hummingbot/model/db_migration/base_transformation.py is to raise an exception if not implemented, but at the same time using pass allows a default behavior to be called by child classes using super.
> The ABC class with its decorator comes to solve the need to use NotImplementedError because it fails at instantiation time instead of execution of the method.
> I personally liked the way it's explained here [dbader.org/blog/abstract-base-classes-in-python](https://dbader.org/blog/abstract-base-classes-in-python)

I have reverted the fix :sparkles:

TheHolyRoger:
@withshubh If you look at the build log for your changes: https://github.com/CoinAlpha/hummingbot/runs/2221682913

You'll see that your changes are failing flake8, something you claimed "DeepSource can automatically fix some of the issues it detects (also includes some flake8 issues)"...

As I said before, this seems overkill especially if it's adding more errors instead of fixing them.

withshubh:
> @withshubh If you look at the build log for your changes: [CoinAlpha/hummingbot/runs/2221682913](https://github.com/CoinAlpha/hummingbot/runs/2221682913)

Hi @TheHolyRoger :wave: I have fixed the flake8 errors.

> You'll see that your changes are failing flake8, something you claimed "DeepSource can automatically fix some of the issues it detects (also includes some flake8 issues)"...
> 
> As I said before, this seems overkill especially if it's adding more errors instead of fixing them.


DeepSource offers [transformers](https://deepsource.io/docs/transformer/) like autopep8 and isort which you have to enable while creating the configuration file (`.deepsource.toml`) which obviously I didn't do in my configuration for this repo and that was the reason for the CI to throw flake8 errors.

phbrgnomo:
@Nullably could you review these fixes when you have the time to make a call if we are going to merge them or not?

Nullably:
I am ok with the purposed changes, though I think we'll need to first decide if we're going to use Deepsource.
@martinkou What do you think?

Also, the changes are minor but QA will still need to test all these areas where the code changed, which is a lot of testing. 

Note that Deepsource doesn't seem to detect / recommend fixes in all area of our repo, limitation of the tool?

withshubh:
> I am ok with the purposed changes, though I think we'll need to first decide if we're going to use Deepsource.
> @martinkou @dennisocana What do you think?
> 
> Also, the changes are minor but QA will still need to test all these areas where the code changed, which is a lot of testing.
> 
> Note that Deepsource doesn't seem to detect / recommend fixes in all area of our repo, limitation of the tool?

There's no such limitation of the tool. DeepSource analyzes each and every file of the repo for code quality issues unless explicitly defined exclude file patterns in the configuration file.

Why do you think so? @Nullably

TheHolyRoger:
@Nullably  @dennisocana 
https://github.com/OpenMined/PySyft/pull/5261#issuecomment-806281105

I suggest scrolling through his profile :)

PR #3118, opened 2021-03-26
Fix / Rate Oracle returns None price for `USD`

TheHolyRoger:

I'm not sure if this is the best way to handle this, but the new rate oracle returns None price for `USD` pairs (e.g. CoinZoom and HitBTC).
This fixes the issue with the `balance` command for me as seen here:
https://github.com/CoinAlpha/hummingbot/pull/3112#issuecomment-808150326

@Nullably could you take a look at this?

### DIFF
--- a/hummingbot/core/rate_oracle/rate_oracle.py
+++ b/hummingbot/core/rate_oracle/rate_oracle.py
@@ -95,6 +95,7 @@ class RateOracle(NetworkBase):
     @classmethod
     async def global_rate(cls, token: str) -> Decimal:
         prices = await cls.get_prices()
+        token = "USDT" if token == "USD" else token
         pair = token + "-" + cls.global_token
         return find_rate(prices, pair)
 


TheHolyRoger:
Whilst this works as a hotfix, Rate Oracle in general should probably be able to use order book prices from connectors in the long term? That would fix this as well as other tokens that aren't on coingecko/binance, and use more real-time prices

dennisocana:
Moving this in Review/QA pipeline since this affects both HitBTC and CoinZoom connectors. Requested review from Jack

Nullably (changes requested):
This is not a good idea since one is a stable coin (USDT) and the other is fiat currency (USD) and often their values slightly differ.

Nullably:
The problem is Binance doesn't offer USD trade pairs (only BinanceUS does). If you set the oracle to CoinGecko, this should fix the problem.

Nullably:
> Whilst this works as a hotfix, Rate Oracle in general should probably be able to use order book prices from connectors in the long term? That would fix this as well as other tokens that aren't on coingecko/binance, and use more real-time prices

Order books are available only after a strategy is started and only on pairs set in the strategy, this is too limited. Technically we can code the oracle to look for prices in the loaded order books, though this creates more complexity while not adding much value.

TheHolyRoger:
> This is not a good idea since one is a stable coin (USDT) and the other is fiat currency (USD) and often their values slightly differ.

@Nullably both CoinZoom and HitBTC explicitly call "USDT" USD. There is no fiat "USD" on them.

Should this be handled in the connectors themselves? In both HitBTC and CoinZoom I could do the replacement in `convert_to_exchange_trading_pair` and `convert_from_exchange_trading_pair` instead?

I'm not sure if any other existing connectors also refer to USDT as USD though...?

TheHolyRoger:
Screenshot 2021-03-29 at 03 35 51
@Nullably as you say, switching to coingecko does also fix the issue, although still 0 value for the ZOOM token which isn't on coingecko.

I think coingecko should really be the default option for the rate oracle too since it has far more pairs (and all of the binance pairs). I can see this being a regular support ticket otherwise... :D

Nullably:
> > This is not a good idea since one is a stable coin (USDT) and the other is fiat currency (USD) and often their values slightly differ.
> 
> @Nullably both CoinZoom and HitBTC explicitly call "USDT" USD. There is no fiat "USD" on them. 

Oh, ok, now I understand. That is confusing. 

> Should this be handled in the connectors themselves? In both HitBTC and CoinZoom I could do the replacement in `convert_to_exchange_trading_pair` and `convert_from_exchange_trading_pair` instead?

You mean to convert USD to USDT, right? I think this is a good idea, it makes it consistent with other connectors.
@phbrgnomo What do you think?

> I'm not sure if any other existing connectors also refer to USDT as USD though...?

Not that I'm aware of.

TheHolyRoger:
> > Whilst this works as a hotfix, Rate Oracle in general should probably be able to use order book prices from connectors in the long term? That would fix this as well as other tokens that aren't on coingecko/binance, and use more real-time prices
> 
> Order books are available only after a strategy is started and only on pairs set in the strategy, this is too limited. Technically we can code the oracle to look for prices in the loaded order books, though this creates more complexity while not adding much value.

@Nullably afraid i disagree about "not adding much value" - if the rate oracle is used down the road for arbitrage strategies for price conversion, then the price needs to be as real-time as possible for an accurate profitability calculation. In my opinion it would be better to check for an initialised order book and use that first, with coingecko/binance as a fallback if not.

TheHolyRoger:
> @Nullably both CoinZoom and HitBTC explicitly call "USDT" USD. There is no fiat "USD" on them.

@Nullably @dennisocana apologies - CoinZoom DOES use both USDT and USD (Fiat) I was wrong about that...
HitBTC uses USD for USDT though, no fiat option.

PR #3161, opened 2021-04-05
[digifinex] add user-agent header

szmcdull:
Add user-agent header


keithbaum (approved):
Change is ok

dennisocana (approved):
Digifinex team confirmed that the orders are getting picked up. LGTM

PR #3175, opened 2021-04-06
feat / add argument details and values to single-market strategy classes

kmcgowan2000:


phbrgnomo:
@dennisocana This is the PR that was mentioned on latest product meeting.

@kmcgowan2000 Could you create an Issue linked to this PR describing what is the goal?

kmcgowan2000:
> @dennisocana This is the PR that was mentioned on latest product meeting.
> 
> @kmcgowan2000 Could you create an Issue linked to this PR describing what is the goal?

@phbrgnomo yeah that's created and linked now, and the first commit to this PR is a completed example (PMM strat)

setup.py
```diff
@@ -102,6 +106,11 @@ def main():
         ],
     }
     install_requires = [
+        "0x-contract-addresses",
```
Nullably, line 109: Why do we need all of these 0x libraries?

Nullably (changes requested):

setup.py
kmcgowan2000: can't import from hummingbot.connector.utils without them

kmcgowan2000: I guess we could split it off into a specific extras_require ? if it's only really useful for hummingsim users

phbrgnomo:
These new parameters are being created to allow Hbot pro read the information of the strategy configuration.

The PR cover only strategies that trade on a single pair.

After that, we need to expand these values to other strategies, to allow Hbot pro also access this information.

We also need to add on the development docs an explanation about what these parameters are used for. @kmcgowan2000 can help on writing the docs.

@dennisocana PR is ready to review and merge. The test is related to Hbot pro, so after approving the PR, contact @kmcgowan2000 for tests.

setup.py
Nullably: I think that's fine, I just couldn't find where they are used.

Nullably (approved):
LGTM

Nullably:
@phbrgnomo @dennisocana 
Note that not all strategies and connectors can be used by Hummingsim (thus Hummingbot Pro). Python strategies will need to first be converted to Cython. Gateway and derivative connectors are not available as we never collect order book data for them (or they don't have order books).

Nullably:
A simpler solution is proposed. I'm closing this one.

PR #3193, opened 2021-04-09
Changed calculation of avellaneda_market_making parameters to happen at order proposal

keithbaum:
- Changed calculation of avellaneda_market_making parameters to happen only when order_proposal is created instead of every tick. This reduces computation cost (for hummingsim backtesting)
- Added __init__ file to trailing_indicators folder


RC-13 (approved):
LGTM, no issue encountered during testing.

hummingbot/strategy/avellaneda_market_making/avellaneda_market_making.pyx
```diff
@@ -568,7 +569,7 @@ cdef class AvellanedaMarketMakingStrategy(StrategyBase):
             # KAPPA
             # Want the maximum possible spread but with restrictions to avoid negative kappa or division by 0
             max_spread_around_reserved_price = max_spread * (2-self._inventory_risk_aversion) + min_spread * self._inventory_risk_aversion
-            if max_spread_around_reserved_price <= self._gamma * (vol ** 2):
+            if (max_spread_around_reserved_price * self._gamma - (vol * self._gamma) **2) <= s_decimal_zero:
```
CrimsonJacket, line 572: Is there a reason for this change? Mathematically they are the same.

CrimsonJacket (approved):
Just one quick question regarding the if-condition for the calculation of kappa. 
Also, including the `__init__.py` file in PR #3194 would be more appropriate.
Other than that, LGTM.

hummingbot/strategy/avellaneda_market_making/avellaneda_market_making.pyx
keithbaum: Decimal minimum possible number was sometimes mapped to 0 when multiplying by gamma**2. That condition was not emulated when using the original condition. By using this condition which is exactly the same calculation made in the denominator, this situations are caught

CrimsonJacket (approved):
👍

PR #3195, opened 2021-04-09
fix / avellaneda status msg

keithbaum:
Changed condition to check for "volatility is not NaN" since when volatility=0 although being valid number "bool(volatility) =False" and status message was not showing. 


RC-13:
Tested and confirmed that PR fix the reported issue. 
Liquid:
image
Kraken:
image
Observe the bots for 1 hour and the strategy parameters doesn't go away

RC-13 (dismissed):

hummingbot/strategy/avellaneda_market_making/avellaneda_market_making.pyx
```diff
@@ -355,7 +355,7 @@ cdef class AvellanedaMarketMakingStrategy(StrategyBase):
             lines.extend(["", "  No active maker orders."])
 
         volatility_pct = self._avg_vol.current_value / float(self.get_price()) * 100.0
-        if all((self._gamma, self._kappa, volatility_pct)):
+        if all((self._gamma, self._kappa, volatility_pct is not NaN)):
```
CrimsonJacket, line 358: I don't think this is correct. I do know that checking if a variable is `NaN`, we can't just use `a is not NaN` or `a != NaN`. We should use `math.isnan()` or its equivalent.

image

CrimsonJacket (changes requested):
I don't think this is correct. I do know that checking if a variable is `NaN`, we can't just use `a is not NaN` or `a != NaN`. We should use `math.isnan()` or its equivalent.

image

hummingbot/strategy/avellaneda_market_making/avellaneda_market_making.pyx
keithbaum: Change applied

PtrckM (approved):
Re-test no issue found (run time 1hr), reapproving
image

CrimsonJacket (approved):
LGTM

Issue #3134, opened 2021-03-30
Gitcoin Grant: Support EVM chain AMM Arbitrage ($30k)


thebaoman:
Currently Hummingbot supports AMM arbitrage between any exchange and Uniswap.

Hummingbot does this through the Gateway API (https://github.com/CoinAlpha/gateway-api) that connects to Ethereum and defines the Uniswap Router contract within its code.

By changing the RPC node that Hummingbot connects to its easy enough to interact with new chains, but hardcoded values related to ChainIDs, custom tokens, router contracts, and loading other modules (such as xDAI always attempting to load balances from Perp Finance) mean that users cannot attach the system to another chain.

Since EVM chains are growing in popularity and each chain now has multiple Uniswap forks, we're funding an update for a party to improve Hummingbot and their Gateway API to support EVM uniswap forks in AMM.

The fix should:

- Allow users to overwrite and declare a ChainID and chain name. (This may require modification to libraries used in defining the chain)
- Allow users to define the router (or if needed factory) for the Uniswap fork they are using. This could be in the connector section of Hummingbot but will also be in the Gateway-API where these values are hard coded.
- (I believe the connector looks up pairs using the router or factory address, but if not and it calculates the pairs, a user will need to declare the unique hex string used in the factory for pair calculation)
- Allow users to declare what the base asset ticker ID is when it is not ETH as this may be crucial for price lookup.
- Allow users to override gas fees.
- Allow users to override trade fee in the event that it is not the standard 0.3%.

In order for this work to be accepted for a bounty it must:

1. Be tested and demonstrate on at least 3 EVM based networks that it can be configured to run the AMM-Arb model on different networks and Uniswap forks.
2. Must be flexible in declaring new networks without needing substantial updates. (i.e. user declaration of a new network ID and RPC should be sufficient)
3. Must not create new upstream dependencies that will not be maintained. (i.e. if ChainID is declared from an upstream package, a new package should not be created to replace it. Instead a work around should be implemented directly in Hummingbot so that new dependencies are not created)
4. Must be reviewed by the CoinAlpha team, meeting their contribution standards and merged into master. Please read the CoinAlpha contribution guidelines carefully and be sure to comply with their standards.

The Bao Finance core team is putting up a $30k bounty on this issue paid in ETH via Gitcoin.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


[...]

PR #3220, opened 2021-04-16
refactor / arbitrary evm uniswap v2

jondwillis:

[For grant: Issue #3134](https://github.com/CoinAlpha/hummingbot/issues/3134)

[See also: gateway-api](https://github.com/CoinAlpha/gateway-api/pull/53)

### Description
Currently Hummingbot supports AMM arbitrage between any exchange and Uniswap.

Hummingbot does this through the Gateway API (https://github.com/CoinAlpha/gateway-api) that connects to Ethereum and defines the Uniswap Router contract within its code.

By changing the RPC node that Hummingbot connects to its easy enough to interact with new chains, but hardcoded values related to ChainIDs, custom tokens, router contracts, and loading other modules (such as xDAI always attempting to load balances from Perp Finance) mean that users cannot attach the system to another chain.

Since EVM chains are growing in popularity and each chain now has multiple Uniswap forks, we're funding an update for a party to improve Hummingbot and their Gateway API to support EVM uniswap forks in AMM.

The fix should:

- [x] Allow users to overwrite and declare a ChainID and chain name. (This may require modification to libraries used in defining the chain)
- [x] Allow users to define the router (or if needed factory) for the Uniswap fork they are using. This could be in the connector section of Hummingbot but will also be in the Gateway-API where these values are hard coded.
(I believe the connector looks up pairs using the router or factory address, but if not and it calculates the pairs, a user will need to declare the unique hex string used in the factory for pair calculation)
- [x] Allow users to declare what the base asset ticker ID is when it is not ETH as this may be crucial for price lookup.
- [x] (already implemented) Allow users to override gas fees. 
- [x] (already implemented) Allow users to override trade fee in the event that it is not the standard 0.3%.
In order for this work to be accepted for a bounty it must:

- [x] Be tested and demonstrate on at least 3 EVM based networks that it can be configured to run the AMM-Arb model on different networks and Uniswap forks.
- [ ] Must be flexible in declaring new networks without needing substantial updates. (i.e. user declaration of a new network ID and RPC should be sufficient)
- [x] Must not create new upstream dependencies that will not be maintained. (i.e. if ChainID is declared from an upstream package, a new package should not be created to replace it. Instead a work around should be implemented directly in Hummingbot so that new dependencies are not created)
- [ ] Must be reviewed by the CoinAlpha team, meeting their contribution standards and merged into master. Please read the CoinAlpha contribution guidelines carefully and be sure to comply with their standards.


dennisocana:
@mifeng please advise on the latest update on this PR and attached issue

jondwillis:
@dennisocana the last plan I had would be to throw most of this work away and rebase on the latest hummingbot / gateway branches.

https://github.com/CoinAlpha/gateway-api/pull/53#issuecomment-906637722

mifeng:
> @dennisocana the last plan I had would be to throw most of this work away and rebase on the latest hummingbot / gateway branches.

Thanks for doing this work @jondwillis - even though we are subsuming it into the new Gateway V2 architecture, we really appreciate your efforts and will try to make it easier for external devs to contribute in the future with Gateway V2.

PR #3235, opened 2021-04-22
Fix/avellaneda eta calculation

keithbaum:
Fixed eta calculation. Now at the maximum distance from inventory_target, opposing order decays 90% from original amount

With this change in case inventory_risk_aversion is the maximum possible and inventory_target is the furthest possible (i.e. inventory_target = 100% with initial inventory = 0), opposing order will reduce its order_amount in 90%.
On the other side, if inventory_risk_aversion is the minimum possible, it won't matter distance to inventory_target, it will always have symmetrical orders in both sides.

* Example with inventory_risk_aversion almost equal to 0

image

With inventory_target far from the current inventory, anyways order_amounts are equal to config value

image

* Example with inventory_risk_aversion almost equal to 1 and inventory the furthest possible from target

image

In this extreme case opposing order's order_amount = 0.1 * config order_amount

image


PtrckM (approved):
No issue found. Confirmed working and passed QA testing.

CrimsonJacket (approved):
LGTM

Issue #3226, opened 2021-04-19
Binance Perpetual - Error fetching funding income


RC-13:
**Describe the bug**
Opening ticket for tracking: The issue is quite difficult to replicate since we have to wait for the funding payment. During this period, the bot encounters the following error:
```ruby
2021-04-19 08:00:00,941 - 2327 - hummingbot.connector.derivative.binance_perpetual.binance_perpetual_derivative - ERROR - Error fetching /fapi/v1/income
Traceback (most recent call last):
  File "/home/ubuntu/binance_perpetual/hummingbot/connector/derivative/binance_perpetual/binance_perpetual_derivative.py", line 1016, in request
    error_response = await response.json()
  File "/home/ubuntu/miniconda3/envs/hummingbot/lib/python3.8/site-packages/aiohttp/client_reqrep.py", line 1026, in json
    raise ContentTypeError(
aiohttp.client_exceptions.ContentTypeError: 0, message='Attempt to decode JSON with unexpected mimetype: ', url=URL('https://fapi.binance.com/fapi/v1/income?incomeType=FUNDING_FEE&limit=1&recvWindow=20000&symbol=ETHUSDT&timestamp=1618819200000&signature=588efa81ddfe71cfdf89c099fc050a9b60936a5028ed77a5096e7d09f1947bbd')
2021-04-19 08:00:00,941 - 2327 - hummingbot.connector.derivative.binance_perpetual.binance_perpetual_derivative - WARNING - 0, message='Attempt to decode JSON with unexpected mimetype: ', url=URL('https://fapi.binance.com/fapi/v1/income?incomeType=FUNDING_FEE&limit=1&recvWindow=20000&symbol=ETHUSDT&timestamp=1618819200000&signature=588efa81ddfe71cfdf89c099fc050a9b60936a5028ed77a5096e7d09f1947bbd')
2021-04-19 08:00:00,942 - 2327 - hummingbot.connector.derivative.binance_perpetual.binance_perpetual_derivative - ERROR - Unexpected error in user stream listener loop: 'ContentTypeError' object is not subscriptable
Traceback (most recent call last):
  File "/home/ubuntu/binance_perpetual/hummingbot/connector/derivative/binance_perpetual/binance_perpetual_derivative.py", line 559, in _user_stream_event_listener
    await self.get_funding_payment()
[...]

PR #3346, opened 2021-05-10
Fix / Fix funding fee events on binance perpetual

zappra:

Fixes for get_funding_info() on binance perpetual - #3226 

Replace HTTP POST with GET
Correctly parse returned data which is two nested arrays
Use correct trading pair key for _funding_info dict

The rather confusing "attempt to decode json with incorrect mime type" error in the linked issue was because the use of POST instead of GET was returning HTTP status 405 and no json payload - maybe the error handing needs improving?


zappra:
Marking this as WIP; the function is executing now but I'm getting duplicate events with the wrong funding amounts so will investigate more

zappra:
There's a couple of problems here

The ACCOUNT_UPDATE event is received for every fee payment that occurs, regardless of pair, so if running multiple bots you'll get account update events for other bots' pairs

When retrieving funding payment info from fapi/v1/income it seems a small delay might be required since in live trading I was getting old funding payments, not the current ones

Am testing an implementation that explicitly polls for funding payments at the specified 8 hour times

zappra:
Have working fix now:

Remove the check for funding fees in account updates
Added polling task to retrieve funding fees at correct UTC times 
Added "startTime" to get_funding_fee to ensure we only get the most recent funding payment

PtrckM:
- Got an error during funding fee event, but bot still able to retry and complete.
image
image

```2021-05-21 21:34:00,063 - 2054 - hummingbot.connector.derivative.binance_perpetual.binance_perpetual_derivative - ERROR - Unexpected error updating funding info. Retrying after 10 seconds.$
  File "/home/ubuntu/miniconda3/envs/hummingbot/lib/python3.8/site-packages/websockets/protocol.py", line 827, in transfer_data
    message = await self.read_message()
  File "/home/ubuntu/miniconda3/envs/hummingbot/lib/python3.8/site-packages/websockets/protocol.py", line 895, in read_message
    frame = await self.read_data_frame(max_size=self.max_size)
  File "/home/ubuntu/miniconda3/envs/hummingbot/lib/python3.8/site-packages/websockets/protocol.py", line 971, in read_data_frame
    frame = await self.read_frame(max_size)
  File "/home/ubuntu/miniconda3/envs/hummingbot/lib/python3.8/site-packages/websockets/protocol.py", line 1047, in read_frame
    frame = await Frame.read(
  File "/home/ubuntu/miniconda3/envs/hummingbot/lib/python3.8/site-packages/websockets/framing.py", line 105, in read
    data = await reader(2)
  File "/home/ubuntu/miniconda3/envs/hummingbot/lib/python3.8/asyncio/streams.py", line 721, in readexactly
    raise exceptions.IncompleteReadError(incomplete, n)
asyncio.exceptions.IncompleteReadError: 0 bytes read on a total of 2 expected bytes

The above exception was the direct cause of the following exception:

  File "/home/ubuntu/binance_perpetual/hummingbot/connector/derivative/binance_perpetual/binance_perpetual_derivative.py", line 695, in _funding_info_polling_loop
    raw_msg: str = await asyncio.wait_for(ws.recv(), timeout=10.0)
  File "/home/ubuntu/miniconda3/envs/hummingbot/lib/python3.8/asyncio/tasks.py", line 483, in wait_for
    return fut.result()
  File "/home/ubuntu/miniconda3/envs/hummingbot/lib/python3.8/site-packages/websockets/protocol.py", line 509, in recv
    await self.ensure_open()
  File "/home/ubuntu/miniconda3/envs/hummingbot/lib/python3.8/site-packages/websockets/protocol.py", line 803, in ensure_open
    raise self.connection_closed_exc()
websockets.exceptions.ConnectionClosedError: code = 1006 (connection closed abnormally [internal]), no reason
```

- logs
[logs_perp_mm_eth-usdt.log](https://github.com/CoinAlpha/hummingbot/files/6525972/logs_perp_mm_eth-usdt.log)
image

zappra:
Those two errors are not connected to the funding payments - they're unrelated and both have a retry mechanism. 

Worth pointing out that the function that retrieves the funding payment doesn't have a retry currently, so if an error did occur when querying the payments no events would be generated

rxlxrxsx (dismissed):

dennisocana:
@phbrgnomo QA confirmed the fix and we can probably get a quick review from @vic-en today. If all looks good, we can include this in next week's v0.40 release instead. Let me know what you think.

zappra:
I'd suggest making a default 1 minute delta and adding a retry mechanism up to some maximum number, eg for 1600 UTC, use start time of 15:59 and request payments at 16:01 and every minute thereafter for up to attempts

hummingbot/connector/derivative/binance_perpetual/binance_perpetual_derivative.py
```diff
@@ -134,9 +135,9 @@ def __init__(self,
         self._user_stream_event_listener_task = None
         self._trading_rules_polling_task = None
         self._funding_info_polling_task = None
+        self._funding_fee_polling_task = None
         self._last_poll_timestamp = 0
         self._throttler = Throttler((10.0, 1.0))
-        self._funding_payment_span = [0, 15]
```
vic-en, line 139: you shouldn't remove this. Some strategies will have issues.

vic-en (changes requested):

vic-en:
> I'd suggest making a default 1 minute delta and adding a retry mechanism up to some maximum number, eg for 1600 UTC, use start time of 15:59 and request payments at 16:01 and every minute thereafter for up to 5 attempts
> 
> (happy to do this btw, its not a big change)

We can also rely on the value from the `self._funding_payment_span`. Binance mentioned that funding payment settlement can take up to 15 seconds. Hence, we can use that value in addition to a retry logic to help if the first attempt fails.

zappra:
Have added retry mechanism and changed to use time spans in self._funding_payment_span

Will test for a couple days then update the branch

hummingbot/connector/derivative/binance_perpetual/binance_perpetual_derivative.py
zappra: Fixed in latest commit

zappra:
Updated to retry in the event of failure, and using funding span info for timing

rxlxrxsx (approved):
Re-tested and confirmed working by QA.

vic-en (approved):
LGTM. Nice job!

PR #3357, opened 2021-05-11
Fix wrong module references in tests

aarmoa:
I made some changes to import statements in some tests, that where referencing to incorrect or inexistent modules.
The module with more changes is test/integration/test_composite_order_book.py. I had to mark the test as skipped since it has dependencies with hummingsim, that is no longer a module inside the client project. For the test to run without failing I also had to comment out several parts. I did not remove it so that we can discuss how to replace it.

### DIFF
--- a/bin/hummingbot.py
+++ b/bin/hummingbot.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-import path_util        # noqa: F401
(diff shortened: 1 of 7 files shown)


aarmoa:
I changed the base branch as per @Nullably request

Nullably (approved):
LGTM

PR #3364, opened 2021-05-12
Refactor/ move tests to proper location

Nullably:

- Move connector tests to connector folder
- Remove integration folder
- Tests that are legacy or don't pass are moved to debug


test/hummingbot/strategy/amm_arb/test_utils.py
```diff
@@ -35,10 +36,14 @@ def get_order_price(self, trading_pair: str, is_buy: bool, amount: Decimal) -> D
 
 
 class AmmArbUtilsUnitTest(unittest.TestCase):
+
     def test_create_arb_proposals(self):
+        asyncio.get_event_loop().run_until_complete(self._test_create_arb_proposals())
```
aarmoa, line 41: I think it would be better to use asyncio.run()

aarmoa (commented):
> The new module hummingbot/core/mock_api/mock_web_socker_server.py has a type (socker instead of socket).
> We should favor the use of asyncio.run() instead of asyncio.get_event_loop().run_until_complete() wherever we don't need to change the configuration of the loop.

test/hummingbot/strategy/amm_arb/test_utils.py
Nullably: agree, since we are now on python 3,8, this is now possible.

Nullably:
> > The new module hummingbot/core/mock_api/mock_web_socker_server.py has a type (socker instead of socket).
> > We should favor the use of asyncio.run() instead of asyncio.get_event_loop().run_until_complete() wherever we don't need to change the configuration of the loop.

Thanks for spotting that, they are both addressed.

james-hummingbot (approved):

PR #3386, opened 2021-05-18
fix/ AscendEx not handling partial fills

CrimsonJacket:


RC-13:
Re-posting comment: history appears to be incorrect after partial fill:
image
In this image, my order amount is 50. I have a partial fill of 33 on the sell side, then after few seconds, the remaining got filled. If you run `history --verbose` the first partial fill shows the correct value (33) however the last fill shows 50 instead of 17. This resulted in an incorrect total volume for the sell-side both ZIL and USDT volume. 

The expected total ZIL after the fill should be 50 and USDT volume should be 7.7362
image

RC-13 (approved):
Confirming fix for the reported issue. Partial fills are now recorded and displayed with the correct size and added accordingly. LGTM

james-hummingbot (approved):

PR #3391, opened 2021-05-18
Refactor/unit tests test estimate fee

james-hummingbot:

I would like @aarmoa 's opinion on this approach to documentation and unit testing. This is a simple addition but I would like this to serve as an example of better documentation and unit testing.


aarmoa (commented):
I think the change is good. I have one comment regarding the assertions for the exceptions. In my opinion it is better to validate also that the message or string representation of the exception is the expected one. That can be done with assertRaises and also with assertRaisesRegex.

PR #3402, opened 2021-05-18
Refactor/unit tests test ethereum

james-hummingbot:

Other than adding tests, I also added comments to the original code.

### DIFF
--- a/hummingbot/core/utils/ethereum.py
+++ b/hummingbot/core/utils/ethereum.py
@@ -8,6 +8,9 @@ from hummingbot.core.utils import async_ttl_cache
 
 
 def check_web3(ethereum_rpc_url: str) -> bool:
+    """
+    Confirm that the provided url is a valid Ethereum RPC url.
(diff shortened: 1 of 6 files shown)


hummingbot/core/utils/ethereum.py
```diff
@@ -27,23 +31,13 @@ def check_web3(ethereum_rpc_url: str) -> bool:
     return ret
 
 
-def block_values_to_hex(block: AttributeDict) -> AttributeDict:
```
james-hummingbot, line 30: code search revealed this one never used. Please double check!

aarmoa (approved):
The changes look good to me.
Regarding "async def _test_fetch_trading_pairs(self):" instead of having an async method that tests everything, I would put all the testing logic in the normal test method, and use asyncio.get_event_loop().run_until_complete() to run any required async model method. In that way all the testing logic and the asserts stay in the proper test.

PR #3421, opened 2021-05-21
Test functions from async_retry

james-hummingbot:


aarmoa (commented):
In my experience the use of global variables could cause issues that are difficult to debug and find. In this case I would suggest to move the decorated functions as methods of the test class, and use instead instance variables. It would be also good to have the setUp() method initializing the counters, in case in the future someone adds more tests methods to the class.

aarmoa (approved):
Changes look good to me

PR #3429, opened 2021-05-23
Hackathon/ Multi market for amm_arb

GTaf:
Allows the user that will use the amm_arb startegy to set more than two markets to watch for arbitrage.

The modification to the create_command.py file could be used to allow multi markets for other strategies as well.


dennisocana:
Thank you for participating. The bounty for the hackathon has been awarded to one of the participants: https://gitcoin.co/issue/CoinAlpha/hummingbot/3287/100025613

PR #3430, opened 2021-05-23
Hackathon/AroonOscillatorStrategy

squarelover:
One of the major downsides to many of the Market-Making strategies in Hummingbot is that they don't understand trends. In my experience, I've often had my bots trade on the wrong side of a trend. This is what it frequently looks like: 
Bad Bot

My strategy attempts to take a well-known set of Market Indicators called Aroon Indicators. These indicators collect trade prices over a configurable set of periods of a given duration. The indicators represent how recent the highest highs and the lowest lows are. And the Oscillator indicator can strongly suggest a market trend. I've tried to distill what the indicators signify and use them to adjust spreads so traders are positioned at the hopefully the best point to execute at profitable positions. In other words, it tries to be better at buying low and selling high.

You can learn more about Aroon Indicators, here:
https://www.investopedia.com/terms/a/aroon.asp


rvie17 (commented):
looking fine to me also is not my best language

dennisocana:
@squarelover we're going to review the strategy and test starting next week, please update your branch with the most recent `development` branch.

petioptrv:
@squarelover thank you for the submission. I will be going over the code. Before that however, please rebase this branch on `CoinAlpha:development`, add the strategy's directory to `.coveragerc` to exclude it from the coverage checks, and add the strategy package to `setup.py`'s `packages` list (line 91).

PR #3433, opened 2021-05-24
refactor/ fix and simplify tests in test_ethereum

james-hummingbot:

Originally I setup the tests to create a server. While it worked well on Linux it seems like it was a problem on Mac. We have decided to the patch system instead of a mock server. This simplifies the code, makes the tests faster to run and avoid networking issues we ran into cross platform.


aarmoa (approved):
The change is an important improvement to the validation in the test

PR #3459, opened 2021-05-30
(fix) Fix _parameters_based_on_spread=False for avellaneda model

ericjang:
The calculation of spread_inflation_due_to_volatility crashes when self._parameters_based_on_spread = False, failing with the following error. This change fixes that by moving it into the self._parameters_based_on_spread branch.

```
    spread_inflation_due_to_volatility = max(self._vol_to_spread_multiplier * vol, price * self._min_spread)/(price * self._min_spread)
TypeError: unsupported operand type(s) for *: 'decimal.Decimal' and 'NoneType'
```


ericjang:
Oops, realized that development branch already has this change in.

PR #3466, opened 2021-05-31
Feat/amm arb oracle

aarmoa:
Change in amm_arb strategy to make it posible to use rate oracle to look for rates.
There is a new parameter that enables the use of the oracle to get the price of quote tokens in ETH.
Also the status message was extended to include the information about the quote rates in ETH, that shows the user if the strategy is using the oracle or not.

https://app.zenhub.com/workspaces/hummingbot-5f340755c5e17300167a8f30/issues/coinalpha/hummingbot/3216
https://github.com/CoinAlpha/hummingbot/issues/3216


hummingbot/templates/conf_amm_arb_strategy_TEMPLATE.yml
```diff
@@ -33,3 +33,7 @@ market_2_slippage_buffer: null
 # If false, the bot will wait for first exchange order filled before submitting the other order
 concurrent_orders_submission: null
 
+# Whether to use rate oracle on unmatched trading pairs
+# Set this to either True or False
+use_oracle_conversion_rate: null
+
```
PtrckM, line 39: Should we increment the version on https://github.com/CoinAlpha/hummingbot/blob/2c7403db8c7d6574bcce0b0f5ad350db11add633/hummingbot/templates/conf_amm_arb_strategy_TEMPLATE.yml#L5 when adding new parameter on the template?

aarmoa: That is a good point. I have committed the change.

hummingbot/strategy/amm_arb/amm_arb_config_map.py
```diff
@@ -121,4 +121,11 @@ def order_amount_prompt() -> str:
         default=False,
         validator=validate_bool,
         type_str="bool"),
+    "use_oracle_conversion_rate": ConfigVar(
```
Nullably, line 124: You'll need to set `rate_oracle_pairs` global setting variable, you can see Arbitrage config map for an example.

Nullably (changes requested):

aarmoa:
I close this PR because it has been replaced with https://github.com/CoinAlpha/hummingbot/pull/3492

PR #3473, opened 2021-06-01
Fix/  Ascend ex not checking min order amount on quantize

Nullably:

- Add AscendExTradingRule (extends from TradingRule)
- Add `quantize_order_amount` to account for min and max notional order value
- Add `ExchangePyBase` base class, this is to fix a flaw in connector base class where `c_quantize_order_amount` and `c_quantize_order_price` cannot be overwritten by its python sub classes


hummingbot/connector/exchange_py_base.pyx
```diff
@@ -0,0 +1,26 @@
...
+            return price
+        price_quantum = self.c_get_order_price_quantum(trading_pair, price)
+        return ExchangePyBase.quantize_value(price, price_quantum)
+
+    cdef object c_quantize_order_amount(self, str trading_pair, object amount, object price=s_decimal_NaN):
+        return self.quantize_order_amount(trading_pair, amount)
+
+    def quantize_order_amount(self, trading_pair: str, amount: Decimal) -> Decimal:
+        order_size_quantum = self.c_get_order_size_quantum(trading_pair, amount)
+        return ExchangePyBase.quantize_value(amount, order_size_quantum)
```
CrimsonJacket, line 26: I personally feel this could simply be added into `ExchangeBase`. Reason being that these methods could be used by other exchange connectors that extends `ExchangeBase`

CrimsonJacket (approved):
LGTM. As discussed, I initially felt that the functions could proof to be useful in `ExchangeBase` as well. However, considering that having AscendEx inherit the new `ExchangePyBase` class would definitely be a safer approach.

Note: We should definitely spend some time to investigate how we should design & implement the `ExchangeBase` or `ExchangePyBase` across the other connectors.

rapcmia (approved):
Tested and confirmed ok by QA

Steps to reproduce  from issue/3419
1. Create simple PMM, the market used ETH-USDT
2. Change `min_quote_order_amount` to 0 (both ETH and USDT)
3. Import config, change `order_amount` to 0.0019 ($5)
4.  Start the bot

The bot created orders
client
connector


If the bot's order amount is less than $5 it will not create orders anymore and will only create if more than Ascendex's minimum order amount

PR #3479, opened 2021-06-01
Add coverage checks to github workflow

aarmoa:
Add coverage checks to github workflow.
There are two calculations of coverage. The global coverage for the whole project is calculated by nosetests as part of the unit tests run. That run generates the file coverage.xml, with the details. It is configured to fail if the coverage is less than 45%.
Then there is a coverage calculated based on the changes included in the PR compared to the base branch. The changes only coverage is calculated with diff-cover, using the coverage.xml file. This validation fails if the coverage is under 50%.

https://app.zenhub.com/workspaces/hummingbot-5f340755c5e17300167a8f30/issues/coinalpha/hummingbot/3469
https://github.com/CoinAlpha/hummingbot/issues/3469

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -97,7 +97,7 @@ jobs:
           ./compile
 
       # Install tools for coverage calculation
-      - name: Install Hummingbot
+      - name: Install Coverage Tools
         if: steps.program-changes.outputs.cache-hit != 'true' || steps.conda-dependencies.outputs.cache-hit != 'true'
         shell: bash -l {0}
         run: |
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -96,16 +96,6 @@ jobs:
           conda env export
           ./compile
 
-      # Install tools for coverage calculation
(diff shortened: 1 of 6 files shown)


Nullably:
@aarmoa Can we make exceptions in some cases where we don't need to apply coverage check or at a lower percentage, e.g on UI, e.g. commands, where unit testing here will be very difficult if at all possible?
Can the coverage be calculated for cython files as well?

aarmoa:
> @aarmoa Can we make exceptions in some cases where we don't need to apply coverage check or at a lower percentage, e.g on UI, e.g. commands, where unit testing here will be very difficult if at all possible?
> Can the coverage be calculated for cython files as well?

As far as I know there is no easy way to add exceptions. But the command that runs the tests and calculates the coverage receives a parameter indicating the module/modules for which the coverage has to be calculated. Maybe we could have the code that is not easy to cover with tests in separate modules that are not included in the validation.
Regarding cython files, unfortunately the Python tool to calculate coverage can't handle them. One more disadvantage of using Cython.

Nullably (approved):
LGTM

PR #3520, opened 2021-06-06
Fix / Coinzoom balance updates not always firing

TheHolyRoger:

This fix should help this issue, if not fix it I hope!
https://github.com/CoinAlpha/hummingbot/issues/3190


dennisocana:
Thanks @TheHolyRoger - we will review for the July release

rapcmia:
While testing it with Pure market-making with order-refresh of  30s and multiple orders (10) on bid side only to monitor the order creation and the balances, the bot seems to be having delays in cancelling the order. It takes more or less 40s to cancel all orders and there are instances that the bot still inconsistently creating orders due to balance delay. Please see the recording below:

https://user-images.githubusercontent.com/73840223/126282784-331d8eb0-cfbf-4d97-a487-66f9c42bb81c.mp4

Attached config/log files
[3520.zip](https://github.com/CoinAlpha/hummingbot/files/6846553/3520.zip)

TheHolyRoger:
Force pushed this branch, completely rewritten the previous changes using recent connector additions, should behave much better!

hummingbot/connector/exchange/coinzoom/coinzoom_exchange.py
```diff
@@ -103,6 +104,10 @@ def __init__(self,
         self._trading_rules_polling_task = None
         self._last_poll_timestamp = 0
         self._throttler = Throttler(rate_limit = (8.0, 6))
+        self._real_time_balance_update = False
+        self._update_balances_fetching = False
+        self._update_balances_queued = False
```
vic-en, line 109:
```suggestion
        self._update_balances_fetching = False
        self._update_balances_queued = False
```
I think we should remove this to simplify balance updates.

hummingbot/connector/exchange/coinzoom/coinzoom_exchange.py
```diff
@@ -589,8 +600,35 @@ async def _update_balances(self):
...
+                if self._update_balances_finished.is_set():
+                    self._update_balances_finished = asyncio.Event()
+                else:
+                    self._update_balances_finished.set()
+                self._update_balances_queued = False
+            if self._update_balances_fetching:
+                self._update_balances_fetching = False
+            warn_msg = (f"Could not fetch balance update from {Constants.EXCHANGE_NAME}")
+            self.logger().network(f"Unexpected error while fetching balance update - {str(e)}", exc_info=True,
+                                  app_warning_msg=warn_msg)
```
vic-en, line 631:
```suggestion
"""
I'll propose a suggestion to simplify this. Since balance updates will be applied with inflight order snapshot, I suggest we make this function to only execute once per time. i.e no need to queue an extra call.
"""
```

vic-en (dismissed):

TheHolyRoger:
@vic-en I don't really agree with these changes tbh :) I tested this balance queue mechanism quite a bit with the gate.io connector, and to not queue a final balance update means you miss some order update events.

I also think keeping the network check separate is best, since ideally CoinZoom will add a `health` endpoint to their API in the future since it's less expensive for their servers

Feel free to add to this PR or close it, I was just including the useful bits I made for gate.io, I no longer have a CoinZoom to test with either!

hummingbot/connector/exchange/coinzoom/coinzoom_exchange.py
vic-en: I simplified this because I believe we don't need an extra queue since balance will eventually be calculated with snapshot.

Nullably:
@TheHolyRoger With your approach, If we, say, cancel 10 orders at the same time, the balance update (by REST API) will be fired 10 times, right? I don't think this is a good approach. With @vic-en s approach, some balance updates will get skipped (as if there is one currently going,) but we might not get the last update call which is the most important one. 

I think we are in a wrong direction here to fix the issue. We have the offline balance update mechanism that's supposed to address preciously this kind of issue. If that doesn't work for this connector or Gate.io we should look into fixing that.

On the NETWORK_CHECK vs SYMBOL, I don't mind either way.

Nullably (changes requested):
@TheHolyRoger With your approach, If we, say, cancel 10 orders at the same time, the balance update (by REST API) will be fired up to 10 times, right? I don't think this is a good approach. With @vic-en s approach, some balance updates will get skipped (as if there is one currently going,) but we might not get the last update call which is the most important one.

I think we are in a wrong direction here to fix the issue. We have the offline balance update mechanism that's supposed to address preciously this kind of issue. If that doesn't work for this connector or Gate.io we should look into fixing that.

On the NETWORK_CHECK vs SYMBOL, I don't mind either way.

TheHolyRoger:
> @TheHolyRoger With your approach, If we, say, cancel 10 orders at the same time, the balance update (by REST API) will be fired up to 10 times, right? I don't think this is a good approach. With @vic-en s approach, some balance updates will get skipped (as if there is one currently going,) but we might not get the last update call which is the most important one.
> 
> I think we are in a wrong direction here to fix the issue. We have the offline balance update mechanism that's supposed to address preciously this kind of issue. If that doesn't work for this connector or Gate.io we should look into fixing that.
> 
> On the NETWORK_CHECK vs SYMBOL, I don't mind either way.

Sorry @Nullably just seen this, no that's incorrect - with my approach the first and last updates will ALWAYS be fired, with any requests in between skipped unless none waiting for a response. In practice firing 10 order updates at once will only trigger 2 balance updates, maybe one in the middle if the API + bot are fast enough and no throttling is being hit - in which case it's a non-issue

TheHolyRoger:
@vic-en by the way sorry if my earlier response sounded blunt, didn't mean it to! Just mean I'm not overly attached to this PR... I thought I was helping but if I'm not then close it :)

Nullably (dismissed):

Nullably:
As mentioned during our stand-up, this connector is likely going to be a community maintaining one. So it is ok if it doesn't address the issue the same way our team does. 
@CoinAlpha/qa-team Please test this.

RC-13:
Re-tested the PR and it appears that the issue reported by @rapcmia still persists. In this scenario, there is about 5 seconds delay before the entire order gets canceled:
coinzoom

TheHolyRoger:
@RC-13 is that a new issue that only exists with this PR?

TheHolyRoger:
I've increased the throttler rates as they were a bit low

TheHolyRoger:
Not sure why this is failing to compile in the CI, compiles fine for me...

dennisocana:
Ran the tests again

RC-13:
> @RC-13 is that a new issue that only exists with this PR?

@TheHolyRoger Yes, we only see the issue in this PR. 

I've been observing the bot for 20 minutes now after the latest commit and it appears to have resolved the reported issue. No more delayed cancelation. LGTM!

RC-13 (approved):

PR #3568, opened 2021-06-11
Change in MockWebSocketServer to use daemon threads

aarmoa:

Change the implementation of MockWebSocketServer to use daemon threads, that will be automatically stopped once the application run finishes.


james-hummingbot (approved):
Just a note for my self, a daemon thread means it is non-blocking. I suppose the previous problem was that the non-daemon thread was blocking the parent one?

aarmoa:
That is correct @james-hummingbot. In the previous implementation the thread that was created was non daemon, and in the stop logic the thread was sent the message join(), to wait until it finished running. If the thread was not stoping for any reason, the test would never finish and would hang waiting for the thread.

PR #3579, opened 2021-06-14
feat/make twap a production strategy

aarmoa:

- Change the logic to work based on a total amount of base asset to operate and the order size, instead of total amount of base asset and number of orders.
- Update the logic to only use Limit orders
- Update the logic to update the quantity remaining when orders are cancelled, they expire or fail.
- Update configuration prompts
- Improve strategy status message

### DIFF
--- a/hummingbot/strategy/dev_4_twap/dev_4_twap.py
+++ b/hummingbot/strategy/dev_4_twap/dev_4_twap.py
@@ -115,7 +115,7 @@ class Dev4TwapTradeStrategy(StrategyPyBase):
         return [trade
                 for trade
                 in self.trades
-                if trade.trade_type == trade_type and trade.order_type == OrderType.LIMIT]
+                if trade.trade_type == trade_type.name and trade.order_type == OrderType.LIMIT]
 
     def format_status(self) -> str:
         lines: list = []
--- a/hummingbot/strategy/dev_4_twap/dev_4_twap.py
+++ b/hummingbot/strategy/dev_4_twap/dev_4_twap.py
@@ -32,23 +32,21 @@ class Dev4TwapTradeStrategy(StrategyPyBase):
 
     def __init__(self,
                  market_infos: List[MarketTradingPairTuple],
-                 order_type: str = "limit",
-                 order_price: Optional[float] = None,
-                 cancel_order_wait_time: Optional[float] = 60.0,
                  is_buy: bool = True,
-                 time_delay: float = 10.0,
-                 num_individual_orders: int = 1,
-                 order_amount: Decimal = Decimal("1.0"),
+                 target_asset_amount: Decimal = Decimal("1.0"),
+                 order_step_size: Decimal = Decimal("1.0"),
+                 order_price: Optional[Decimal] = None,
+                 order_delay_time: float = 10.0,
+                 cancel_order_wait_time: Optional[float] = 60.0,
                  status_report_interval: float = 900):
         """
         :param market_infos: list of market trading pairs
-        :param order_type: type of order to place
         :param order_price: price to place the order at
         :param cancel_order_wait_time: how long to wait before cancelling an order
         :param is_buy: if the order is to buy
-        :param time_delay: how long to wait between placing trades
+        :param order_delay_time: how long to wait between placing trades
         :param num_individual_orders: how many individual orders to split the order into
-        :param order_amount: qty of the order to place
+        :param target_asset_amount: qty of the order to place
         :param status_report_interval: how often to report network connection related warnings, if any
         """
 
@@ -63,13 +61,12 @@ class Dev4TwapTradeStrategy(StrategyPyBase):
         self._all_markets_ready = False
         self._place_orders = True
         self._status_report_interval = status_report_interval
-        self._time_delay = time_delay
-        self._num_individual_orders = num_individual_orders
-        self._quantity_remaining = order_amount
+        self._order_delay_time = order_delay_time
+        self._quantity_remaining = target_asset_amount
         self._time_to_cancel = {}
-        self._order_type = order_type
         self._is_buy = is_buy
-        self._order_amount = order_amount
+        self._target_asset_amount = target_asset_amount
+        self._order_step_size = order_step_size
         self._first_order = True
         self._previous_timestamp = 0
         self._last_timestamp = 0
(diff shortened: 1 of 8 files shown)


RC-13:
- When choosing a connector to use, some connectors that don't support the strategy are listed. I think it would be best to not have them on the list to avoid any confusion.
image

- The average filled order price doesn't seem to be working, after getting filled orders, the value remained at 0 USDT 
image
image

- Also, is this an expected behavior: there was an order placed by the bot prior to the order getting canceled, the strategy is stopped then resumed after about 5-10 sec. The pending order amount (150 DOGE) was retained and the previous order wasn't canceled. It only gets canceled after reaching the cancel order time.


https://user-images.githubusercontent.com/57189990/122169525-119d0080-ceb0-11eb-9f99-514bdd7cf856.mov

aarmoa:
> * When choosing a connector to use, some connectors that don't support the strategy are listed. I think it would be best to not have them on the list to avoid any confusion.
>   image
> * The average filled order price doesn't seem to be working, after getting filled orders, the value remained at 0 USDT
>   image
>   image
> * Also, is this an expected behavior: there was an order placed by the bot prior to the order getting canceled, the strategy is stopped then resumed after about 5-10 sec. The pending order amount (150 DOGE) was retained and the previous order wasn't canceled. It only gets canceled after reaching the cancel order time.
> 
>  Screen.Recording.2021-06-16.at.2.29.25.PM.mov

Hello @RC-13
I have commited a change that fixes the list of exchanges suggestions (to do that the parameter is now called connector). It also fixes the calculation of the average filled order price.

RC-13:
@aarmoa I'm getting the following error when importing an existing strategy and even using a newly created strategy which prevents us from confirming the fix for the initial reported issue:
image
```python
2021-06-18 07:58:58,131 - 6290 - hummingbot.client.hummingbot_application - ERROR - Unknown error during initialization.
  File "/Users/clint_g/Documents/Github-Repo/twap/hummingbot/strategy/dev_4_twap/start.py", line 19, in start
    exchange = dev_4_twap_config_map.get("exchange").value.lower()
AttributeError: 'NoneType' object has no attribute 'value'
2021-06-18 07:58:58,134 - 6290 - hummingbot.client.hummingbot_application - INFO - start command initiated.
```

aarmoa:
Sorry @RC-13, I forgot to rename the `exchange` parameter key from exchange to connector in start.py. I should have tested again manually the client after that rename, but I didn't and I missed the error.
It is fixed now. Would it be possible for you to try again?

RC-13:
- Confirming all reported issues have been fixed. 
- Re-posting question here: Under status, the `hang` label and `spread` is displayed, since we don't use this parameter, would it be best to hide them?
image

hummingbot/templates/conf_dev_4_twap_strategy_TEMPLATE.yml
```diff
@@ -8,14 +8,14 @@ strategy: null
 # twap strategy
 
 # Exchange and token parameters
-exchange: null
-market: null
+connector: null
```
CrimsonJacket, line 11: Is there a reason for changing this to `connector`? 
Also considering that there are changes to the configuration, It would be necessary to change the `template_version` number, so that the bot will prompt users for the new configurations.

CrimsonJacket (changes requested):
Considering that there are changes to the configuration, It would be necessary to change the `template_version` number, so that the bot will prompt users for the new configurations.

hummingbot/templates/conf_dev_4_twap_strategy_TEMPLATE.yml
aarmoa: I decided to change the configuration key from exchange to connector to make it easier for the user to identify it. I had to change the prompt text displayed to request that information, for the completer to only offer the spot connectors. The prompt now asks for the connector instead of the exchange.

I have changed the version number. Thank you for pointing that out.

RC-13 (approved):
LGTM

hummingbot/strategy/dev_4_twap/dev_4_twap.py
```diff
@@ -32,23 +38,21 @@ def logger(cls) -> HummingbotLogger:
...
+                 cancel_order_wait_time: Optional[float] = 60.0,
                  status_report_interval: float = 900):
         """
         :param market_infos: list of market trading pairs
-        :param order_type: type of order to place
         :param order_price: price to place the order at
         :param cancel_order_wait_time: how long to wait before cancelling an order
         :param is_buy: if the order is to buy
-        :param time_delay: how long to wait between placing trades
+        :param order_delay_time: how long to wait between placing trades
```
james-hummingbot, line 53: A little pedantic but came you make the order of param comments match the order of the params?

hummingbot/strategy/dev_4_twap/dev_4_twap.py
```diff
@@ -1,18 +1,24 @@
 from datetime import datetime
```
james-hummingbot, line 1: I would like to have module descriptions at the top of each file. Can you include a short description of what this module is trying to achieve. For example, what is twap, what kind of trading behavior does it have, etc.

hummingbot/strategy/dev_4_twap/dev_4_twap.py
```diff
@@ -107,11 +110,28 @@ def market_info_to_active_orders(self) -> Dict[MarketTradingPairTuple, List[Limi
     def place_orders(self):
         return self._place_orders
 
+    def filled_trades(self):
```
james-hummingbot, line 113: This is more a question for the base class, but `self.trades` is an array of already completed trades, can we clarify that in strategy_base.pyx?

hummingbot/templates/conf_dev_4_twap_strategy_TEMPLATE.yml
james-hummingbot: hmm, I remember @Nullably said ideally we would call these things exchanges, but if it is for consistency, I suppose let's stick to connector for now.

james-hummingbot (approved):
LGTM, just some small requests for documentation

aarmoa:
Thank you @james-hummingbot. I will add the documentation in the rename PR (#3607) to avoid triggering again the review requirement.

CrimsonJacket (approved):

PR #3580, opened 2021-06-15
feat/ add unit tests for TimeIterator

CrimsonJacket:

Includes unit test cases for `TimeIterator` and `PyTimeIterator`


aarmoa (approved):
In my opinion the PR is in good shape to be merged. The two new test classes look good.

As a possible improvement that does not affect the logic in this PR, the setUp method should always call super(), even if the superclass does nothing, in case in the future a new superclass is added in the middle (the same goes to tearDown, but that method is not used in these tests)

PR #3590, opened 2021-06-16
fix/ KuCoin balance pulling from both "Trading" and "Main" accounts

### LINKED ISSUE #2971: KuCoin - Total balance randomly pulling from "Trading" or "Main" account
Let's say you are running the bot against LUNA-USDT. If you type the "status" command over and over, you will sometimes see the Total USDT balance from your "Trading" account, but other times it will show you the value from the "Main" account.

This causes many issues with inventory skew, because it can sometimes show that you own 20% of the base currency, but then randomly see it as only 2% (depending how much USDT you have on either account). It can also mess up your strategy if you are using a script that relies on the bot knowing the total balance of Quote currency, as it it sporadically changes. For example, I'm trying to use the included sample script: inv_skew_using_spread_script.py
This value randomly changes:
self.quote_balance = self.all_total_balances[f"{self.pmm_market_info.exchange}"].get(self.quote_asset, self.quote_balance)

Since this is a trading bot, I don't think it should ever look at the Main account and should be limited to only the Trading account. Or at the very least, it should be consistent and SUM from both accounts.

CrimsonJacket:

Resolves #2971

It appears the KuCoin account balance updates have another attribute, `relationEvent`. This differentiates the Main and Trading account update.

With this, only balance from **Trading** account will be reflected within Hummingbot.

### DIFF
--- a/hummingbot/connector/exchange/kucoin/kucoin_exchange.pyx
+++ b/hummingbot/connector/exchange/kucoin/kucoin_exchange.pyx
@@ -281,12 +281,13 @@ cdef class KucoinExchange(ExchangeBase):
                         self.logger().debug(f"Event: {event_message}")
                         continue
                 elif event_type == "message" and event_topic == "/account/balance":
-                    currency = convert_asset_from_exchange(execution_data["currency"])
-                    available_balance = Decimal(execution_data["available"])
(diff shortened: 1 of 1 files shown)


Nullably (approved):
LGTM

rxlxrxsx (approved):

Issue #3596, opened 2021-06-17
Probit KR - Some unsupported markets are listed when choosing a market


RC-13:
**Describe the bug**
Recently probit made some changes and made a dedicated order book for Probit global and Probit Korea (https://support.probit.com/hc/en-us/articles/900007419443). With this recent change, our bot still recognizes some pairs that are currently for Probit global, choosing this pair will result in failing to submit / or fail to cancel the order (existing order prior to the change).
![image](https://user-images.githubusercontent.com/57189990/122340829-59399000-cf75-11eb-94ed-6cd2065a75f1.png)
```python
  2021-06-17 00:23:19,042 - 8043 - hummingbot.connector.exchange.probit.probit_exchange - NETWORK - Error submitting BUY LIMIT_MAKER order to ProBit for 0.00450 ETH-USDT 2397.25.
  Traceback (most recent call last):
    File "/home/ubuntu/probit/hummingbot/connector/exchange/probit/probit_exchange.py", line 341, in _api_request
      headers = await self._probit_auth.get_auth_headers(client)
    File "/home/ubuntu/probit/hummingbot/connector/exchange/probit/probit_auth.py", line 71, in get_auth_headers
      raise e
    File "/home/ubuntu/probit/hummingbot/connector/exchange/probit/probit_auth.py", line 60, in get_auth_headers
      token_resp = await resp.json()
    File "/home/ubuntu/miniconda3/envs/hummingbot/lib/python3.8/site-packages/aiohttp/client_reqrep.py", line 1026, in json
      raise ContentTypeError(
  aiohttp.client_exceptions.ContentTypeError: 0, message='Attempt to decode JSON with unexpected mimetype: text/html', url=URL('https://accounts.probit.kr/token')

  During handling of the above exception, another exception occurred:

  Traceback (most recent call last):
    File "/home/ubuntu/probit/hummingbot/connector/exchange/probit/probit_exchange.py", line 477, in _create_order
      order_result = await self._api_request(
    File "/home/ubuntu/probit/hummingbot/connector/exchange/probit/probit_exchange.py", line 357, in _api_request
      raise IOError(f"Error parsing data from {path_url}. Error: {str(e)}")
  OSError: Error parsing data from https://api.probit.kr/api/exchange/v1/new_order. Error: 0, message='Attempt to decode JSON with unexpected mimetype: text/html', url=URL('https://accounts.probit.kr/token')
```
```python
[...]

PR #3605, opened 2021-06-18
fix/ fix ProBit listing inactive markets

CrimsonJacket:

While investigating issue #3596 with ProBit KR still listing BTC/ETH/USDT trading pairs, I noticed this section of code, where there is a possibility that the ProBit connector will list inactive markets.

Below are the findings from initial investigation

1. When querying the following endpoint `https://api.probit.kr/api/exchange/v1/market`, it appears the markets only supported by Global is still being listed(i.e. `ETH-USDT`). See image attached.

2. Because these markets are still being considered as listed, the bot believes these are still active on the KR platform. Hence the bot still allows the user to configure a strategy with the de-listed markets.

3. The bot is unable to place the order by returning a {'market_id': 'not_found'} error, and although this is an expected behaviour, we felt that it would be not be an ideal experience from the users' perspective.

telegram-cloud-photo-size-5-6222103733303618733-y


RC-13 (approved):
LGTM confirmed that orders are placed, cancelled, and records filled orders.

aarmoa (approved):
I agree with the added condition to filter the trading pairs and do not return the closded ones

PR #3606, opened 2021-06-18
feat/ add unit tests for `Clock`

CrimsonJacket:

Adds unit tests for `Clock`. Includes tests both `REALTIME` and `BACKTEST` Clock modes.


test/hummingbot/core/test_clock.py
```diff
@@ -0,0 +1,127 @@
...
+class ClockUnitTest(unittest.TestCase):
+
+    backtest_start_timestamp: float = pd.Timestamp("2021-01-01", tz="UTC").timestamp()
+    backtest_end_timestamp: float = pd.Timestamp("2021-01-01 01:00:00", tz="UTC").timestamp()
+    tick_size: int = 1
+
+    @classmethod
+    def setUpClass(cls):
+        cls.ev_loop: asyncio.BaseEventLoop = asyncio.get_event_loop()
+        return super().setUpClass()
```
aarmoa, line 22: In the case of setUp methods, the call to super should be done as the first thing (because the superclass could be initializing things that are prerequisites for the initialization of the subclass).
It is correct to call super at the end in the tearDown method (since the cleanup is done in the reverse order from the setUp)

test/hummingbot/core/test_clock.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def setUp(self):
+        self.realtime_start_timestamp = time.time()
+        self.realtime_end_timestamp = self.realtime_start_timestamp + 2.0  #
+        self.clock_realtime = Clock(ClockMode.REALTIME, self.tick_size, self.realtime_start_timestamp, self.realtime_end_timestamp)
+        self.clock_backtest = Clock(ClockMode.BACKTEST, self.tick_size, self.backtest_start_timestamp, self.backtest_end_timestamp)
+
+        return super().setUp()
```
aarmoa, line 30: Same comment here as in setUpClass method

aarmoa (changes requested):

aarmoa (approved):
All changes look good to me

Issue #3271, opened 2021-04-23
Unit tests for Spot-Perpetual arbitrage strategy


Nullably:
Task for adding unit tests to Spot-Perpetual Arbitrage once the rework of the strategy's logic is complete.

Prerequisite : #3791

dennisocana:
Moving to backlog for further discussion. @vic-en please add your comments in this story.

Nullably:
As I listed bugs and code improvements on the #3679 tickets, I feel those should be addressed first before continuing with this ticket, or they could be done at the same time.

phbrgnomo:
Blocked by strategy rework mentioned on #3679.

dennisocana:
Adding the unit tests is part of the strategy rework story. Closing this issue.

PR #3624, opened 2021-06-23
Refactor/ Spot Perp Arb strategy with PerpetualTrading

Nullably:

- Update the strategy to comply with PerpetualTrading
- Since there is no unit test for Spot Perp Arb strategy yet, I created a starting test so that the PR passes the coverage requirement and this will serve and starting point for work on #3271

### DIFF
--- a/hummingbot/connector/perpetual_trading.py
+++ b/hummingbot/connector/perpetual_trading.py
@@ -42,8 +42,7 @@ class PerpetualTrading:
     def position_mode(self) -> PositionMode:
         return self._position_mode
 
-    @position_mode.setter
-    def position_mode(self, value: PositionMode):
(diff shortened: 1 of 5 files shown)


hummingbot/connector/perpetual_trading.py
```diff
@@ -63,8 +63,7 @@ def funding_payment_span(self) -> List[int]:
     def position_mode(self) -> PositionMode:
         return self._position_mode
 
-    @position_mode.setter
-    def position_mode(self, value: PositionMode):
+    def set_position_mode(self, value: PositionMode):
```
aarmoa, line 66: Why was the setter annotation removed? Do we want to prevent developers from using this as a property for setting the value?

aarmoa (approved):
The changes look good to me

PR #3628, opened 2021-06-24
Feat/ Add Kucoin and Ascend Ex rate oracle sources

Nullably:

- Add Kucoin and Ascend Ex rate oracle sources
- Update unit tests to mock API calls


aarmoa (approved):
Changes look good to me.
the functions get_*_prices look very similar. We should add a task in our backlog to refactor the common part out of the implementation for each provider.
The big if in get_prices(cls) is a call for action. The current implementation is not extensible. We should evaluate the possibility of creating different subclasses for each each of the providers (if the logic will differ a lot between them), or a list of already created instances for each of the valid providers (if the only thing that changes is configuration)

PtrckM (approved):
Tested and confirmed working by QA team

PR #3640, opened 2021-06-25
feat / Historical Volatility & rename Average Volatility

keithbaum:
Added Historical Volatility indicator.
Renamed Average volatility to Instant Volatility Indicator (changed in avellaneda strategy)

**Before submitting this PR, please make sure**:

- [x] Your code builds clean without any errors or warnings
- [x] You are using approved title ("feat/", "fix/", "docs/", "refactor/")

**A description of the changes proposed in the pull request**:

Solves #3629

PR #3646, opened 2021-06-28
add historical volatility and unit test for it as well

vic-en:


test/hummingbot/strategy/utils/test_historical_volatility.py
```diff
@@ -0,0 +1,19 @@
+import unittest
+from hummingbot.strategy.__utils__.trailing_indicators.historical_volatility import HistoricalVolatilityIndicator
+
+
+class HistoricalVolatilityIndicatorUnitTest(unittest.TestCase):
+
+    def test_historical_volatility(self):
```
CrimsonJacket, line 7: I think this unit test is not isolated enough. It appears to look more like an test for the entire class.
Unit tests should cover the smallest piece of code, i.e. `test_add_sample()`, `test_indicator_calculation()`, `test_processing_calculation()`.

hummingbot/strategy/__utils__/trailing_indicators/historical_volatility.py
```diff
@@ -0,0 +1,26 @@
+from .base_trailing_indicator import BaseTrailingIndicator
```
CrimsonJacket, line 1: I am not sure if the team has set a coding standard, but I personally try not to use relative imports.

CrimsonJacket (changes requested):
I think this unit test is not isolated enough. It appears to look more like a test for the entire class. Unit tests should cover the smallest piece of code, i.e. `test_add_sample()`, `test_indicator_calculation()`, `test_processing_calculation()`.

vic-en:
I added more assert checks.  Thanks.

CrimsonJacket (approved):
LGTM ⚡

aarmoa (changes requested):
As discussed during today's stand up meeting, this calculations in this PR might not be the correct ones.
We should evaluate first #3640

hummingbot/strategy/__utils__/trailing_indicators/historical_volatility.py
```diff
@@ -0,0 +1,26 @@
...
+
+s_decimal_0 = Decimal("0")
+
+
+class HistoricalVolatilityIndicator(BaseTrailingIndicator):
+    def __init__(self, sampling_length: int = 30, processing_length: int = 15):
+        super().__init__(sampling_length, processing_length)
+        self._previous_sample = s_decimal_0
+
+    def add_sample(self, value: float):
```
phbrgnomo, line 13: add_sample should be handled on the strategy file, not inside the indicator calculation, as it happens with the avellaneda strategy and the `AverageVolatilityIndicator`

We want to be possible to use different inputs to populate the price buffer.

vic-en: this simply override the add_sample from the base class.

PR #3652, opened 2021-06-29
fix/avellaneda hanging orders and market balance

### LINKED ISSUE #3610: Avellaneda_MM - Hanging orders not recreated when max order age is reached
**Describe the bug**
There are instances that when the hanging order hit the max_order_age, it was never recreated. When max order age is reached and the hanging orders were canceled, two normal orders were created. Additional info from @phbrgnomo:
It seems that the bot is recognizing these two new normal orders created right after the hanging order cancelation, as both normal AND hanging:

Normal because they use the price/order size as a normal order
Hanging because these orders are not canceled after order_refresh_time (which I set to 300s). They are canceled only after the max_order_age had passed. Only after the max_order_age pass for these conflicting orders, the hanging orders are recreated, along with the two other normal orders.

Screen Shot 2021-06-20 at 3 44 17 PM

**Steps To Reproduce**
1. Run a avellaneda_market_making bot with hanging order enabled
2. For testing purposes you can set the max order age to a lower value

**Release version**
dev-0.41.0

**Attachments**
[ave_hanging.zip](https://github.com/CoinAlpha/hummingbot/files/6681902/ave_hanging.zip)

aarmoa:

Added 2 test cases that intend to reproduce the issues in #3610.
I made changes in the Avellaneda strategy, specifically in the interaction with the hanging orders tracker during the orders cancellation cycle and the orders creation cycle.
I have also updated the logic that calculates the available balance for the budget constraint calculation.

Fix #3610


RC-13 (approved):
LGTM, hanging orders gets re-created every after the refresh cycle if it got cancelled by max order age.

hummingbot/strategy/hanging_orders_tracker.py
```diff
@@ -171,10 +171,12 @@ def update_strategy_orders_with_equivalent_orders(self):
         equivalent_orders = self.equivalent_orders
         orders_to_create = equivalent_orders.union(self.orders_to_be_created).\
             difference(self.strategy_current_hanging_orders)
-        orders_to_cancel = self.strategy_current_hanging_orders.difference(equivalent_orders).\
-            difference(orders_to_create)
+        orders_to_cancel = self.strategy_current_hanging_orders.difference(equivalent_orders)
```
CrimsonJacket, line 174: @aarmoa Will this cover the case when we have a single accumulated hanging order?

CrimsonJacket (commented):
If I recall correctly, from our numerous discussions before, we determined that it was necessary to handle hanging orders differently for the different aggregation types.

I think it was brought up before that it might be better to separate the calling of the hanging order tracking functions into its own thread loop. This is to isolate, as much as possible, the logic of tracking hanging orders away from the strategy. Yes it will still need to place/cancel the orders using the strategy's `buy/sell_with_specific_market()` methods but I feel that by doing so, we can truly modularise the `HangingOrderTracker` class.

aarmoa:
> If I recall correctly, from our numerous discussions before, we determined that it was necessary to handle hanging orders differently for the different aggregation types.
> 
> I think it was brought up before that it might be better to separate the calling of the hanging order tracking functions into its own thread loop. This is to isolate, as much as possible, the logic of tracking hanging orders away from the strategy. Yes it will still need to place/cancel the orders using the strategy's `buy/sell_with_specific_market()` methods but I feel that by doing so, we can truly modularise the `HangingOrderTracker` class.

Yes, I totally agree with you. I was planning on doing so in the context of this task **Apply max order age refresh fixes made to PMM into Avellaneda** #3660, and leave this PR only for fixing the reported issue.

PR #3581, opened 2021-06-15
Fix/ Max order age on hanging orders

Nullably:
**Before submitting this PR, please make sure**:

- [x] Your code builds clean without any errors or warnings
- [x] You are using approved title ("feat/", "fix/", "docs/", "refactor/")

**A description of the changes proposed in the pull request**:
- Refactor order_age fn to a utils file
- Max order age on active orders are recreated at refresh interval
- Max order age on hanging orders are first cancelled, then wait for cancellation event before recreate the hanging orders

PR #3705, opened 2021-07-08
Fix/apply max order age refresh fixes from pmm to avellaneda

### LINKED ISSUE #3660: Apply max order age refresh fixes made to PMM into Avellaneda
**Feature Suggestion**
Apply the fixes made by Jack to PMM regarding max order age refresh into Avellaneda strategy.

https://github.com/CoinAlpha/hummingbot/pull/3581

aarmoa:

Apply improvements made to Pure Market Making strategy in #3581 to Avellaneda strategy. Changes are related to the hanging orders refresh when they reach the max_order_age configured for the strategy.
I took the change to directly connect the strategy to the order cancellation and order completion events, not requiring any more to forward them to the hanging orders tracker from the strategy (and simplifying this way the inclusion of the tracker in other strategies).

**Tests performed by developer**
Added unit tests for hanging order refresh escenario, both with grouping and non grouping strategies configured in the tracker.
Added unit tests for the functionality when a buy hanging order is fully filled.
Added unit tests  for the functionality when a sell hanging order is fully filled.
Added unit tests to verify the hanging orders are removed when they are far from the current price.

**Hints for QA team**
Please make sure to test the hanging orders are refreshed when their life time reaches the max_order_age period. When refreshing a hanging order the new hanging order should be created as soon as the original is canceled and the cancelation is confirmed by the exchange (the strategy no longer waits until the next order creation period to recreate the hanging orders).
Also please take a look at the functionality that removes hanging orders when they are far from price. This could be done by configuring a small percentage to consider an order far from price.

Closes #3660


RC-13 (dismissed):
Tested and confirm working by QA, LGTM

hummingbot/strategy/hanging_orders_tracker.py
```diff
@@ -98,31 +225,25 @@ def remove_all_sells(self):
...
+                if hanging_order.creation_timestamp
+                else -1)
+
     def renew_hanging_orders_past_max_order_age(self):
-        max_age = getattr(self.strategy, "max_order_age", None)
         to_be_cancelled: Set[HangingOrder] = set()
-        to_be_created: Set[HangingOrder] = set()
-        if max_age:
+        max_order_age = getattr(self.strategy, "max_order_age", lambda: None)
+        if max_order_age:
```
keithbaum, line 239: In case max_order_age is lambda: None this condition will also be True

keithbaum (approved):
Thanks for the improvements @aarmoa 💯
Just left one small comment to be reviewed

hummingbot/strategy/hanging_orders_tracker.py
aarmoa: Good catch Nico, thank you for that.

aarmoa:
I fixed the problem detected by @keithbaum. The change will not require retesting since that logic is only used with strategies that do not have the concept of max_order_age.

aarmoa:
I have added unit tests to verify the logic that removes hanging orders far from current price. I detected the issue reviewing errors in logs that @keithbaum sent me.
I have also added more tests cases to check the tracker can correctly specify if a particular limit order is currently a hanging order or is in the process of converting into one.

rxlxrxsx (approved):
Re-tested and confirmed working by QA.

hummingbot/strategy/strategy_base.pyx
```diff
@@ -138,6 +138,9 @@ cdef class StrategyBase(TimeIterator):
     def order_tracker(self) -> OrderTracker:
         return self._sb_order_tracker
 
+    def limit_order_from_order_tracker(self, order_id: str):
+        return self.order_tracker.get
+
```
CrimsonJacket, line 143: Not quite sure what this is for?

.github/workflows/workflow.yml
```diff
@@ -2,9 +2,9 @@ name: ci
 
 on:
   push:
-    branches: [ master, development, 'refactor/unit_tests**' ]
+    branches: [master, development, 'refactor/unit_tests**', 'main/**']
   pull_request:
-    branches: [ master, development, refactor/unit_tests ]
+    branches: [master, development, 'refactor/unit_tests**', 'main/**']
```
CrimsonJacket, line 7: I dont believe this is part of this PR.

CrimsonJacket (changes requested):
Made some comments on certain changes to the `workflow.yml` and `StrategyBase` files. 
But other than that, I like that now the managing of the hanging orders are entirely isolated within the `HangingOrderTracker`'s `process_tick()` method. Makes it easier for other strategies to utilize the `HangingOrderTracker`

.github/workflows/workflow.yml
aarmoa: I am not sure why this is being shown as a difference coming from this PR. This change is already part of `development` and it was applied there using a different branch and a different PR.

hummingbot/strategy/strategy_base.pyx
aarmoa: This was a change that should have never been included in the branch. Thank you @CrimsonJacket, good catch. I have removed it.

aarmoa:
The last commit does not affect any functionality (just removes a function that was not being used). A new validation from QA is not required.

.github/workflows/workflow.yml
aarmoa: Since the contents in `workflow.yml` are the same we currently have in `development` I suggest we ignore that being shown by GitHub as changes.

PR #3794, opened 2021-07-26
Feat/ndax config parameters

### LINKED ISSUE #3793: Add NDAX parameters and configurations to the client
Add required parameters and configuration entries to make NDAX connector available and usable from the client.

### Definition of Done

- NDAX connector is included in connectors table
- NDAX connector can be created and all its parameters specified: user id, user name, API key, API secret
- The balance from the NDAX account can be retrieved

aarmoa:

Added configuration parameters to establish the connection with NDAX exchange.
Also added a fix to the initialization of the account ID (it is required for the client to get the account balance even before the connector is started)

I tested the changes manually by configuring the connection with NDAX and getting the balance without issues.

Closes #3793


hummingbot/connector/exchange/ndax/ndax_utils.py
```diff
@@ -9,5 +10,65 @@
...
+# USE_ETHEREUM_WALLET not required because default value is false
+# FEE_TYPE not required because default value is Percentage
+# FEE_TOKEN not required because the fee is not flat
+
+
 def convert_to_exchange_trading_pair(hb_trading_pair: str) -> str:
     return hb_trading_pair.replace("-", "")
+
+
+KEYS = {
```
CrimsonJacket, line 22: Should we include an optional config for account id?

hummingbot/connector/exchange/probit/probit_utils.py
```diff
@@ -75,7 +75,7 @@ def convert_diff_message_to_order_book_row(message: OrderBookMessage) -> Tuple[L
 KEYS = {
     "probit_api_key":
         ConfigVar(key="probit_api_key",
-                  prompt="Enter your ProBit Client ID >>> ",
+                  prompt="Enter your ProBit API key >>> ",
```
CrimsonJacket, line 78: For this, we decided to use Client ID because that's what their UI suggest it was; a ClientID-Secret key pair

setup.py
```diff
@@ -55,27 +54,29 @@ def main():
...
+        "hummingbot.connector.exchange.ndax"
         "hummingbot.connector.exchange.okex",
-        "hummingbot.connector.exchange.liquid",
-        "hummingbot.connector.exchange.dolomite",
-        "hummingbot.connector.exchange.eterbase",
-        "hummingbot.connector.exchange.beaxy",
-        "hummingbot.connector.exchange.hitbtc",
-        "hummingbot.connector.exchange.gate_io",
-        "hummingbot.connector.exchange.k2",
+        "hummingbot.connector.exchange.probit"
```
CrimsonJacket, line 78: You're missing a `,` at the end of the probit entry

CrimsonJacket (dismissed):
Left some comments. Minor changes requested

CrimsonJacket (changes requested):
Left some comments. Minor changes requested

hummingbot/connector/exchange/ndax/ndax_utils.py
aarmoa: Yes. In fact we need to ask for the name of the account the user wants to use (they could have more than one account associated with their user account)

hummingbot/connector/exchange/probit/probit_utils.py
aarmoa: Thank you for the clarification Daniel. I will revert the change.

hummingbot/connector/exchange/ndax/ndax_api_order_book_data_source.py
```diff
@@ -222,40 +227,37 @@ async def listen_for_order_book_diffs(self, ev_loop: asyncio.BaseEventLoop, outp
...
+                                "instrument_id": msg_product_code,
+                            }
+
+                            if msg_event == CONSTANTS.WS_ORDER_BOOK_CHANNEL:
+                                snapshot_msg: OrderBookMessage = NdaxOrderBook.snapshot_message_from_exchange(msg=content,
+                                                                                                              timestamp=msg_timestamp,
+                                                                                                              metadata=metadata)
+                                output.put_nowait(snapshot_msg)
+                            elif msg_event == CONSTANTS.WS_ORDER_BOOK_L2_UPDATE_EVENT:
+                                diff_msg: OrderBookMessage = NdaxOrderBook.diff_message_from_exchange(msg=content,
```
CrimsonJacket, line 257: The reason why I decided to include the clause to `continue` when (1) receiving an unrecognised msg event and (2) receiving updates for other trading pairs is to make it st it wont be an indentation hell like this. But nonetheless I feel both achieve the same thing,

CrimsonJacket (changes requested):
I am noticing some changes pertaining to pure market making and I dont think its intended. I amd also seeing some duplicate additions to the `Exchange` class. (i.e. property functions)

aarmoa:
Thank you for the feedback @CrimsonJacket. It seems GitHub did something odd when comparing the branch with the epic base, because I did not change anything related to PMM. I have forced the refresh of the base and now the changes shown make more sense. Would you mind reviewing them again?

hummingbot/connector/exchange/ndax/ndax_api_order_book_data_source.py
aarmoa: I agree with you regarding the indentation. But I think it would be better to solve that problem by extracting code to methods instead of using the continue statement.

hummingbot/connector/exchange/ndax/ndax_exchange.py
```diff
@@ -197,22 +198,25 @@ async def _get_account_id(self) -> int:
...
-              The assumption here is that the FIRST entry in the list is the accountId the user intends to use
-        """
-        return resp[0]
+        account_info = next((account_info for account_info in resp
+                             if account_info.get("AccountName") == self._auth.account_name),
+                            None)
+        if account_info is None:
+            self.logger().error(f"There is no account named {self._auth.account_name} "
+                                f"associated with the current NDAX user")
+            raise ValueError(f"NDAX account named {self._auth.account_name} does not exist")
```
CrimsonJacket, line 217: Not sure if this `ValueError` will be raised. I figure we dont need to raise an error here. Since in `ready`, there are checks to ensure that the `account_id` is initialized. If it is not initialized, it will not start, which is an expected behaviour.

CrimsonJacket (changes requested):
Just a minor change requested. But other than that, everything looks good.

CrimsonJacket (approved):
Changes LGTM

PR #3832, opened 2021-07-30
fix/ fix outstanding issues with NDAX connector

CrimsonJacket:

Issues addressed
1. Auth Details(uid, account name) not loaded from file on start up.
2. OrderBook snapshots not using the right domains
3. `order_book` command failing due to the missing `@property` `order_books`

### DIFF
--- a/hummingbot/connector/exchange/ndax/ndax_exchange.py
+++ b/hummingbot/connector/exchange/ndax/ndax_exchange.py
@@ -774,7 +774,7 @@ class NdaxExchange(ExchangeBase):
                 "UserId": self._auth.uid,
                 "InstrumentId": trading_pair_ids[trading_pair],
                 "StartTimestamp": min_ts,
-                "EndTimestamp": int(time.time() * 1e3),
+                "EndTimestamp": int(time.time()),
             }
             trade_history_tasks.append(
                 asyncio.create_task(self._api_request(method="GET",
--- a/hummingbot/connector/exchange/ndax/ndax_exchange.py
+++ b/hummingbot/connector/exchange/ndax/ndax_exchange.py
@@ -150,6 +150,10 @@ class NdaxExchange(ExchangeBase):
         return all(self.status_dict.values())
 
     @property
+    def order_books(self) -> Dict[str, OrderBook]:
+        self._order_book_tracker.order_books
+
+    @property
     def limit_orders(self) -> List[LimitOrder]:
         return [
             in_flight_order.to_limit_order()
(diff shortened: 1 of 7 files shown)


hummingbot/connector/exchange/ndax/ndax_websocket_adaptor.py
```diff
@@ -81,17 +81,20 @@ async def recv(self):
...
-            )
+                try:
+                    msg: str = await asyncio.wait_for(self.recv(), timeout=self.MESSAGE_TIMEOUT)
+                    yield msg
+                except asyncio.TimeoutError:
+                    await asyncio.wait_for(
+                        self.send_request(CONSTANTS.WS_PING_REQUEST, payload={}),
+                        timeout=self.PING_TIMEOUT
+                    )
+                except Exception as e:
```
aarmoa, line 92: No need to capture the exception if we will just pass it (or raise it again)

hummingbot/connector/exchange/ndax/ndax_exchange.py
```diff
@@ -779,7 +783,7 @@ async def _update_order_status(self):
         # Initial parsing of responses. Joining all the responses
         parsed_history_resps: List[Dict[str, Any]] = []
         for resp in raw_responses:
-            if not isinstance(resp, Exception):
+            if not isinstance(resp, Exception) and resp["ClientOrderId"] in self.in_flight_orders:
```
aarmoa, line 786: It might be a good idea to separate the validation of "ClientOrderId" being in our collection of in flight orders. If we receive and update for an order we are not tracking it could be confusing if we log that as an error.

aarmoa (commented):
I have added two comments. @CrimsonJacket let me know your opinion about that and I will review the PR again afterwards.

hummingbot/connector/exchange/ndax/ndax_websocket_adaptor.py
CrimsonJacket: Yea I overlooked this. Yea my intention was to capture this error from in the `NdaxAPIOrderBookDataSource` class, where `iter_messages()` is called.

hummingbot/connector/exchange/ndax/ndax_exchange.py
CrimsonJacket: Yea I agree. Alright will make the necessary changes.

CrimsonJacket: I realise that it will simply be ignored in `_process_order_event_message`. So I will be removing this condition.

hummingbot/connector/exchange/ndax/ndax_websocket_adaptor.py
```diff
@@ -93,6 +93,9 @@ async def iter_messages(self) -> AsyncIterable[str]:
                     raise e
         except websockets.exceptions.ConnectionClosed:
             return
+        except Exception as e:
+            # Raises whatever Exception to the calling functions
+            raise e
```
aarmoa, line 98: No reason to catch the exception if it will be only raised again

aarmoa (approved):
I agree with the changes

PR #3834, opened 2021-07-30
Fix/change websocket iterator in user stream

aarmoa:

Change in the implementation of the web socket received messages iterator in the NDAX user stream data source. Now it reuses the functionality already present in the NDAX web socket, that also takes care of sending a ping request regularly to keep the connection alive.
I also added a fix to the issue that was causing error in the cancel_all functionality from the connector.
Includes the fix to an error caused by the account_id being referenced before it was initialized by the start_network logic (it is required for the cancel_all logic kicked off by the start command)


CrimsonJacket (approved):
Changes LGTM

PR #3843, opened 2021-08-03
Fix/ndax order book diff entry removal

aarmoa:

This PR includes two fixes:
- Ensure the updates received regarding order book events of type DELETE are processed as deletes, even if their amount is not zero.
- Ensure requesting order book data from the exchange does not fail when the response does not include events.

Added unit tests to cover the new scenarios.


CrimsonJacket (approved):
Changes LGTM. Will be merging to epic branch

PR #3846, opened 2021-08-03
Fix/ NDAX GetTradesHistory not filtering by StartTimestamp

CrimsonJacket:

Addresses an issue with the `GetTradesHistory` endpoint where the filters are not applied appropriately.

For Developers
- Update the `GET` request to a `POST` request and include the relevant parameters as body parameters
- `StartTimestamp` and `EndTimestamp` parameters are in milliseconds (seconds * 1000)
- Update the relevant unit tests

For QA
- Nothing much changes, ensure that status polling is working as intended.

### DIFF
--- a/hummingbot/connector/exchange/ndax/ndax_exchange.py
+++ b/hummingbot/connector/exchange/ndax/ndax_exchange.py
@@ -776,7 +776,6 @@ class NdaxExchange(ExchangeBase):
                 "UserId": self._auth.uid,
                 "InstrumentId": trading_pair_ids[trading_pair],
                 "StartTimestamp": min_ts,
-                "EndTimestamp": int(time.time() * 1e3),
             }
             trade_history_tasks.append(
                 asyncio.create_task(self._api_request(method="POST",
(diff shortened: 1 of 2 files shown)


hummingbot/connector/exchange/ndax/ndax_exchange.py
```diff
@@ -763,25 +763,25 @@ async def _update_order_status(self):
...
         for trading_pair in self._trading_pairs:
-            query_params = {
+            body_params = {
                 "OMSId": 1,
                 "AccountId": await self.initialized_account_id(),
                 "UserId": self._auth.uid,
                 "InstrumentId": trading_pair_ids[trading_pair],
                 "StartTimestamp": min_ts,
-                "EndTimestamp": int(time.time()),
+                "EndTimestamp": int(time.time() * 1e3),
```
aarmoa, line 779: Since we want all the trades history up to now, and there could be differences between the client time and the exchange time, I think it would be better to not include the EndTimestamp parameter (ie: let the exchange return all the latest information)

aarmoa (approved):
I agree with the changes

hummingbot/connector/exchange/ndax/ndax_exchange.py
CrimsonJacket: Yes I agree that `EndTimestamp` is essentially redundant and might not be an accurate representation of the orders we want to capture due to the difference in local and server time. I will make the necessary changes.

PR #3855, opened 2021-08-04
Fix/increase ndax order book snapshot frequency

aarmoa:

I have changed the frequency of full order book snapshot captures by the client. It was 1 hour and it is now 5 minutes. We need to do this because NDAX is not sending reliable updates for the order book entries.

I executed a PMM strategy connected to the production NDAX account. I validated that any difference is solved every 5 minutes.


CrimsonJacket (approved):
Changes LGTM

rxlxrxsx:
Tested and confirmed that there are still instances that the order book is not in sync with the exchange. When this happens, it recovers within 5 minutes.

PR #3896, opened 2021-08-10
feat/allow_relative_and_abs_path_in_script

vic-en:
… abs and relative paths


installation/docker-commands/create-gateway.sh
```diff
@@ -26,11 +26,15 @@ fi
 
 # Ask the user for the hummingobt data folder location
 prompt_hummingbot_data_path () {
-read -p "   Enter the full location path where your Hummingbot files are located (example: /Users/hbot/hummingbot_files) >>> " FOLDER
+read -p "   Enter the location where your Hummingbot files are located (example: /Users/hbot/hummingbot_files) >>> " FOLDER
 if [ "$FOLDER" == "" ]
 then
   prompt_hummingbot_data_path
 else
+  if ! [[ "$FOLDER" == *"$HOME"* ]]; then
```
martinkou, line 34: The correct method to detect whether a path is absolute or relative is to look for a leading "/". If there's a leading "/", then it's absolute. Otherwise, it's relative.

installation/docker-commands/create.sh
```diff
@@ -30,15 +30,17 @@ fi
 read -p "   Enter a folder name where your Hummingbot files will be saved (default = \"$DEFAULT_FOLDER\") >>> " FOLDER
 if [ "$FOLDER" == "" ]
 then
-  FOLDER=$DEFAULT_FOLDER
+  FOLDER=$PWD/$DEFAULT_FOLDER
+elif ! [[ "$FOLDER" == *"$HOME"* ]]; then
```
martinkou, line 34: Same. Use the leading slash to check, not the home directory.

installation/docker-commands/update.sh
```diff
@@ -59,7 +59,9 @@ prompt_folder () {
    read -p "   Enter the destination folder for $instance (default = \"$DEFAULT_FOLDER\") >>> " FOLDER
    if [ "$FOLDER" == "" ]
    then
-     FOLDER=$DEFAULT_FOLDER
+     FOLDER=$PWD/$DEFAULT_FOLDER
+   elif ! [[ "$FOLDER" == *"$HOME"* ]]; then
```
martinkou, line 63: ditto.

martinkou (changes requested):

installation/docker-commands/create-gateway.sh
martinkou: Another thing is, since this detection and conversion function is common across a few scripts - it'll be good to refactor it to a bash function in a common utilities library file - and source that file in all the scripts that needs it.

martinkou (approved):

PtrckM (approved):
Tested and confirmed working.

PR #3922, opened 2021-08-12
refactor/ Consolidate APIThrottler

CrimsonJacket:

Consolidates APIThrottler to a single `AsyncThrottler` class.
Refactor CryptoCom to use `AsyncThrottler`

Refactor CryptoCom connector to use `AsyncThrottler`
Removed unused throttler classes

**Tips for QA testing**:
Test that API throttling is working on CryptoCom connector


hummingbot/core/api_throttler/async_throttler_base.py
```diff
@@ -0,0 +1,73 @@
...
+        :param retry_interval: Time between every capacity check.
+        :param safety_margin: Percentage of limit to be added as a safety margin when calculating capacity to ensure calls are within the limit.
+        """
+
+        # Rate Limit Definitions
+        self._rate_limits: List[RateLimit] = copy.deepcopy(rate_limits)
+        limits_pct_conf: Optional[Decimal] = global_config_map["rate_limits_share_pct"].value
+
+        # If configured, users can define the percentage of rate limits to allocate to the throttler.
+        self.limits_pct: Optional[Decimal] = Decimal("1") if limits_pct_conf is None else limits_pct_conf // Decimal("100")
```
aarmoa, line 39: I am not sure we need to consider the percentage an integer value (it should be possible for the user to specify a percentage of 33.33)

hummingbot/core/api_throttler/async_request_context_base.py
```diff
@@ -0,0 +1,88 @@
...
+    @abstractmethod
+    def within_capacity(self) -> bool:
+        raise NotImplementedError
+
+    async def acquire(self):
+        while True:
+            self.flush()
+
+            if self.within_capacity():
+                break
```
aarmoa, line 76: Instead of using "break" we can use self.within_capacity() in the while condition

aarmoa (commented):
I am not sure I agree with the concept of the linked limits. This implementation might not be enough to handle grouped limits like required for example for Bybit.
There is no need to have abstract classes for the Throttler and the Context if there will be only one concrete class for each of them

hummingbot/core/api_throttler/async_throttler_base.py
CrimsonJacket: Yea I agree with you we should not round result. Will resolve this

aarmoa (dismissed):
Changes look good to me

PtrckM:
- intentionally set to lower `rate_limits_share_pct 10` to hit rate limit
  - unable to stop / exit / ctrl + c (twice) / restart bot
  - throws error
     - force exit by killing the PID

Screen Shot 2021-08-14 at 12 03 38 PM

- config and logs
[crypto_com.zip](https://github.com/CoinAlpha/hummingbot/files/6985871/crypto_com.zip)

rxlxrxsx:
Just to add to the issue that @PtrckM encountered, it seems that the issue will happen as long as you set the `rate_limits_share_pct` to 30 or below and will immediately happen after starting the bot either by creating a new strategy or importing a previous one as long as crypto_com is the connector used.

Screen Shot 2021-08-15 at 1 51 49 PM
Screen Shot 2021-08-15 at 2 09 22 PM

```
2021-08-15 06:08:00,001 - 11889 - hummingbot.core.api_throttler.async_request_context_base - INFO - API rate limit on private/get-account-summary (0.9 calls per 0.1s) has almost reached. Number of calls is 0 in the last 0.1 seconds
```

[throttler.zip](https://github.com/CoinAlpha/hummingbot/files/6987611/throttler.zip)

dennisocana:
For re-testing and verification by @CoinAlpha/qa-team

aarmoa (approved):
The fix looks good to me

RC-13 (approved):
Confirming the reported issue has been fixed. LGTM

PR #3959, opened 2021-08-19
refactor/ Integrates the `AsyncThrottler` into the `NdaxExchange`.

petioptrv:

- It adds the `AsyncThrottler` rate-limiting lock to the HTTP and WS API requests.

- The existing tests pass.

**Tips for QA testing**:
Test that the connector does not throw an error on api-rate limit (maybe by launching a PMM with 300 levels on each side to reach the 600/minute limit).


hummingbot/connector/exchange/ndax/ndax_constants.py
```diff
@@ -31,8 +34,132 @@
...
 # WebSocket Message Events
 WS_ORDER_BOOK_L2_UPDATE_EVENT = "Level2UpdateEvent"
 
 API_LIMIT_REACHED_ERROR_MESSAGE = "TOO MANY REQUESTS"
+
+MINUTE = 60
+HTTP_ENDPOINTS_LIMIT = "AllHTTP"
+HTTP_LIMIT = 600
+WS_AUTH_LIMIT = "AllWsAuth"
+WS_ENDPOINTS_LIMIT = "AllWs"
```
CrimsonJacket, line 48: Would prefer these constants to end with `LIMIT_ID`.

hummingbot/connector/exchange/ndax/ndax_api_order_book_data_source.py
```diff
@@ -48,22 +58,27 @@ async def init_trading_pair_ids(cls, domain: Optional[str] = None):
...
-                if response.status == 200:
-                    resp_json: Dict[str, Any] = await response.json()
-
-                    results = {
-                        f"{instrument['Product1Symbol']}-{instrument['Product2Symbol']}": int(
-                            instrument["InstrumentId"])
-                        for instrument in resp_json
-                        if instrument["SessionStatus"] == "Running"
-                    }
+            throttler = throttler or cls._get_throttler_instance()
```
CrimsonJacket, line 61: You can simply include the default value of throttler here instead of using the following.
`throttler = throttler or cls._get_throttler_instance()`

hummingbot/connector/exchange/ndax/ndax_user_stream_tracker.py
```diff
@@ -39,7 +43,9 @@ def data_source(self) -> UserStreamTrackerDataSource:
         :return: UserStreamTrackerDataSource
         """
         if not self._data_source:
-            self._data_source = NdaxAPIUserStreamDataSource(auth_assistant=self._auth_assistant, domain=self._domain)
+            self._data_source = NdaxAPIUserStreamDataSource(
+                self._throttler, auth_assistant=self._auth_assistant, domain=self._domain
```
CrimsonJacket, line 47: Just a code styling issue. Do include the expected parameter since it is available.
i.e. `ds = NdaxAPIUserStreamDataSource(throttler=self._throttler, ...)`

CrimsonJacket (changes requested):
Left some comments wrt code styling. Seems like there are quite a bit of changes to include when we are integrating the `AsyncThrottler` especially when dealing with class methods and websocket client.

hummingbot/connector/exchange/ndax/ndax_api_order_book_data_source.py
aarmoa: We have to be careful when using defaults values for parameters. Since those values are calculated when the module is loaded (and not every time the function is executed), if the default value is a mutable object, all executions of the function will be sharing the same mutable object (with all the problems that situation can produce)

petioptrv: I agree with @aarmoa on the non-literal defaults in the signature, but I think @CrimsonJacket is suggesting we move the creation of the throttler to this line instead of going through a class method, which would make things a bit cleaner. My thinking was that, should we ever need to change something with how the throttler is initialized, this approach will require us changing only one place in the code

hummingbot/connector/exchange/ndax/ndax_constants.py
petioptrv: Yes, good point. Changed

hummingbot/connector/exchange/ndax/ndax_user_stream_tracker.py
petioptrv: Changed it to match the style of the existing code. However, while I am very much pro passing arguments by key-value when they add additional information (e.g. `order = self.place_buy_order(price=30)`, I think in cases like these, it adds verbosity without necessarily increasing the information

CrimsonJacket (dismissed):
Changes LGTM

dennisocana:
Moving to QA for testing before merge

RC-13:
Re-tested the PR and I keep experiencing an issue with cancelling orders whenever the API rate usage is shown in the logs. There are instances that the bot recover after few seconds to 1 minute however there are scenarios as well that it doesn't. This can be triggered easily by changing the rate_limit_pct to 1% (Tested thru cloud and the local machine).
failing to cancel
Using the same configuration:
image
I cross-check the development branch and the behaviour doesn't happen. The bot is running for almost 15 min without issue
dev

petioptrv:
@RC-13 thank you for testing. I pushed a fix.

@CrimsonJacket the last two commits address an issue where the order-creation API request gets rate-limited and a cancellation is requested before the call goes through. Please review the new changes carefully as they have an effect on the order creation/cancellation flow.

PtrckM (approved):
Test current changes confirmed working, re-approving in behalf of QA.

CrimsonJacket (approved):
Changes LGTM.

PR #3960, opened 2021-08-19
fix/ PMM Invalid price proposal filtering

petioptrv:

- Adds a check in PMM for valid price source on proposal creation.
- Fixes NDAX order book initialization to ensure valid prices are available upon book initialized.

- Adds test a test case to ensure that a valid price can be obtain once the order books are initialized.
- Adapts existing test cases.

**Tips for QA testing**:
Test that PMM does not throw a `c_tick` error on startup when using the NDAX exchange. There is no sure way of testing this; it used to fail some of the time, but often enough that if you successfully start/stop the strategy 10 times, it should suffice as validation.


hummingbot/connector/exchange/ndax/ndax_order_book_tracker.py
```diff
@@ -52,17 +51,6 @@ def exchange_name(self) -> str:
         """
         return CONSTANTS.EXCHANGE_NAME
 
-    async def _init_order_books(self):
```
aarmoa, line 55: The function `_init_order_books()` was overwritten in NdaxOrderBookTracker on purpose. Since the L2 order book event NDAX provides in its websocket API returns as the first response the complete order book status, we want to use that information to create the order book. If we don't overwrite the `_init_order_books()` function then the traditional order book initialization is performed: by making a request to the REST API. If this happens we are requesting two times the same information from the server (and consuming limit). It is true that initializing the order book might take a couple more seconds (because the order book data stream has to be initialized and started), but that is not really a problem for the strategy performance.

aarmoa (changes requested):

hummingbot/connector/exchange/ndax/ndax_order_book_tracker.py
petioptrv: Thank you for the clarification. I have re-introduced the delayed initialization, adjusting the point at which `_order_books_initialized` is set to reflect that

petioptrv: After further investigation and discussions with @aarmoa, function was removed once again to ensure the validity of the `self._order_books_initialized event.

rxlxrxsx (approved):

CrimsonJacket (approved):
Changes LGTM. But @aarmoa, @petioptrv  I'd like to know more about the reason for removing `_init_order_books()` completely.

petioptrv:
> Changes LGTM. But @aarmoa, @petioptrv I'd like to know more about the reason for removing `_init_order_books()` completely.

There is a time-period between the call to the custom `_init_order_books` method and when the order book data was delivered through the websocket. During that time, the order book is not actually initialized as requesting a price returns NaN, but it's still reported as initialized. This was throwing off PMM at `c_start`, which is now handled (PMM checks for NaN), but we reasoned it may throw off other strategies too. For that reason, and to preserve the meaning of the initialization event, we returned back to the default function.

aarmoa (approved):
I agree with the changes (I have discussed them with @petioptrv)

dennisocana:
Re-ran the tests since it failed now but it was passing earlier today.

PR #4013, opened 2021-08-27
Refactor/hummingbot uniswap interaction

james-hummingbot:


gateway/conf/gateway-config-example.yml
```diff
@@ -26,6 +26,6 @@ ETH_GAS_STATION_REFRESH_TIME: 60
 ETH_MANUAL_GAS_PRICE: 100
 
 # Uniswap
-UNISWAP_ALLOWED_SLIPPAGE: 1.5
+UNISWAP_ALLOWED_SLIPPAGE: '1/100'
```
james-hummingbot, line 29: This has changed to a ratio for correctness

gateway/src/services/base.ts
```diff
@@ -29,6 +29,40 @@ export const bigNumberWithDecimalToStr = (n: BigNumber, d: number): string => {
...
+  const zeros = '0'.repeat(d);
+  return BigNumber.from(numberStr + zeros);
+};
+
+export const gasCostInEthString = (
+  gasPrice: number,
+  gasLimit: number
+): string => {
+  return bigNumberWithDecimalToStr(
+    BigNumber.from(gasPrice)
```
martinkou, line 59: This is unsafe because floating point numbers (e.g. 2.1) are actually valid gas prices. `BigNumber` on the other hand, only deals with integers.

I'd recommend doing the `gasPrice` * `gasLimit` calculation in floating point and then convert to integer first - since that part usually doesn't overflow the precision of floating point numbers. After that, convert to `BigNumber`. e.g.

`BigNumber.from(Math.ceil(gasPrice * gasLimit)).mul(BigNumber.from(1e9))`

gateway/src/chains/ethereum/ethereum.routes.ts
```diff
@@ -38,6 +40,78 @@ export namespace EthereumRoutes {
...
+          const token = ethereum.getTokenBySymbol(symbol);
+          if (!token) {
+            continue;
+          }
+
+          tokens[symbol] = token;
+        }
+
+        let spender: string;
+        if (req.body.spender === 'uniswap') {
```
martinkou, line 80: This block is being repeated. Better refactor it out.

gateway/src/chains/ethereum/ethereum.routes.ts
```diff
@@ -38,6 +40,78 @@ export namespace EthereumRoutes {
...
+      async (
+        req: Request<{}, {}, EthereumAllowancesRequest>,
+        res: Response<EthereumAllowancesResponse | string, {}>
+      ) => {
+        const initTime = Date.now();
+        const wallet = ethereum.getWallet(req.body.privateKey);
+
+        const tokens: Record<string, Token> = {};
+
+        for (var i = 0; i < req.body.tokenSymbols.length; i++) {
```
martinkou, line 69: This block is also being repeated. Refactor.

martinkou (changes requested):

gateway/src/services/base.ts
james-hummingbot: Good to know, I didn't realize that about gas prices. Refactored.

gateway/src/chains/ethereum/ethereum.routes.ts
james-hummingbot: Thanks, refactored.

gateway/src/chains/ethereum/ethereum.routes.ts
james-hummingbot: Thanks, refactored.

martinkou (approved):

PR #4018, opened 2021-08-27
feat/adapt data streams connection to all perpetual bybit markets

aarmoa:

Added logic in order book data stream and user data stream to connect to both the inverse perpetual and the usdt perpetual markets when registering to events.

Unit tests

**Tips for QA testing**:
This PR does not require testing from QA. The testing will be done in the epic PR.

### DIFF
--- a/hummingbot/connector/derivative/bybit_perpetual/bybit_perpetual_api_order_book_data_source.py
+++ b/hummingbot/connector/derivative/bybit_perpetual/bybit_perpetual_api_order_book_data_source.py
@@ -14,7 +14,8 @@ from typing import (
 from hummingbot.connector.derivative.bybit_perpetual import bybit_perpetual_constants as CONSTANTS, \
     bybit_perpetual_utils
 from hummingbot.connector.derivative.bybit_perpetual.bybit_perpetual_order_book import BybitPerpetualOrderBook
-from hummingbot.connector.derivative.bybit_perpetual.bybit_perpetual_websocket_adaptor import BybitPerpetualWebSocketAdaptor
+from hummingbot.connector.derivative.bybit_perpetual.bybit_perpetual_websocket_adaptor import \
+    BybitPerpetualWebSocketAdaptor
 from hummingbot.core.data_type.funding_info import FundingInfo
 from hummingbot.core.data_type.order_book import OrderBook, OrderBookMessage
 from hummingbot.core.data_type.order_book_tracker_data_source import OrderBookTrackerDataSource
@@ -119,8 +120,8 @@ class BybitPerpetualAPIOrderBookDataSource(OrderBookTrackerDataSource):
                     resp_json = await response.json()
                     if "result" in resp_json:
                         for token_pair_info in resp_json["result"]:
-                            token_pair = trading_pair_symbol_map[token_pair_info["symbol"]]
-                            if token_pair in trading_pairs:
+                            token_pair = trading_pair_symbol_map.get(token_pair_info["symbol"], None)
+                            if token_pair and token_pair in trading_pairs:
                                 result[token_pair] = float(token_pair_info["last_price"])
         return result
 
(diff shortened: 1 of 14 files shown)


hummingbot/connector/derivative/bybit_perpetual/bybit_perpetual_user_stream_data_source.py
```diff
@@ -109,8 +116,30 @@ async def listen_for_user_stream(self, output: asyncio.Queue):
...
+            tasks = []
+            tasks.append(self._listen_for_user_stream_on_url(
+                url=bybit_perpetual_utils.wss_linear_private_url(self._domain),
+                output=output))
+            tasks.append(self._listen_for_user_stream_on_url(
+                url=bybit_perpetual_utils.wss_non_linear_private_url(self._domain),
+                output=output))
+
+            tasks_future = asyncio.gather(*tasks)
+            await tasks_future
```
CrimsonJacket, line 141: Technically speaking, the coroutine `_listen_for_user_stream_on_url()` does not end right? So `asyncio.gather()` does not really complete. I think we should use `safe_ensure_future()` instead and cancel when the bot stops.

hummingbot/connector/derivative/bybit_perpetual/bybit_perpetual_api_order_book_data_source.py
```diff
@@ -240,28 +248,61 @@ async def listen_for_subscriptions(self):
...
-                            self.logger().error("There was an error subscribing to the topic "
-                                                f"{json_message['request']['args']} ({json_message['ret_msg']})")
+                            self.logger().error(
+                                "There was an error subscribing to the topic "
+                                f"{json_message['request']['args']} ({json_message['ret_msg']}) on {url}")
                     else:
                         topic = json_message["topic"]
                         topic = ".".join(topic.split(".")[:-1])
-                        self._messages_queues.get(topic).put_nowait(json_message)
+                        await self._messages_queues[topic].put(json_message)
```
CrimsonJacket, line 259: Is there a reason why we are using `put()` instead of `put_nowait()`?

hummingbot/connector/derivative/bybit_perpetual/bybit_perpetual_api_order_book_data_source.py
```diff
@@ -240,28 +248,61 @@ async def listen_for_subscriptions(self):
...
 
             except asyncio.CancelledError:
                 raise
             except Exception as ex:
                 self.logger().network(
-                    f"Unexpected error with WebSocket connection ({ex})",
+                    f"Unexpected error with WebSocket connection on {url} ({ex})",
                     exc_info=True,
-                    app_warning_msg="Unexpected error with WebSocket connection. Retrying in 30 seconds. "
+                    app_warning_msg="Unexpected error with WebSocket connection on. Retrying in 30 seconds. "
```
CrimsonJacket, line 267: I think you're missing the `url` variable in the warning msg

hummingbot/connector/derivative/bybit_perpetual/bybit_perpetual_api_order_book_data_source.py
```diff
@@ -240,28 +248,61 @@ async def listen_for_subscriptions(self):
...
+                    url=bybit_perpetual_utils.wss_linear_public_url(self._domain),
+                    trading_pairs=linear_trading_pairs))
+            if non_linear_trading_pairs:
+                tasks.append(self._listen_for_subscriptions_on_url(
+                    url=bybit_perpetual_utils.wss_non_linear_public_url(self._domain),
+                    trading_pairs=non_linear_trading_pairs))
+
+            if tasks:
+                tasks_future = asyncio.gather(*tasks)
+                await tasks_future
```
CrimsonJacket, line 300: Similar comment to user stream.

CrimsonJacket (changes requested):
Some issues wrt creating the individual tasks to listen to linear and non-linear perpetuals.

hummingbot/connector/derivative/bybit_perpetual/bybit_perpetual_api_order_book_data_source.py
aarmoa: There are two reasons:
1.- There are going to be two tasks in parallel and I want to be sure there will be no conflicts caused by them inserting new elements in the same queue at the same time.
2.- Using the async put I am sure that I giving more opportunities for other tasks to grab the processor if there is any delay in the interaction with the queue.

hummingbot/connector/derivative/bybit_perpetual/bybit_perpetual_user_stream_data_source.py
aarmoa: I am not sure that is a good idea. If we do that, the result would be that the task created in the tracker, the one calling listen_for_user_stream, would finish (because once it creates the future if would finish). But currently the whole architecture is designed for that task to be a live task, that should be canceled if something goes wrong. Using `gather` I am keeping that same behaviour.

CrimsonJacket (approved):
I agree with the changes in this PR. Some revision is needed on accessing the `_funding_info` variable but it will be addressed in a separate PR #4026

PR #4019, opened 2021-08-27
Refactor/gateway v2 general changes

james-hummingbot:


willyPat:
- Replaces deprecated libraries (bodyparser/momentjs)
- Change sync calls for async calls for handling files (fs => fs/promises)
- Makes use of native JS features (find/forEach)
- Rewrite code to avoid unnecessary if/else and try/catch structures
- Adds init function to Ethereum base in order to keep context on failure
- Code reutilization improvements

PR #4071, opened 2021-09-02
Refactor/kucoin api throttler

petioptrv:

This PR refactors the KuCoin connector to use the the new `AsyncThrottler` class. It also introduces unit-tests for the portions of the code affected by the changes. The tests use `aioresponses` ~(pending approval from @martinkou)~, as well as the new network assistant class ~currently [under development](https://github.com/CoinAlpha/hummingbot/pull/4048)~.

- The unit-tests test the relevant portions of the code affected by the changes in this PR.

**Tips for QA testing**:
- Ensure that the requests are throttled (e.g. by setting the allocated rate-limit percentage to 5% and creating a strategy with 2 orders on each side).
- Ensure that the connector otherwise functions as expected (places, cancels , and detects fills on orders)


aarmoa (commented):
I have added a couple of comments. We can discussed about them in a meeting.

rxlxrxsx:
- When transferring USDT token from main account to trading account or vice versa, the bot fetches the balance either from your main account or trading account, depending on where you last transferred USDT. This only affects USDT token.
Screen Shot 2021-09-05 at 11 30 28 AM

Screen Shot 2021-09-05 at 11 30 57 AM

- Getting several log messages upon starting the bot.

Screen Shot 2021-09-05 at 2 47 05 PM
Screen Shot 2021-09-05 at 11 46 12 AM

- There are instances of inconsistent order creation happening as well.

Screen Shot 2021-09-05 at 1 39 10 PM

- Setting the `rate_limits_share_pct` to 5 will crash the bot after orders are placed, the only way to exit the bot is to close the terminal. Upon checking on the exchange the orders that were created also got stuck.

Screen Shot 2021-09-05 at 3 08 29 PM

Screen Shot 2021-09-05 at 3 08 16 PM

- Although the status output shows that the orders were placed at the same time, there are also delays on orders being created usually when rate limit warning is encountered, I tried stopping the bot midway before the remaining orders were not yet placed based on the logs and those orders got stuck in the exchange.

Screen Shot 2021-09-05 at 2 19 49 PM

Screen Shot 2021-09-05 at 2 46 50 PM
Screen Shot 2021-09-05 at 2 48 12 PM

Logs and config: [kucoin_throttler.zip](https://github.com/CoinAlpha/hummingbot/files/7111105/kucoin_throttler.zip)

hummingbot/connector/exchange/kucoin/kucoin_exchange.pyx
```diff
@@ -412,14 +411,15 @@ cdef class KucoinExchange(ExchangeBase):
...
+            str path_url = CONSTANTS.ACCOUNTS_PATH_URL
             str asset_name
             set local_asset_names = set(self._account_balances.keys())
             set remote_asset_names = set()
 
-        data = await self._api_request("get", path_url=path_url, is_auth_required=True)
-        if data:
-            for balance_entry in data["data"]:
+        data = {"type": "trade"}  # this GET request requires that arguments be submitted as data
+        response = await self._api_request("get", path_url=path_url, data=data, is_auth_required=True)
```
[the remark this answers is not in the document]
petioptrv: I tried sending the params as params instead of body, but was getting `Invalid KC-API-Sign` back from the server. When signing, the params should be passed as body and then encoded in the signature as well. I originally wanted to separate the parameters from the path url, but it's true that passing them directly into the url circumvents this issue, so I will return it back to the original configuration

hummingbot/connector/exchange/kucoin/kucoin_exchange.pyx
```diff
@@ -645,29 +647,30 @@ cdef class KucoinExchange(ExchangeBase):
...
-            "post",
-            path_url=path_url,
-            params=params,
-            data=params,
-            is_auth_required=True,
-            is_partner_required=True
-        )
+            data["price"] = str(price)
+            data["postOnly"] = True
+        async with self._throttler.execute_task(CONSTANTS.POST_ORDER_LIMIT_ID):
```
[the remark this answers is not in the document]
petioptrv: Absolutely. This is a mistake. Thanks for spotting it

test/hummingbot/connector/exchange/kucoin/test_kucoin_api_order_book_data_source.py
```diff
@@ -0,0 +1,244 @@
...
+        self.ev_loop.run_until_complete(client.__aexit__(None, None, None))
+
+    @aioresponses()
+    def test_get_snapshot_no_auth(self, mock_api):
+        url = CONSTANTS.BASE_PATH_URL + CONSTANTS.SNAPSHOT_NO_AUTH_PATH_URL
+        regex_url = re.compile(f"^{url}".replace(".", r"\.").replace("?", r"\?"))
+        resp = self.get_snapshot_mock()
+        mock_api.get(regex_url, body=json.dumps(resp))
+
+        client = self.ev_loop.run_until_complete(aiohttp.ClientSession().__aenter__())
```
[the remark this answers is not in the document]
petioptrv: True; fixed

hummingbot/connector/exchange/kucoin/kucoin_exchange.pyx
```diff
@@ -380,30 +372,37 @@ cdef class KucoinExchange(ExchangeBase):
...
+                           limit_id: Optional[str] = None) -> Dict[str, Any]:
         url = KUCOIN_ROOT_API + path_url
         client = await self._http_client()
         if is_auth_required:
             if is_partner_required:
-                headers = self._kucoin_auth.add_auth_to_params(method, path_url, params, partner_header=True)
+                headers = self._kucoin_auth.add_auth_to_params(method, path_url, data, partner_header=True)
             else:
-                headers = self._kucoin_auth.add_auth_to_params(method, path_url, params)
+                headers = self._kucoin_auth.add_auth_to_params(method, path_url, data)
```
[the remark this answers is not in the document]
petioptrv: This comes from the docs for [authentication](https://docs.kucoin.com/#authentication). It's the data that has to be encoded in the signature. This wasn't causing any issues before. I mostly made the change for the sake of the "leave the campground cleaner than you found it" principle :) The only call to `_api_request` that used `is_auth_required=True` and the `params` field was the `place_order` method, and I switched it to pass the params as `data`

petioptrv:
Thanks for testing this @rxlxrxsx.

> * When transferring USDT token from main account to trading account or vice versa, the bot fetches the balance either from your main account or trading account, depending on where you last transferred USDT. This only affects USDT token.

This was due to my changes to the way that specific request was done. I've reverted that as part of addressing @aarmoa's review comments.

> * Getting several log messages upon starting the bot.

This is me forgetting to remove my debugging statements. Now removed.

> * There are instances of inconsistent order creation happening as well.

I left the bot running for quite a while over lunch and did not encounter this, so I think it was resolved with the account issue above.

> * Setting the `rate_limits_share_pct` to 5 will crash the bot after orders are placed, the only way to exit the bot is to close the terminal. Upon checking on the exchange the orders that were created also got stuck.

After a long chase to figure out why this is happening, I now know that it has to do with reaching the websocket rate limits. You can circumvent this by setting the limits for `WS_CONNECTION_LIMIT_ID` and `WS_REQUEST_LIMIT_ID` in `kucoin_constants.py` to `NO_LIMIT` and testing again with the 5% allocation.

```Python
    RateLimit(WS_CONNECTION_LIMIT_ID, limit=NO_LIMIT, time_interval=60),
    RateLimit(WS_REQUEST_LIMIT_ID, limit=NO_LIMIT, time_interval=10),
```
That being said, I will dedicate some more time on trying to figure out exactly where the dead-lock happens since I think this may be useful for other connectors as well.

> * Although the status output shows that the orders were placed at the same time, there are also delays on orders being created usually when rate limit warning is encountered, I tried stopping the bot midway before the remaining orders were not yet placed based on the logs and those orders got stuck in the exchange.

This one is due to the status being derived from the strategy combined with the fact the order is created asynchronously by the exchange. The strategy sends the request to the exchange and starts tracking the order's age from that point on. If the request gets rate limited on the exchange's side, then the order's age according to the strategy will be wrong.

hummingbot/connector/exchange/kucoin/kucoin_exchange.pyx
```diff
@@ -766,30 +767,32 @@ cdef class KucoinExchange(ExchangeBase):
                              f"{trading_rule.min_order_size}.")
 
         try:
-            exchange_order_id = await self.place_order(order_id, trading_pair, decimal_amount, False, order_type,
-                                                       decimal_price)
             self.c_start_tracking_order(
                 client_order_id=order_id,
-                exchange_order_id=exchange_order_id,
+                exchange_order_id="",
```
aarmoa, line 772: By default when the exchange_order_id is still not set it should be None (this can be checked in the code from InFlightOrderBase)

hummingbot/connector/exchange/kucoin/kucoin_exchange.pyx
```diff
@@ -694,30 +693,32 @@ cdef class KucoinExchange(ExchangeBase):
                 raise ValueError(f"Buy order amount {decimal_amount} is lower than the minimum order size "
                                  f"{trading_rule.min_order_size}.")
         try:
-            exchange_order_id = await self.place_order(order_id, trading_pair, decimal_amount, True, order_type,
-                                                       decimal_price)
             self.c_start_tracking_order(
                 client_order_id=order_id,
-                exchange_order_id=exchange_order_id,
+                exchange_order_id="",
```
aarmoa, line 698: By default when the exchange_order_id is still not set it should be None (this can be checked in the code from InFlightOrderBase)

hummingbot/connector/exchange/kucoin/kucoin_exchange.pyx
```diff
@@ -819,13 +822,20 @@ cdef class KucoinExchange(ExchangeBase):
...
+                raise KucoinInFlightOrderNotCreated(
+                    f"Failed to cancel order - {order_id}. Order not yet created."
+                    f" This is most likely due to rate-limiting."
+                )
             path_url = f"{CONSTANTS.ORDERS_PATH_URL}/{tracked_order.exchange_order_id}"
             await self._api_request(
                 "delete", path_url=path_url, is_auth_required=True, limit_id=CONSTANTS.DELETE_ORDER_LIMIT_ID
             )
+        except KucoinInFlightOrderNotCreated:
+            raise
```
aarmoa, line 838: If there is a possibility that the order creation is delayed by the throttler and the cancelation is executed before the creation, maybe we should apply the same fix it was applied for AscendEx, to be sure that the creation is not executed at all afterwards.

aarmoa (changes requested):
I have added some comments to the last changes

rxlxrxsx:
> * When transferring USDT token from main account to trading account or vice versa, the bot fetches the balance either from your main account or trading account, depending on where you last transferred USDT. This only affects USDT token.

This is confirmed fixed.

> * Getting several log messages upon starting the bot.

I still get this message every 10 seconds.

Screen Shot 2021-09-07 at 2 21 22 PM

And this one every 2 minutes.

Screen Shot 2021-09-07 at 2 30 07 PM

> * There are instances of inconsistent order creation happening as well.

This is confirmed fixed.

hummingbot/connector/exchange/kucoin/kucoin_exchange.pyx
petioptrv: Ah yes, fixed (in next push)

hummingbot/connector/exchange/kucoin/kucoin_exchange.pyx
petioptrv: ditto

hummingbot/connector/exchange/kucoin/kucoin_exchange.pyx
petioptrv: I forgot to update you on that one @aarmoa: @CrimsonJacket and I had a meeting after our initial one and he pointed out that, on cancel request, we can simply detect that the order is still only "locally" created and not confirmed, and subsequently ignore the request. Eventually, the rate-limited order-creation request will go through, the strategy will request a cancel, and that will go through as well.

It's true that we open ourselves to a possibility where this no-longer wanted order gets submitted post rate-limit and immediately filled, but we reasoned that this is a very unlikely scenario and that the reduction in complexity is worth it.

You can see the code for this in `execute_cancel:828`

petioptrv:
> I still get this message every 10 seconds.

@rxlxrxsx a straggler debugging statement, now removed

aarmoa (approved):
Changes look good to me

rxlxrxsx:
> Although the status output shows that the orders were placed at the same time, there are also delays on orders being created usually when rate limit warning is encountered, I tried stopping the bot midway before the remaining orders were not yet placed based on the logs and those orders got stuck in the exchange.

- I tried reproducing this again and noticed that the behavior changed. As you can see on the 1st screenshot I tried hitting `stop` while the orders are still being created and 1 sell order still pushed though resulting to the order getting stuck. Entering `start` again prevents me from restarting the bot (second screenshot) and only when exiting the bot that the stuck order will get cancelled. I randomly experience this only when the rate limit warning appears, usually when using rate limit % of 9 below. Tested with `WS_CONNECTION_LIMIT_ID` and `WS_REQUEST_LIMIT_ID` set to NO_LIMIT.

Screen Shot 2021-09-08 at 2 52 47 PM
Screen Shot 2021-09-08 at 2 53 09 PM
Screen Shot 2021-09-08 at 2 53 23 PM

Logs and config: [kucoin_ratelimit.zip](https://github.com/CoinAlpha/hummingbot/files/7126663/kucoin_ratelimit.zip)

petioptrv:
As per my discussions with @rxlxrxsx and @CrimsonJacket, the issues have been addressed in the following way:

- The infinite-loop scenario was diagnosed and addressed.
- The connector now checks the rate limit for the websocket connections and warns the user if it's too low for the connector's proper operation.
- API endpoints that have no rate limits are no longer passing through throttler context acquisition to prevent being blocked by the rate-limited endpoints.

@rxlxrxsx to test the changes, please run again with the 5% rate-limit allocation to observe the new warning and the fix for the client-freeze due to the infinite loop. The warning message should not show at 10% and above.

rxlxrxsx:
- Confirmed that the bot doesn't crash anymore when setting the rate limit to 9% or below however there are still issues found:

- Tested with rate limit set to 10% and I get this behavior where the bot takes time before it successfully cancels the orders

Screen Shot 2021-09-08 at 4 18 09 PM

- With rate limit set to 10% as well, I still get the warning message.

Screen Shot 2021-09-08 at 7 07 17 PM

Logs and config: [kucoin_1.zip](https://github.com/CoinAlpha/hummingbot/files/7128925/kucoin_1.zip)

petioptrv:
> * Tested with rate limit set to 10% and I get this behavior where the bot takes time before it successfully cancels the orders

On this specific instance, it looks like the websocket has not started so the cancelled orders are only detected by the API requests made in the backup loop. This seems to be isolated to that run however. The log message `The market order buy-DOGE-BTC-XYZ has been cancelled according to order status API.` appears elsewhere in the log, but it seems accounted for. For instance, the run right after this one starts off with this issue, but then the websocket kicks in and we start receiving the cancellation confirmations more promptly. Do you have a config that can reliably reproduce this issue?

> * With rate limit set to 10% as well, I still get the warning message.

This message is issued by the throttler simply warning that we're nearing the limit. The newly added message would say:

```Python
            self.logger().warning(
                f"The KuCoin connector requires 3 websocket connections to operate. The current rate limit percentage"
                f" allows the creation of {int(effective_ws_connection_limit)} websocket connections every"
                f" {CONSTANTS.WS_CONNECTION_TIME_INTERVAL}s. This will prevent the client from functioning properly."
            )
```

hummingbot/connector/exchange/kucoin/kucoin_api_order_book_data_source.py
```diff
@@ -107,13 +107,12 @@ async def get_ws_connection_context(cls, throttler: Optional[AsyncThrottler] = N
         throttler = throttler or cls._get_throttler_instance()
         async with aiohttp.ClientSession() as session:
             url = CONSTANTS.BASE_PATH_URL + CONSTANTS.PUBLIC_WS_DATA_PATH_URL
-            async with throttler.execute_task(CONSTANTS.PUBLIC_WS_DATA_PATH_URL):
-                async with session.post(url, data=b'') as resp:
-                    response: aiohttp.ClientResponse = resp
-                    if response.status != 200:
-                        raise IOError(f"Error fetching Kucoin websocket connection data."
-                                      f"HTTP status is {response.status}.")
-                    data: Dict[str, Any] = await response.json()
```
aarmoa: I think we should always pass through the throttler before performing any request. If there are no limits for the particular endpoint that should be detected by the throttler based on its configuration.

hummingbot/connector/exchange/kucoin/kucoin_exchange.pyx
```diff
@@ -389,17 +393,19 @@ cdef class KucoinExchange(ExchangeBase):
...
-                response = await client.post(url, params=params, data=data, headers=headers)
+            response = await client.post(url, params=params, data=data, headers=headers)
         elif method == "delete":
-            async with self._throttler.execute_task(limit_id):
-                response = await client.delete(url, headers=headers)
+            response = await client.delete(url, headers=headers)
         else:
             response = False
+        if not throttler_passthrough:
+            await context.__aexit__(None, None, None)
```
aarmoa: We should not invoke the enter and exit codes like this. The execution using with takes care of, between other things, ensuring the exit code is always executed. We should not break encapsulation like this.

hummingbot/connector/exchange/kucoin/kucoin_api_order_book_data_source.py
petioptrv: Agreed and reverted

hummingbot/connector/exchange/kucoin/kucoin_exchange.pyx
petioptrv: ditto

PtrckM (approved):
confirmed fixed on the reported issues. Approving in behalf of QA

aarmoa (approved):
I agree with the latest changes

PR #4097, opened 2021-09-06
Feat/gateway v2 tests

willyPat:
- Adds fixtures and unit test for recreating the out of gas scenario during polling.


gateway/manual-tests/curl.sh
```diff
@@ -14,11 +14,11 @@ curl -X POST -H "Content-Type: application/json" -d "{\"privateKey\":\"$ETH_PRIV
 
 # approve Ethereum to spend your WETH
 
-curl -X POST -H "Content-Type: application/json" -d "{\"privateKey\":\"$ETH_PRIVATE_KEY\",\"spender\":\"0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D\",\"token\":\"DAI\"}" localhost:5000/eth/approve
+curl -X POST -H "Content-Type: application/json" -d "{\"privateKey\":\"0b8040a6719bc21e2fce075b843e33cc1fab503c105c76bd5375e3f784724bc6\",\"spender\":\"0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D\",\"token\":\"DAI\"}" localhost:5000/eth/approve
```
james-hummingbot, line 17: You don't want do this! If you don't want to use env var then make another file and add it to gitignore or outside the hummingbot project.

willyPat: Yes you are right! I forgot to remove it, I paste it there for convenience in my local env. 
Its a dummy wallet

gateway/test-scripts/ethereum-routes.test.ts
```diff
@@ -0,0 +1,25 @@
...
+
+describe('Eth endpoints', () => {
+  let eth: Ethereum;
+  beforeAll(async () => {
+    eth = Ethereum.getInstance();
+    await eth.init();
+  });
+  it('should get an OUT of GAS error', async () => {
+    eth.getTransaction = jest.fn().mockReturnValue(transactionOutOfGas);
+    eth.getTransactionReceipt = jest.fn().mockReturnValue(transactionOutOfGasReceipt);
```
james-hummingbot, line 17: It looks like this is mocked, so I am guessing this is a unit test? the dir `gateway/test-scripts` are things we would run against a live network and not in CI, perhaps it is poorly named. The unit tests are in `gateway/test`

willyPat: I moved it to the test folder and it runs with npm run test:unit

james-hummingbot (approved):
Great, it looks good now

PR #4099, opened 2021-09-06
Feat/gateway v2 eth init

willyPat:
- Adds middlewares for ETH and Uniswap to ensure they are available before the route code is executed.
- Adds init functions for eth-base and uniswap

### DIFF
--- a/gateway/src/services/ethereum-base.ts
+++ b/gateway/src/services/ethereum-base.ts
@@ -20,7 +20,7 @@ export class EthereumBase {
   // there are async values set in the constructor
   private _ready: boolean = false;
   private initializing: boolean = false;
-  private initPromise: Promise<void>;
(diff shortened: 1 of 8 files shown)


gateway/src/services/ethereum-base.ts
```diff
@@ -19,6 +19,8 @@ export class EthereumBase {
   private tokenMap: Record<string, Token> = {};
   // there are async values set in the constructor
   private _ready: boolean = false;
+  private initializing: boolean = false;
+  private initPromise: Promise<void>;
```
james-hummingbot, line 23: This isn't initiated in the constructor. Typescript is complaining.

james-hummingbot (approved):
After reviewing and testing the code with @willyPat , I think this is ready.

Issue #4178, opened 2021-09-14
Update the Kraken throttler rate limits to decay-based


petioptrv:
As a Kraken connector user, I want to use the entirety of API calls available for my account subscription.

## Details
The Kraken API uses [decay-based](https://docs.kraken.com/rest/#section/Rate-Limits/REST-API-Rate-Limits) rate limits rather than sliding time window ones. Currently, the Kraken connector limits API calls to 20 per minute, but if a user has a Pro account, they should be able to perform a 20-requests burst in one second and then perform one request for each of the 59 remaining seconds in that minute window, assuming we start from a baseline of zero requests. So, the user should be able to perform 79 requests in a minute if those requests are spread out as previously described. 

The frequency of the polling loop in the connector was set to 30s. This task includes investigating if we can increase that frequency while remaining in the capacity of the basic account tier.

## Definition of Done
* [ ] The Kraken connector is updated to use the decay-based schema according to the Kraken documentation.
* [ ] ~The Kraken connector is updated to request the account tier from the user and then dynamically set the rate-limits according to the tier.~
* [ ] ~The Kraken connector sets the `retry_interval` of the `AsyncThrottler` to the appropriate given the account tier.~
* [ ] Check if the polling loop frequency can be increased while remaining within the capacity of the basic account tier.

Edit: the tier-related points were moved into a separate task (#4254)

┆Issue is synchronized with this [Clickup task](https://app.clickup.com/t/1ran4m6) by [Unito](https://www.unito.io)

aarmoa:
I think we need to make this part of #4163. Without the inclusion of the decaying throttler in Kraken connector QA will not be able to test the changes in the throttler.

dennisocana:
After our recent discussion, we agreed to create both PRs and merge them into an epic branch where QA can test it when both tasks have already been completed.

dennisocana:
Refer to this comment: https://github.com/CoinAlpha/hummingbot/issues/4163#issuecomment-925360622

Issue #4163, opened 2021-09-10
Extend `AsyncThrottler` to handle decay-based rate limits


petioptrv:
As a developer, I want to be able to model decay-based rate limits when implementing connectors.

## Details

The Kraken API uses [decay-based](https://docs.kraken.com/rest/#section/Rate-Limits/REST-API-Rate-Limits) rate limits rather than sliding time window ones. Currently, the Kraken connector limits API calls to 20 per minute, but if a user has a Pro account, they should be able to perform a 20-requests birst in one second and then perform one request for each of the 59 remaining seconds in that minute window, assuming we start from a baseline of zero requests. So, the user should be able to perform 79 requests in a minute if those requests are spread out as previously described. 

There exists a [library](https://github.com/dominiktraxl/pykrakenapi) that models this rate-limiting behaviour. The issue is that the library is not asynchronous. However, the rate-limiting portion of the code is short.

The proposition is to integrate that functionality into our existing throttler. 

## Implementation Suggestions

This can be performed by adding an internal counter for the rate limits, adding a new optional argument per `RateLimit` object to accept the decay rate, and making the time interval optional as well. The user can then supply either the decay rate or the time window and that rate limit will be evaluated accordingly.

## Definition of Done

- [ ] The `AsyncThrottler` is extended to support decay-based rate limits.
- [ ] The Kraken connector is updated to use this new schema according to the Kraken documentation.

dennisocana:
@petioptrv since we reverted https://github.com/CoinAlpha/hummingbot/issues/3892 to its original approach and no longer do as described here, do we still need this story?

petioptrv:
@dennisocana, the original story was reverted so that, for now, we only change the throttler used in Kraken to the unified async throttler, keeping the rate-limits currently in place. However, this story will enable us to handle decaying limits, and so, to eventually more accurately model the Kraken limits.

aarmoa:
It would be interesting to dedicate some time to the solution design. We can create a new time of rate limit for this, and maybe refactor the current throttler logic, to make the same throttler able to handle both time of limits (weighed and decaying) without extending the logic in the current classes.

dennisocana:
Since the current `AsyncThrottler` is already complex, adding more functionality into it may overcomplicate which will make it harder to maintain. A better approach is to create an entirely new throttler by refactoring the base class to make it cleaner and easier.

[...]

PR #4180, opened 2021-09-14
refactor / Kraken `AsyncThrottler` integration

petioptrv:

This PR integrates the unified `AsyncThrottler` into the Kraken exchange. It preserves the existing throttling schema, with tasks #4163 and #4178 created to update the schema to the decay-based one used by the exchange.

- Test cases added for the affected sections of the code.

**Tips for QA testing**:
- Test rate-limitations by lowering the allocated rate-limit percentage (current rate limit is 20 requests/minute).


hummingbot/connector/exchange/kraken/kraken_api_user_stream_data_source.py
```diff
@@ -66,30 +66,30 @@ async def get_auth_token(self) -> str:
...
 
-    async def listen_for_user_stream(self, ev_loop: asyncio.BaseEventLoop, output: asyncio.Queue):
+    async def listen_for_user_stream(self, ev_loop: asyncio.AbstractEventLoop, output: asyncio.Queue):
         while True:
             try:
-                async with websockets.connect(KRAKEN_WS_URL) as ws:
-                    ws: websockets.WebSocketClientProtocol = ws
-
+                async with self._throttler.execute_task(CONSTANTS.WS_CONNECTION_LIMIT_ID):
+                    ws = await websockets.connect(CONSTANTS.WS_AUTH_URL)
```
aarmoa, line 92: If we are not using `async with websockets.connect ...` then we need to add a `finally`component to the try to close the ws connection if it was stablished.

hummingbot/connector/exchange/kraken/kraken_exchange.pyx
```diff
@@ -98,17 +93,10 @@ cdef class KrakenExchange(ExchangeBase):
     API_CALL_TIMEOUT = 10.0
     KRAKEN_TRADE_TOPIC_NAME = "kraken-trade.serialized"
     KRAKEN_USER_STREAM_TOPIC_NAME = "kraken-user-stream.serialized"
```
aarmoa, line 95: These globals seem to not be required

hummingbot/connector/exchange/kraken/kraken_exchange.pyx
```diff
@@ -192,10 +179,12 @@ cdef class KrakenExchange(ExchangeBase):
...
         self._in_flight_orders.update(in_flight_orders)
 
-    async def asset_pairs(self) -> Dict[str, Any]:
+    async def get_asset_pairs(self) -> Dict[str, Any]:
         if not self._asset_pairs:
             client = await self._http_client()
-            asset_pairs_response = await client.get(ASSET_PAIRS_URI)
+            url = f"{CONSTANTS.BASE_URL}{CONSTANTS.ASSET_PAIRS_PATH_URL}"
+            async with self._throttler.execute_task(CONSTANTS.ASSET_PAIRS_PATH_URL):
+                asset_pairs_response = await client.get(url)
```
aarmoa, line 187: Is it ok to not reuse `_api_request` to make this http request?

hummingbot/connector/exchange/kraken/kraken_exchange.pyx
```diff
@@ -364,6 +355,8 @@ cdef class KrakenExchange(ExchangeBase):
             for order_update, tracked_order in zip(results, tracked_orders):
                 client_order_id = tracked_order.client_order_id
 
+                print(f"Checking {client_order_id}")
```
aarmoa, line 358: This print is not required

hummingbot/connector/exchange/kraken/kraken_in_flight_order.pyx
```diff
@@ -86,6 +94,11 @@ cdef class KrakenInFlightOrder(InFlightOrderBase):
             "userref": self.userref
         }
 
+    def update_exchange_order_id(self, exchange_id: str):
+        super().update_exchange_order_id(exchange_id)
+        if self.is_local:
+            self.last_state = "pending"
```
aarmoa, line 100: Before the PR the initial status was "NEW". Now, the status for the confirmed order in the exchange is "pending". Is that correct?

test/hummingbot/connector/exchange/kraken/test_kraken_api_order_book_data_source.py
```diff
@@ -0,0 +1,251 @@
...
+    @aioresponses()
+    def test_get_new_order_book(self, mocked_api):
+        url = f"{CONSTANTS.BASE_URL}{CONSTANTS.SNAPSHOT_PATH_URL}"
+        regex_url = re.compile(f"^{url}".replace(".", r"\.").replace("?", r"\?"))
+        resp = self.get_depth_mock()
+        mocked_api.get(regex_url, body=json.dumps(resp))
+
+        ret = self.async_run_with_timeout(self.data_source.get_new_order_book(self.trading_pair))
+
+        self.assertTrue(isinstance(ret, OrderBook))
```
aarmoa, line 224: It would be better to assert something based on the status of the order book, instead of checking for the class

test/hummingbot/connector/exchange/kraken/test_kraken_api_order_book_data_source.py
```diff
@@ -0,0 +1,251 @@
...
+        self.mocking_assistant.add_websocket_text_message(
+            websocket_mock=ws_connect_mock.return_value, message=json.dumps(resp)
+        )
+        output_queue = asyncio.Queue()
+
+        self.ev_loop.create_task(self.data_source.listen_for_trades(self.ev_loop, output_queue))
+        self.mocking_assistant.run_until_all_text_messages_delivered(websocket_mock=ws_connect_mock.return_value)
+
+        self.assertTrue(not output_queue.empty())
+        self.assertTrue(isinstance(output_queue.get_nowait(), OrderBookMessage))
```
aarmoa, line 251: Here again it would be better to assert something based on the content of the order book message, instead of checking its class.

test/hummingbot/connector/exchange/kraken/test_kraken_exchange.py
```diff
@@ -0,0 +1,437 @@
...
+                self.trading_pair,
+                amount=Decimal(quantity),
+                order_type=OrderType.LIMIT,
+                price=Decimal(price),
+                userref=userref,
+            )
+        )
+
+        self.assertEqual(self.event_listener.events_count, 1)
+        self.assertTrue(order_id in self.exchange.in_flight_orders)
```
aarmoa, line 356: Instead of assertTrue(... in ...) we should use assertIn

test/hummingbot/connector/exchange/kraken/test_kraken_exchange.py
```diff
@@ -0,0 +1,437 @@
...
+                order_id,
+                self.trading_pair,
+                amount=Decimal(quantity),
+                order_type=OrderType.LIMIT,
+                price=Decimal(price),
+                userref=userref,
+            )
+        )
+
+        self.assertEqual(self.event_listener.events_count, 1)
```
aarmoa, line 355: We should add assertions to check the event present in the listener is the expected one

test/hummingbot/connector/exchange/kraken/test_kraken_exchange.py
```diff
@@ -0,0 +1,437 @@
...
```
aarmoa, line 385: We should assert the event present in the listener is the expected one

test/hummingbot/connector/exchange/kraken/test_kraken_exchange.py
```diff
@@ -0,0 +1,437 @@
...
+            price=2,
+            amount=1,
+            order_type=OrderType.LIMIT,
+            userref=1,
+        )
+        self.exchange.in_flight_orders[order_id].update_exchange_order_id(exchange_id)
+        self.exchange.add_listener(MarketEvent.OrderCancelled, self.event_listener)
+        ret = self.async_run_with_timeout(self.exchange.execute_cancel(self.trading_pair, order_id))
+
+        self.assertEqual(self.event_listener.events_count, 1)
```
aarmoa, line 417: We should assert the event present in the listener is the expected one

aarmoa (changes requested):

hummingbot/connector/exchange/kraken/kraken_exchange.pyx
petioptrv: Added the use of `_api_request` to standardize the call

hummingbot/connector/exchange/kraken/kraken_in_flight_order.pyx
petioptrv: Ah yes, I see your point. Leaving it as "local" and letting the ws order status update do its work

aarmoa: What I expected was to see we were configuring the status as NEW once we receive the exchange_order_id, instead of "pending", since before this PR the initial status was NEW and not "pending"

rxlxrxsx:
Getting this error after connecting the API keys which prevents us from further testing the PR.

Screen Shot 2021-09-15 at 1 31 32 PM

petioptrv:
> Getting this error after connecting the API keys which prevents us from further testing the PR.

Thanks for testing Rein, this should be now good to go. Note that the API limit is currently been hit quite fast. With a basic single-level PMM that refreshes orders every 10s, it will be attained within the first minute.

rapcmia:
Tested with default settings on a single order at 60s refresh time but getting rate limit of 20calls/60s. The bot just started and failed to created orders

image

Attached: 
[logs_pmm-test.log](https://github.com/CoinAlpha/hummingbot/files/7169794/logs_pmm-test.log)

petioptrv:
I started investigating how many calls the exchange consumes normally, which led me to look further into the Kraken documentation. It turns out the order creations/cancellations are on a separate limit that starts at 60 with decay that decay at a rate of 1/s and up, so that's good. The public endpoints are also rate-limited at 1call/second.

However, the private endpoints are limited to as low as 15 counts with a decay as low as 1/3s, allowing a sustain request rate of about 3 counts every 10s or 10 counts every 30s. `_update_balances` makes two calls that each cost 2 counts. `_update_order_status` makes a request per open trade and each request costs 2 counts. With a strategy of 1 order on each side and a polling loop of 10s, we need 6 counts every 10s, which the base account type does not provide. If we extend the polling interval to 30s, which I did in the latest commits, we can now support the base account (and up), but going to 2 orders on each side will require 10 counts every 30s, which is cutting it close.

For that reason, I've added a check warning the user if the allocated rate limit percentage has been reduced.

hummingbot/connector/exchange/kraken/kraken_in_flight_order.pyx
petioptrv: I looked around but there doesn't seem to be any use of "NEW" other than being the previous default value. Now "NEW" is essentially replaced with "local", which is more in line with what we have in the previous integrations

rapcmia:
Tried to run a pure market-making strategy with default settings but was still not able to create order and rate limit triggered right away `1call/1s`. Tried to wait for 5mins to check if there will changes but no go

image

petioptrv:
> Tried to run a pure market-making strategy with default settings but was still not able to create order and rate limit triggered right away `1call/1s`. Tried to wait for 5mins to check if there will changes but no go
> 
> image

This is due to the an exchange rule that dictates the minimum order amount for that pair be 50, so the strategy does not generate any proposals. I've successfully tested with `BTC-USDT` at order amount of 0.0001. I'll create a task to include logging in PMM letting the user know why no proposals were generated.

RC-13:
> Tried to run a pure market-making strategy with default settings but was still not able to create order and rate limit triggered right away 1call/1s. Tried to wait for 5mins to check if there will changes but no go

- Confirming that orders are created even after receiving rate limit usage warning `API rate limit on PublicEndpointLimitID (1 calls per 1s) has almost reached. Limits used is 1 in the last 1 seconds` as long as the order amount meets the exchange minimum allowable trade size of the exchange.
- For confirmation though, whenever I set a refresh time of 20 sec or lower without adjusting the default rate_limit_sharing, the orders are failing to get cancelled however after the order age goes up to more than 1 min 10 sec, the order gets cancelled and creates a new set. Is this due to the 20 request per minute limit?
image
- Reducing the rate limit appears to be not supported
image

petioptrv:
> * For confirmation though, whenever I set a refresh time of 20 sec or lower without adjusting the default rate_limit_sharing, the orders are failing to get cancelled however after the order age goes up to more than 1 min 10 sec, the order gets cancelled and creates a new set. Is this due to the 20 request per minute limit?

Nice catch @RC-13. This led me to uncover a bunch of deeper-rooted issues with how order statuses are currently handled. I opened an issue to address this (#4208).

@aarmoa, for now, I have re-introduced `last_state = "new"` to fix the problems outlined in the new issue.

aarmoa (approved):
I agree with the changes. I think we need to add some kind of reminder to undo changes to polling interval once the decaying throttler is ready.

RC-13 (approved):
LGTM, approving in behalf of QA. @petioptrv we're unable to get the `Unexpected response code: 504` during testing.

Issue #4106, opened 2021-09-07
Mock transaction got stuck due to low gas fee error case


martinkou:
## Why

See #3991. This task concerns the case where a transaction got stuck due to low gas fee.

## What

* Implement a mock test case with a simulated stuck transaction, and ensure the `/poll` API output conforms to the specs

martinkou:
Acceptance criteria:

* Capture the return / errors he gets from ethers.js for an actual stuck transaction, and document it.
* Simulate the same return value / error in the mock.
* Construct a test case for `/poll` and verify that it processes the return value / error correctly, and returns with the appropriate error code and message.

mifeng:
Currently, using `ethers.js` and Infura, we can't tell whether a transaction is:
- never sent to mempool
- in mempool and likely to be confirmed
- in mempool and unlikely to be confirmed

Instead, for any unconfirmed transaction, `getTransactionReceipt` returns `{ gasUsed: 0, blockNumber: 0, confirmations: 0, status: 0, logs: [] }`

For example, see response for this transaction `https://etherscan.io/tx/0x8a7b3f1acc58cadb4ca22b258cb33d64d8550675b969c2d21ac8a5d46b65f940`, which has a very low gas price and will likely never be confirmed.

I believe we should mock this test for general unconfirmed transactions and create a research ticket to see if there are other ways that we can detect unconfirmed tx status

PR #4211, opened 2021-09-17
Adds unconfirmed transaction test

willyPat:
Creates test for unconfirmed transactions

### DIFF
--- a/gateway/test/services/ethereum-routes.test.ts
+++ b/gateway/test/services/ethereum-routes.test.ts
@@ -3,6 +3,7 @@ import { app } from '../../src/app';
 import { Ethereum } from '../../src/chains/ethereum/ethereum';
 import * as transactionOutOfGas from './fixtures/transaction-out-of-gas.json';
 import * as transactionOutOfGasReceipt from './fixtures/transaction-out-of-gas-receipt.json';
+import * as transactionUnconfirmedReceipt from './fixtures/transaction-unconfirmed-receipt.json';
 
 const OUT_OF_GAS_ERROR_CODE = 1003;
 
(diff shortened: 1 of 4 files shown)


gateway/test/services/ethereum-routes.test.ts
```diff
@@ -12,7 +13,7 @@ describe('Eth endpoints', () => {
     eth = Ethereum.getInstance();
```
martinkou, line 13: Since we're running multiple tests that involve monkey-patching methods in the Ethereum object - it'd be cleaner if we initialize a new Ethereum object for each test, rather than sharing the same object. This would prevent previous methods or states in one test case from leaking into a latter test case.

martinkou (changes requested):

gateway/test/services/ethereum-routes.test.ts
martinkou: Actually... let's just drop this comment for the moment. Looks like we need a more general solution to the mock leakage issue here. I'm gonna file a separate task to find a good solution for all test cases.

martinkou:
@willyPat could you fix the test case s.t. for unconfirmed transactions, the mock transaction receipt should be null, while the mock transaction object should be non-empty but with null block number?

willyPat:
@martinkou To clarify the testing cases:
1) The transaction didn't reach the mempool =>   Tx = null, TxReceipt = null 
2) Transaction stuck (in the mempool but not processed) => Tx = Object, TxReceipt = null **OR** { gasUsed: 0, blockNumber: 0, confirmations: 0, status: 0, logs: [] } 
For case 2 in my tests I always got null for TxReceipt but I added the case of getting that special object with everything in 0 based on @mifeng comment in ticket #4106

PR #4222, opened 2021-09-20
fix/ Bybit Perpetual misc issues

CrimsonJacket:

Resolves remaining issues with Bybit Perpetual connector.

Fix `timestamp` parameter in `Auth` class to use current time.
Fix `_create_order()` function to use adequate parameters depending on `PositionAction`
Fix position event updates to use the right `amount` value depending on `SHORT` or `LONG` position
Includes refactor of unit tests to use `aioresponses` package.

**Tips for QA testing**:
All testing to be performed in `epic/bybit_connector` branch

### DIFF
--- a/test/hummingbot/connector/derivative/bybit_perpetual/test_bybit_perpetual_derivative.py
+++ b/test/hummingbot/connector/derivative/bybit_perpetual/test_bybit_perpetual_derivative.py
@@ -59,6 +59,7 @@ class BybitPerpetualDerivativeTests(TestCase):
         self.connector.logger().addHandler(self)
 
         self.mocking_assistant = NetworkMockingAssistant()
+        self.mock_done_event = asyncio.Event()
 
         BybitPerpetualAPIOrderBookDataSource._trading_pair_symbol_map = {
             "bybit_perpetual_testnet": {self.ex_trading_pair: self.trading_pair}}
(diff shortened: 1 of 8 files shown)


petioptrv (changes requested):
Code looks good. But I think we should add a test case for _update_balances and _update_positions in the case of short positions, and another one for _update_positions validating that it sends a request for each one of the trading pairs

test/hummingbot/connector/derivative/bybit_perpetual/test_bybit_perpetual_derivative.py
```diff
@@ -121,53 +170,18 @@ def test_create_buy_order(self, post_mock, nonce_provider_mock):
...
-                    "updated_at": "2019-11-30T11:03:43.455Z"
-                },
-                "time_now": "1575111823.458705",
-                "rate_limit_status": 98,
-                "rate_limit_reset_ms": 1580885703683,
-                "rate_limit": 100
-            })
-        buy_request_url, buy_request_headers, buy_request_data = asyncio.get_event_loop().run_until_complete(
-            self.mocking_assistant.next_sent_request_data(post_mock))
+        asyncio.get_event_loop().run_until_complete(asyncio.sleep(0.5))
```
petioptrv, line 173: This would be a nice place to use the callback functionality and set an `asyncio.Event` that can be awaited here

petioptrv (dismissed):
I left a comment on a specific test case, but I see that this is relevant to other test cases too: I think that, to check if a call has been made, it's better to use `aioresponses`'s mock callback to set an `asyncio.Event` and then await that event. This will reduce the runtime of the tests.

The rest LGTM

petioptrv (approved):
Changes LGMT!

PR #4235, opened 2021-09-21
refactor/kucoin websockets removal

petioptrv:

- Replaces the use of `websockets` with `aiohttp`.
- Attempts to simplify and cleanup the code where possible (e.g. converts class methods to instance methods if always called from an instance).

- Ensures that the current test cases all pass.

**Tips for QA testing**:
- Check that the functioning of the connector is as expected after the websocket changes.


rapcmia:
When testing Pure market-making, I noticed there are errors of `Unexpected error while maintaining the user event listen key. Retrying after 5 seconds...` on some instances of the order cycle. The bot seems to recover from a sudden disconnection however when the error happens while the bot is about to reach the refresh time or cancellation phase, it fails to cancel the order on the client but is successful on Kucoin

unable to cancel order 2

Here's also a recording of the issue however this time the bot managed to cancel one side of the orders. The issue started at 10s:

https://user-images.githubusercontent.com/73840223/134161108-fe7028ff-519e-4744-8be9-e6789dfc08f1.mp4

Steps to reproduce:
1. Create a Pure market-making strategy
2. Start the bot and notice the error per order refresh time
3. Monitor the order when its closest to the refresh time and when the error occurred, the bot fails to cancel the order

Attached:
[4235.zip](https://github.com/CoinAlpha/hummingbot/files/7202637/4235.zip)

rapcmia (approved):
Tested and working ok. Approved on behalf of QA team

hummingbot/connector/exchange/kucoin/kucoin_api_user_stream_data_source.py
```diff
@@ -60,50 +56,50 @@ async def get_listen_key(self):
...
             return
 
-    async def get_ws_connection(self) -> websockets.WebSocketClientProtocol:
+    async def get_ws_connection(self, client: aiohttp.ClientSession) -> aiohttp.ClientWebSocketResponse:
         stream_url: str = f"{self._current_endpoint}?token={self._current_listen_key}&acceptUserMessage=true"
         self.logger().info(f"Connecting to {stream_url}.")
 
         async with self._throttler.execute_task(CONSTANTS.WS_CONNECTION_LIMIT_ID):
-            return await websockets.connect(stream_url, ping_interval=40, ping_timeout=self.PING_TIMEOUT)
+            return await client.ws_connect(stream_url, autoping=True, heartbeat=CONSTANTS.WS_PING_HEARTBEAT)
```
CrimsonJacket, line 80: With the `autoping` set to `True`, we would no longer update the `_last_recv_time` upon receiving a ping message. I think it would be best to set it to `False` since this would affect the polling interval in `Exchange/Derivative` class.

CrimsonJacket (changes requested):
I see a potential issue with with `autoping` set to `True`. Since the client would handle the ping packets automatically, it will not longer update the `_last_recv_time` variable and this could potentially affect the polling interval in the `Exchange/Derivative` class.

hummingbot/connector/exchange/kucoin/kucoin_api_user_stream_data_source.py
petioptrv: Good point. I've added a ping-pong loop and also added a check of `last_recv_time` in the `status_dict` of the exchange as that was the only use for the property as far as I could tell.

CrimsonJacket (approved):
Changes LGTM!

RC-13 (approved):
Re-approving in behalf of QA team

PR #4244, opened 2021-09-22
Refactor/gateway connectors

vic-en:

This PR addresses refactor as stated in https://github.com/CoinAlpha/hummingbot/issues/3306

**Tips for QA testing**:


martinkou (approved):

Nullably (changes requested):
The design was to have gateway and ethereum base classes, so we're not going with this anymore? How does Terra, Avalanche and other protocols fit into this?

mifeng:
> The design was to have gateway and ethereum base classes, so we're not going with this anymore? How does Terra, Avalanche and other protocols fit into this?

The idea is that we only need a single Gateway class because the endpoints to the existing connectors we support in Gateway are all AMMs, so we can abstract all the endpoints down to a common standard trading interface. If we support other exchange designs in the future, we can add other connectors for them.

Nullably (approved):

PR #4257, opened 2021-09-23
feat/handling rate limited commands

petioptrv:

This PR adds handling for network timeouts for the following commands:
    - balance
    - connect
    - create
    - history
    - import
    - start
    - status
It also makes minor changes to the `HummingbotApplication` and `HummingbotCLI` classes to facilitate testing and introduces a new mocking assistant, `CLIMockingAssistant`, to help test CLI related functionality.

- Adds test cases for the majority of the changes.
    - For commands where arriving at network timeout point requires mocking most of the infrastructure surrounding the command, a test case for the proper functionality of the command is included as well. 
    - For commands where there was significant mocking left to be performed before the command's successful execution can be tested, this was not done as it is outside of the scope of this task.

**Tips for QA testing**:
- Rate limits for the commands can be reached by setting the percentage allocation to 1.
- In the case of the `create` command, the rate-limiting happens when we fetch the minimum order amount, in which case, the client should just continue the prompt with a zero minimum amount.
- In all other cases, the command should abort and return control to the user.
- There are also two new global parameters that can be set: `create_command_timeout` and `other_commands_timeout` that dictate the timeout times for the `create` command and all other commands respectively.


hummingbot/client/command/balance_command.py
```diff
@@ -84,7 +86,14 @@ def balance(self,
...
         self._notify("Updating balances, please wait...")
-        all_ex_bals = await UserBalances.instance().all_balances_all_exchanges()
+        network_timeout = float(global_config_map["other_commands_timeout"].value)
+        try:
+            all_ex_bals = await asyncio.wait_for(
+                UserBalances.instance().all_balances_all_exchanges(), network_timeout
+            )
+        except asyncio.TimeoutError:
+            self._notify("\nA network error prevented the balances to update. See logs for more details.")
+            raise
```
CrimsonJacket, line 96: I don't believe we should raise an error here if we're not catching it.

hummingbot/client/command/connect_command.py
```diff
@@ -60,7 +62,17 @@ async def connect_exchange(self,  # type: HummingbotApplication
...
+            try:
+                err_msg = await asyncio.wait_for(
+                    UserBalances.instance().add_exchange(exchange, **api_keys), network_timeout
+                )
+            except asyncio.TimeoutError:
+                self._notify("\nA network error prevented the connection to complete. See logs for more details.")
+                self.placeholder_mode = False
+                self.app.hide_input = False
+                self.app.change_prompt(prompt=">>> ")
+                raise
```
CrimsonJacket, line 75: Similar comment here with raising an error that is not caught.

test/hummingbot/client/command/test_balance_command.py
```diff
@@ -0,0 +1,62 @@
...
+    def setUp(self, _: MagicMock) -> None:
+        super().setUp()
+        self.ev_loop = asyncio.get_event_loop()
+        self.app = HummingbotApplication()
+        self.cli_mock_assistant = CLIMockingAssistant()
+        self.cli_mock_assistant.start()
+
+    def tearDown(self) -> None:
+        super().tearDown()
+        self.cli_mock_assistant.stop()
```
CrimsonJacket, line 22: Correct me if i'm wrong but I believe `super.tearDown()` should be called at the end instead.

test/hummingbot/client/command/test_balance_command.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def async_run_with_timeout_coroutine_must_raise_timeout(self, coroutine: Awaitable, timeout: float = 1):
+        class DesiredError(Exception):
+            pass
+
+        async def run_coro_that_raises(coro: Awaitable):
+            try:
+                await coro
+            except asyncio.TimeoutError:
+                raise DesiredError
```
CrimsonJacket, line 40: Care to explain the necessity of creating a `DesiredError` here?

test/hummingbot/client/command/test_balance_command.py
```diff
@@ -0,0 +1,62 @@
...
+            raise asyncio.TimeoutError
+        except asyncio.TimeoutError:  # the coroutine did not finish on time
+            raise RuntimeError
+
+    @patch("hummingbot.user.user_balances.UserBalances.all_balances_all_exchanges")
+    def test_show_balances_handles_network_timeouts(
+        self, all_balances_all_exchanges_mock
+    ):
+        all_balances_all_exchanges_mock.side_effect = self.raise_timeout
+        global_config_map["other_commands_timeout"].value = 30
```
CrimsonJacket, line 54: I feel that having the `all_balances_all_exchanges` function raise an `asyncio.TimeoutError` would not be ideal. 

I would rather have the timeout be raised by `asyncio.wait_for(...)` function. So I would have set the `other_commands_timeout` value to a small value say `x`, while the `all_balances_all_exchanges()` simply sleeps for `x+1`. Let me know what you think about this.

hummingbot/client/ui/hummingbot_cli.py
```diff
@@ -70,6 +69,8 @@ def __init__(self,
         loop.create_task(start_trade_monitor(self.trade_monitor))
 
     async def run(self):
+        self.app = Application(layout=self.layout, full_screen=True, key_bindings=self.bindings, style=load_style(),
+                               mouse_support=True, clipboard=PyperclipClipboard())
```
CrimsonJacket, line 73: Is there any particular reason for this?

CrimsonJacket (changes requested):
implementation LGTM and good work on including unit tests for various commands. But I have some comments wrt raising of uncaught exceptions as well as some unit tests.

hummingbot/client/command/balance_command.py
petioptrv: My thinking for raising was that the traceback is logged and we can know where things went wrong. For instance, here, a number of different exchanges can get throttled. With the traceback, we'll know which one.

hummingbot/client/ui/hummingbot_cli.py
petioptrv: This is to avoid having to mock the `Application` startup when using the CLI in tests

test/hummingbot/client/command/test_balance_command.py
petioptrv: True. Fixed

test/hummingbot/client/command/test_balance_command.py
petioptrv: Yes, it's because we want to test that the function raises an `asyncio.TimeoutError`, but we also run the functions in an `asyncio.wait_for` which also raises that error. This means that, if a function gets stuck at some point during its execution, `asyncio.wait_for` will raise `asyncio.TimeoutError` but the test case will wrongly pass. So, it's basically to differentiate between the timeout we want and the one we don't.

test/hummingbot/client/command/test_balance_command.py
petioptrv: I like that. Approach changed.

PtrckM:
- the `conf_global_TEMPLATE.yml` version needs to be updated. 
https://github.com/CoinAlpha/hummingbot/blob/0edd2cc4a105ad3092dbc09149c713345573e3ce/hummingbot/templates/conf_global_TEMPLATE.yml#L6

- the `other_commands_timeout` is not configurable on client (expected behavior?)
image

petioptrv:
@PtrckM, both issues addressed in the latest commit

PtrckM:
Bot unable to start because of:
- accepting decimal 
- accepting 0 as value
Issue: https://github.com/CoinAlpha/hummingbot/issues/4282

petioptrv:
Thank you for testing @PtrckM. As mentioned on Discord, the current behaviour is that, if the network checks don't get rate limited, but do fail (i.e. the call returns False), then the bot will silently abort the start command. That being the case, I think it's appropriate that getting rate-limited also results in aborting the startup process. After all, the goal here is to inform the user that they have been rate-limited and to return back control to them.

I have added extra logging in the case where the status check fails without getting rate limited, and I've also excluded zero from the accepted values, but the user can still enter `0.00001`, which I think is fine. We can confidently say that a value of zero doesn't make sense, but other than that, it's up to the user to enter sensible values.

I'll also update the description of the PR to mention that the start command is included in the rate-limited ones.

RC-13 (approved):
Confirming reported issue has been address by the recent commit. LGTM!

CrimsonJacket (approved):
Changes LGTM.

PR #3930, opened 2021-08-13
epic/bybit connector

aarmoa:
**Before submitting this PR, please make sure**:

- [x] Your code builds clean without any errors or warnings
- [x] You are using approved title ("feat/", "fix/", "docs/", "refactor/")

**A description of the changes proposed in the pull request**:
PR to collect all elements required for Bybit connector. This PR is the one that QA will use to test the implementation.
It does not require developers review since those reviews have already been done in each of the individual components' PRs.

**Tests performed by the developer**:

**Tips for QA testing**:

PR #4259, opened 2021-09-23
fix/ Bybit Perpetual websocket request expired error

CrimsonJacket:

**Investigation findings** 
  - timestamp used in WS Auth payload refers to expiration timestamp
  - Issue reported by QA seems to stem from 2 factors
     i. Seems to also affect connector after reverting to before fixes in #4222
     ii. Network Latency. Reason why I came to this conclusion was because I was not getting any request expired errors, while QA is getting it in their local as well as in their VPS instances
     iii. Difference in server time and local(machine) time.
     
**Possible Fixes**
A simple fix would be to increase the expiration timestamp(in this PR)
But a better solution would be to have the `Auth` class sync up with server time via the `GET /v2/public/time` endpoint 

Simply increase the expiration window in `_get_expiration_timestamp()`

**Tips for QA testing**:
All tests to be done in epic PR #3930

### DIFF
--- a/hummingbot/connector/derivative/bybit_perpetual/bybit_perpetual_auth.py
+++ b/hummingbot/connector/derivative/bybit_perpetual/bybit_perpetual_auth.py
@@ -19,7 +19,7 @@ class BybitPerpetualAuth():
         return str(int(time.time() * 1e3))
 
     def get_expiration_timestamp(self):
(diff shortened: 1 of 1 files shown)


aarmoa (approved):
I agree with the 5 seconds expiration (I think it is also the value recommended by Bybit)

Issue #3791, opened 2021-07-26
Rework of Spot-Perpetual Arbitrage logic


phbrgnomo:
## Why

The trading logic of the spot-perpetual strategy needs improvement. As mentioned on #3679, the current logic of the spot-perpetual arbitrage can result in inefficient trade operations being executed.

The strategy currently allows the trade operation to start with a negative price divergence, initiating the trade at a loss, with the expectation of closing the trade on an opposing divergence.

This trading logic isn't optimal, since it adds more risk to the operation because it usually increases the time the trader is positioned until a profitable exit emerges, increasing the risk of liquidation.

## What

Rework the strategy logic so it does not open a position at a loss, hoping it would go up to make a profit. And implement the following suggestions for improvement based on @Nullably's code review:

- Rename `min_divergence` config to `min_opening_arbitrage_pct` to clearly describe what the config does (to open an arbitrage position)
- Rename `min_convergence` to `min_closing_arbitrage_pct`
- Rename `next_arbitrage_cycle_delay` to `next_arbitrage_opening_delay`
- Rename and replace `derivative` with `perpetual` to clearly specify the type of derivative trading
- Arbitrage pct for both `min_opening_arbitrage_pct` and `min_closing_arbitrage_pct` is calculated as (sell price - buy price) / buy_price, thus negative value indicates negative profit expectation.
- `min_opening_arbitrage_pct` should always be positive value in real trading, it also allows negative value for testing purposes (like min_profitability in other arbitrage strategies)
- `min_closing_arbitrage_pct` can be negative value if the user wants to close the position at lesser profit but at a higher chance of closing.
- The `ArbProposal` should be a simple data class (one buy side and one sell side), there should not be any operation on the connectors here, no calling `get_order_price` here, those prices should be supplied at init
- The strategy should generate a list of all possible ArbProposal and use them as appropriate and the list should be refreshed on every tick (or simply not stored as a strategy object member).
- The conditions for opening arb position and closing it should be clear, separated and easy to follow in the code
- Try not to store calculated values as class members, this can lead to uncertainty on the integrity of the values  
- `first arbitrage` and `second arbitrage` on the messages are unclear here, they are for the same arbitrage position, one is to open and the other is to close it
- Remove `max_funding_rate` configuration parameter for now to get the basics of the strategy right first

## Definition of done
- [ ] Config parameters change as described
[...]

Issue #3679, opened 2021-07-02
Spot Perpetual Arbitrage - Multiple bugs summary


Nullably:
```
INFO:hummingbot.strategy.spot_perpetual_arbitrage.spot_perpetual_arbitrage:Markets are ready. Trading started.
INFO:hummingbot.strategy.spot_perpetual_arbitrage.spot_perpetual_arbitrage:Current spread: 6.32%, minimum divergence spread: 5.00%.
INFO:hummingbot.strategy.spot_perpetual_arbitrage.spot_perpetual_arbitrage:Spread wide enough to execute first arbitrage!: 
Backtestmarket:  Buy  1.000000 HBOT at 101.000 ETH. 
Mockperpconnector:  Sell  1.00000 HBOT at 95.000 ETH. 

INFO:hummingbot.strategy.spot_perpetual_arbitrage.spot_perpetual_arbitrage:Placing BUY order for 1.000000 HBOT at BacktestMarket at 101.000 price [clock=2019-01-01 00:00:01+00:00]
INFO:hummingbot.strategy.spot_perpetual_arbitrage.spot_perpetual_arbitrage:Placing SELL order for 1.00000 HBOT at MockPerpConnector at 95.000 price to OPEN position. [clock=2019-01-01 00:00:01+00:00]
INFO:hummingbot.strategy.spot_perpetual_arbitrage.spot_perpetual_arbitrage:Current spread: 6.32%, minimum convergence spread: 2.00%.
INFO:hummingbot.strategy.spot_perpetual_arbitrage.spot_perpetual_arbitrage:Current spread: 1.00%, minimum convergence spread: 2.00%.
INFO:hummingbot.strategy.spot_perpetual_arbitrage.spot_perpetual_arbitrage:Spread low enough to execute second arbitrage!: 
Backtestmarket:  Buy  1.000000 HBOT at 101.000 ETH. 
Mockperpconnector:  Sell  1.00000 HBOT at 100.000 ETH. 

INFO:hummingbot.strategy.spot_perpetual_arbitrage.spot_perpetual_arbitrage:Placing BUY order for 1.000000 HBOT at BacktestMarket at 101.000 price [clock=2019-01-01 00:00:07+00:00]
[...]

PR #4264, opened 2021-09-24
Refactor/spot perp arb rework

Nullably:

- Implement changes as described in #3791 
- Fixes bugs in #3679 

- Unit tests
- Strategy test on real trading on ETH-USDT on Binance and Binance Perpetual
- Configuration process test (create, import)

**Tips for QA testing**:
- You can set min_opening_arbitrage_pct to negative value for the strategy to immediate open arbitrage position, though you will open the position at negative profit expectation. Similar to other arbitrage strategy, this is mainly intended for testing purpose.


RC-13:
@Nullably When I was testing the PR, the strategy was able to recognized that the Binance test account I was using is in Hedge mode (which is a good thing) however it seems that it permanently save that status and doesn't recognize that I  already switched the position to one-way mode. I tried running a perpetual market making strategy using the same account in the same PR and it was able to recognize that the account is already in one-way mode.

https://user-images.githubusercontent.com/57189990/134636384-caccb091-4908-4251-9d5b-1168a71118f4.mov

.
Cross check on development branch and it seems that the behavior happens in the PR only.
image

Using other perpetual connector, the strategy works as intended.

Nullably:
> @Nullably When I was testing the PR, the strategy was able to recognized that the Binance test account I was using is in Hedge mode (which is a good thing) however it seems that it permanently save that status and doesn't recognize that I already switched the position to one-way mode. I tried running a perpetual market making strategy using the same account in the same PR and it was able to recognize that the account is already in one-way mode.
> 
>  spot-perp.vs.perp-mm.mov 
> .
> Cross check on development branch and it seems that the behavior happens in the PR only.
> image
> 
> Using other perpetual connector, the strategy works as intended.

The perpetual market making set position mode for a user every time the strategy starts. 
Looks like there is a bug in Binance perpetual connector where it doesn't get user position mode at the start, so the spot perp arb strategy doesn't get this information.

dennisocana:
I think that's related to the BInance Perpetual issue here: https://github.com/CoinAlpha/hummingbot/issues/3961

RC-13:
Thank you @Nullably for the explanation.

RC-13 (approved):
LGTM!

mhrvth (approved):

PR #4266, opened 2021-09-24
Feat/allow avellaneda wait cancel confirmation

### LINKED ISSUE #4209: Improve PMM and Avellaneda to delay new orders creation until current orders are cancelled
## Why
As a user of the strategies I want them to create the full set of all expected Buy and Sell orders in each strategy cycle. If the logic that creates the new set of orders is executed before the cancellation of the previous stack of orders has been completed, the balance of the connector will not be entirely available and hence the new set of orders will be incomplete (only the orders for which there is enough balance will be created). Even when at same point the previous stack of orders' cancellation will be completely done, the current stack of orders will be complete for the entire cycle.

## What
Due to the close relationship between balance status and the logic that create orders in the strategies, the step that creates the new set of order should run only after all orders from the previous stack have been cancelled (or at least until we have made the effort of cancelling them).
Since the logic is asynchronic, we need to cover the following scenarios:
- The order refresh cycle for RC1 happened, but order refresh cycle RC2 arrived without RC1 finished. RC1 should be canceled in favor of RC2.
- The tick processing should never be blocked.

## DoD
- [ ] Strategies delay new orders creation until the orders of the corresponding stack cycle are cancelled.
- [ ] Strategies are prepared to handle the situation of long cancellation delays, even when the new refresh cycle is reached.
- [ ] Tick execution is never stopped.

aarmoa:

Added functionality in Avellaneda strategy to wait for orders cancellation confirmation before creating the new set of orders during the refresh time. The new functionality is configurable with a new parameter (configuration) in the strategy called `should_wait_order_cancel_confirmation` for the user to decide if they want to use it or not. By default the value of the configuration is `True` and makes the strategy wait for cancellation confirmation.

New unit tests.
Tested running the bot connected to Gateio exchange, and using 3 order levels (causing the strategy to consume all available balance in each cycle).

**Tips for QA testing**:
To check if the strategy is working correctly with the changes I suggest running it connected to a exchange that does not provide constant balance updates (like Gateio).
I also suggest to check the strategy with and without using hanging orders.

Fix #4209 (for Avellaneda)

### DIFF
--- a/hummingbot/strategy/order_tracker.pyx
+++ b/hummingbot/strategy/order_tracker.pyx
@@ -1,3 +1,4 @@
+import asyncio
 from collections import (
     deque,
     OrderedDict
@@ -22,7 +23,6 @@ cdef class OrderTracker(TimeIterator):
     # ETH confirmation requirement of Binance has shortened to 12 blocks as of 7/15/2019.
     # 12 * 15 / 60 = 3 minutes
     SHADOW_MAKER_ORDER_KEEP_ALIVE_DURATION = 60.0 * 3
-
     CANCEL_EXPIRY_DURATION = 60.0
 
     def __init__(self):
@@ -112,7 +112,7 @@ cdef class OrderTracker(TimeIterator):
         return pd.DataFrame(data=market_orders, columns=["market", "trading_pair", "order_id", "quantity", "timestamp"])
 
     @property
-    def in_flight_cancels(self) -> Dict[str, float]:
+    def in_flight_cancels(self) -> Dict[str, Tuple[float, asyncio.Event]]:
         return self._in_flight_cancels
 
     @property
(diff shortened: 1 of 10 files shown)


hummingbot/strategy/order_tracker.pyx
```diff
@@ -307,3 +310,9 @@ cdef class OrderTracker(TimeIterator):
 
     def remove_create_order_pending(self, order_id: str):
         self.c_remove_create_order_pending(order_id)
+
+    async def all_orders_being_cancelled_confirmation(self, market_info: MarketTradingPairTuple):
```
Nullably, line 314: Can this be done by simply checking if the in_flight_cancels list is empty?
These wait tasks can make it difficult to debug and at least it should have time out accompanied.

aarmoa:
This approach is incorrect because we should not use anything asyncio related inside strategies

PR #4281, opened 2021-09-24
feat/allow pmm and avellaneda wait cancel confirmation

### LINKED ISSUE #3686: Gate.io - Inconsistent order creation due to balance latency
**Describe the bug**
There are random occurrences where orders are not created at certain cycles when running the pure market making strategy using Gate.io

**Steps To Reproduce**
1. Run a pure market making strategy using Gate.io
2. Run status --live command to monitor the orders being created by the bot

**Screenshots**

On multiple order levels (with the current config the bot should be able to place 5 buy orders and 4 sell orders)
Screen Shot 2021-07-04 at 10 32 34 AM
Screen Shot 2021-07-04 at 10 36 02 AM

On single order level
Screen Shot 2021-07-04 at 11 52 39 AM

**Release version**
dev-0.41.0

**Attachments**
[gate_io.zip](https://github.com/CoinAlpha/hummingbot/files/6759352/gate_io.zip)

### LINKED ISSUE #4087: Gate.io: Invalid orders tracked on status window and failed to get canceled
**Describe the bug**
When running any market making strategy and the current asset cannot cover placing orders at both side at the same time. The bot still attempt to create an order that is below the exchange minimum allowable trade size - this is normal, however this invalid orders are displayed in the status window when you run the status command which is not happening on other exchange. If bot is using a fast refresh rate and multiple level, at some point those non-existing order will fail to be canceled and will spam the status window with invalid orders.
Despite getting this issue, bot still places and cancels valid orders.

**Steps To Reproduce**
1. Create any market making strategy, for my case I use Avellaneda MM and use a market pair where your asset can place a one sided order or won't be able to create all levels.
2. Set order refresh time to 30 seconds or lower.
3. Configure multiple level, in my case, it is set to 3 levels.
4. Set inventory_pct to 50 % or adjust if necessary.
5. Start the strategy and leave the bot running for a while. 
...

**Screenshots**
image
After leaving the bot to run overnight, multiple non-exiting order is now displayed in the status window. This also fill the logs (1.9GB for my case)
gate_io-failed-to-cancel

**Release version**
0.42.0

**Attachments**
Sample logs:
[logs_ave-eth-btc-gate.log.zip](https://github.com/CoinAlpha/hummingbot/files/7103615/logs_ave-eth-btc-gate.log.zip)

### LINKED ISSUE #4209: Improve PMM and Avellaneda to delay new orders creation until current orders are cancelled

aarmoa:

Added functionality in Avellaneda strategy to wait for orders cancellation confirmation before creating the new set of orders during the refresh time. The new functionality is configurable with a new parameter (configuration) in the strategy called should_wait_order_cancel_confirmation for the user to decide if they want to use it or not. By default the value of the configuration is True and makes the strategy wait for cancellation confirmation.

New unit tests.
Tested running the bot connected to Gateio exchange, and using 3 order levels (causing the strategy to consume all available balance in each cycle).

**Tips for QA testing**:
To check if the strategy is working correctly with the changes I suggest running it connected to a exchange that does not provide constant balance updates (like Gateio).
I also suggest to check the strategy with and without using hanging orders.

Fix #3686
Fix #4087  
Fix #4209

### DIFF
--- a/hummingbot/strategy/avellaneda_market_making/avellaneda_market_making.pxd
+++ b/hummingbot/strategy/avellaneda_market_making/avellaneda_market_making.pxd
@@ -55,6 +55,7 @@ cdef class AvellanedaMarketMakingStrategy(StrategyBase):
         object _latest_parameter_calculation_vol
         str _debug_csv_path
         object _avg_vol
+        bint _should_wait_order_cancel_confirmation
 
     cdef object c_get_mid_price(self)
     cdef _create_proposal_based_on_order_override(self)
--- a/hummingbot/strategy/avellaneda_market_making/avellaneda_market_making.pyx
+++ b/hummingbot/strategy/avellaneda_market_making/avellaneda_market_making.pyx
@@ -92,6 +92,7 @@ cdef class AvellanedaMarketMakingStrategy(StrategyBase):
                     closing_time: Decimal = Decimal("1"),
                     debug_csv_path: str = '',
                     volatility_buffer_size: int = 30,
+                    should_wait_order_cancel_confirmation = True,
                     is_debug: bool = False,
                     ):
         self._sb_order_tracker = OrderTracker()
@@ -142,6 +143,7 @@ cdef class AvellanedaMarketMakingStrategy(StrategyBase):
         self._optimal_spread = s_decimal_zero
         self._optimal_ask = s_decimal_zero
         self._optimal_bid = s_decimal_zero
+        self._should_wait_order_cancel_confirmation = should_wait_order_cancel_confirmation
         self._debug_csv_path = debug_csv_path
         self._is_debug = is_debug
         try:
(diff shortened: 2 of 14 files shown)


rxlxrxsx:
- There are instances that after an order is filled, it fails to create a hanging order on the next order cycle #4284
- On pure_mm strategy, the opposing order is canceled even before the refresh rate period kicks in - #4285

Nullably (approved):
LGTM

aarmoa:
I have tried to reproduce the issue with other connector (Kraken) and it was not possible (I tried both with PMM and Avellaneda strategy). I also reviewed PR changes and it does not change the logic to determine if an active order in the strategy is a hanging order. That makes me believe that either this is a problem with hanging orders tracker in general or a problem with the Gateio connector.
Since this fix was created also for the scenarios when the strategies run without using the hanging orders tracker, I will merge it and I will continue investigating the reported issues independently.

PR #4310, opened 2021-09-28
Feat/gateway v2 route validators

james-hummingbot:
Validate all request bodies before performing the route controller.

### DIFF
--- a/gateway/src/chains/ethereum/ethereum.validators.ts
+++ b/gateway/src/chains/ethereum/ethereum.validators.ts
@@ -7,26 +7,6 @@ import {
   Validator,
 } from '../../services/validators';
 
-// import {
-//   EthereumNonceRequest,
-//   EthereumAllowancesRequest,
-//   EthereumBalanceRequest,
-//   EthereumApproveRequest,
-//   EthereumPollRequest,
-// } from './ethereum.requests';
-
-// validate request parameters
-
-// test if a string matches the shape of an Ethereum public key
-export const isPublicKey = (str: string): boolean => {
-  return /^0x[a-fA-F0-9]{40}$/.test(str);
-};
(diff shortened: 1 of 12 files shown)


gateway/src/chains/ethereum/ethereum.validators.ts
```diff
@@ -0,0 +1,149 @@
...
+export const validateEthereumAllowancesRequest = (
+  req: EthereumAllowancesRequest
+): void => {
+  const errors = validatePrivateKey(req)
+    .concat(validateSpender(req))
+    .concat(validateTokenSymbols(req));
+  throwErrorsIfExist(errors);
+};
+
+export const validateEthereumBalanceRequest = (
```
martinkou, line 128: There are really two types of validator functions here - the first type are the validator "components" that output a list of error strings. The other type are the request validators which raises an error if any of the components output an error.

We should declare the function types of these two validator types, for better documentation. e.g.

```
type Validator = (req: any) => Array<string>;
type RequestValidator = (req: any) => void;
```

Also, a lot of the request validator functions are really the same - as in they're concatenating the results from multiple validators. It looks like we can just turn that into a factory method for the common case.

martinkou (changes requested):

gateway/src/chains/ethereum/ethereum.validators.ts
james-hummingbot: Thanks, those are great suggestions. I've added them and renamed the `validate` function to `mkValidator` since that makes it clear what it is doing. It creates a validating function, it doesn't validate by itself. Then i did the same for `RequestValidator` and made `mkRequestValidator`.

martinkou (approved):

PR #4370, opened 2021-10-04
Fix/quantization round to smallest value

### LINKED ISSUE #3909: AscendEX CRO order precision API error
**Describe the bug**
// A clear and concise description of what the bug is.
CRO on AscendEX should have no decimals but connector sends to the AscendEX API that has decimals. 

**Steps To Reproduce**
// A concise description of the **steps to reproduce the buggy behavior**:
Use a Hummingbot trade on AscendEX with a decimal. 

Arbitrage CRO-USDT

**Screenshots**
// If applicable, add screenshots to help explain your problem.
https://i.imgur.com/OJhrkn0.png
**Release version**
// Include your bot's version number (Can be found at the upper left corner of your CLI).
v0.41.0

**Attachments**
// Please attach your config file and log file to help us debug. It would be difficult for us to help you without those! 
On request. I think the issue is clear with the order precision. 

@Poroburu#1024

aarmoa:

Made a change in the quantization logic to use `math.floor` to round instead of the native `round` Python function.

Added a test case to verify the border scenarios

**Tips for QA testing**:
The best way to test the fix is to operate with non integer amount values in exchanges and for tokens that only allow integer amounts (check the bug this PR solves)

Fixes #3909


rxlxrxsx:
Minor cosmetic issue with values shown on Amount (Orig) not appearing. This only happens on pure_mm strategy.

Screen Shot 2021-10-05 at 10 27 14 AM

Nullably:
`connector_base.pyx` also uses `round` for rounding on `c_quantize_order_price`, should we fix there too?

rxlxrxsx (approved):
Tested and confirmed working by QA. The Issue reported earlier was found out to be happening on development branch as well.

aarmoa:
I have implemented the same change for quantization in ConnectorBase as suggested by Jack. Please review the new changes and test the PR again.

leastchaos:
Thank you for this fix, I have been having this issue quite long and this fixes it.
I want to mention that flooring order price seems unnecessary and may introduce issue.
e.g if i set a spread at 0.1% for token with 2dp (0.01) at price 9.9. the price calculated is 9.9099 this means for sell the price will be rounded down to 9.9 resulting in a spread of 0% instead of current calculated price of 9.91?.  This is extreme case but hope you get the point.
In real application it might not be affecting much since usually the order price is at a much higher precision, but edge case like mentioned might have issue?

aarmoa:
Hello @leastchaos. Thank you for your comment. I do understand your point. The need to round price also to a smaller value is to address the buy scenario. If we just round to the closest value, and it ends up being the bigger value, that means that we could be ending up trying to place an order with a price higher than the current account balance.
We might need to look for a solution that applies different rounding strategies for buy and sell, base token or quote token (and their combinations), but until we have that functionality I believe it would be better to ensure we never try to spend more than the current balance.

PR #4386, opened 2021-10-06
fix/Kucoin filled order logged twice

petioptrv:

The bug arises when the order if filled during the execution of `_update_order_status`. If timed correctly, both the `_update_order_status` function and the user stream listener loop fire the events.

- Test cases covering the issue in the `_update_order_status` function.

**Tips for QA testing**:
I haven't tested this, but one way to attempt to trigger the problem is to set the `SHORT_POLL_INTERVAL` and `LONG_POLL_INTERVAL` values of the `KucoinExchange` class to a small value (e.g. 1 second) in an attempt to have it coincide with the order fill.


PtrckM:
Crosscheck releases from latest vs development and `fix/kucoin_filled_order_logged_twice` we're unable to replicate the issue. All test passed with flying colors.

PR #4461, opened 2021-10-13
Feat/custom color configuration

mhrvth:

Added configuration variables to customize colors of the client.

Ran the client and changed values of all the new config variables, observed changes in the interface. 
Tried to enter invalid values. Tried to enter upper case and lower case values.
Tested persistence of settings by turning it off and on again.

**Tips for QA testing**:

Test all the new config variables, compare the result with the intended result described in the issue. 
Test on Windows especially where hex colors are not available, only ANSI colors.
Test in different environments and different terminals, the results should be everywhere the same except on Windows.

### DIFF
--- a/hummingbot/client/ui/hummingbot_cli.py
+++ b/hummingbot/client/ui/hummingbot_cli.py
@@ -48,7 +48,7 @@ class HummingbotCLI:
         self.timer = create_timer()
         self.process_usage = create_process_monitor()
         self.trade_monitor = create_trade_monitor()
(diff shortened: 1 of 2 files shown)


mhrvth:
On Windows it merely replicates the existing state where background color is not defined in the style, it's the default black. I think it can be a good idea to implement defining the background though, I'm just not sure if it will work and how will it look.

Adding a reset option is possible.

I don't think the client should automatically adjust and override any style parameters as it can be a frustrating experience for users when the settings don't remain as they chose them to be.

RC-13:
@mhrvth Re-tested PR on windows and color change are now working including the reset command. Before we approve I think the template version needs updating.

PR #4641, opened 2021-11-02
(refactor) Introduces the `api_delegate` package.

petioptrv:
The PR is quite massive, but a good portion of that is just imports sorting.

* [x] Your code builds clean without any errors or warnings
* [x] You are using approved title ("feat/", "fix/", "docs/", "refactor/")

* Introduces the basic infrastructure for abstracting away REST and WS API logic from the main connector classes.
* Refactors the Gate.io connector to use the new classes.
* Runs `isort -m=3 -l=120 .` in the `gate_io` and `api_delegate` directories to sort the imports.

As proposed by @aarmoa, I will make a couple of diagrams to help the community understand the way these classes are meant to be used, which is to abstract away and break-down the logic around making WS and REST API calls that currently lives inside the main classes of the connectors. I've left here and there code that may not be extremely useful atm, but is with the idea of eventually pushing more functionality into these classes, like the `throttler_limit_id` for requests, and the more invasive pre- and post-processor logic. Here are some spin-off tasks that I can create should this PR be approved:

* Implement an Auth pre-processor
  * Ideally, for an exchange that requires making API calls to renew the authentication
* Push throttler functionality into the connector classes
* Implement an error-handling post-processor
* Push re-connection logic inside WSDelegate
  * It should store the `subscribe` messages that the user has passed and re-run them when it re-establishes the new connection
* Implement a mock connector for testing (?)
  * Ideally, one that only requires us to replace the `@aioreponses()` decorator with a new one and not change anything else about the test cases
  * This will allow us to be truly independent from any REST and WS library

* Introduces tests for the new classes.
* Ensures Gate.io tests pass.

**Tips for QA testing**:
Please hold off on testing until dev approval for this one.
* Ensure that Gate.io functions as expected (anything that has to do with contacting the Gate.io API):
  * Orders are placed and cancelled
  * Orderbook is updated
  * Order fills are received quickly (though websocket)

┆Issue is synchronized with this [Clickup task](https://app.clickup.com/t/1u9eupe) by [Unito](https://www.unito.io)


hummingbot/connector/exchange/gate_io/gate_io_api_order_book_data_source.py
```diff
@@ -56,29 +61,43 @@ def _get_throttler_instance(cls) -> AsyncThrottler:
         return throttler
 
     @classmethod
-    async def get_last_traded_prices(
-        cls, trading_pairs: List[str], throttler: Optional[AsyncThrottler] = None
-    ) -> Dict[str, Decimal]:
-        throttler = throttler or cls._get_throttler_instance()
+    async def get_last_traded_prices(cls, trading_pairs: List[str]) -> Dict[str, Decimal]:
+        throttler = cls._get_throttler_instance()
+        rest_delegate = await build_gate_io_api_factory().get_rest_delegate()
```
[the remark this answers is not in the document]
petioptrv: That's the case for `get_order_book_data`, but only because that one is called from within the `GateIoAPIOrderBookDataSource` class and it then passes the `self.` of those parameters. However, this method is only called form `hummingbot/core/utils/market_price.py:27` and `hummingbot/core/data_type/order_book_tracker.py:161`. And in both cases, the generic signature of the method is used and those parameters are not passed, so I omitted them to be more explicit

aarmoa (changes requested):
The only change I think we need before approving the PR is consistently renaming the delegate methods and variables.
Besides that there is something that we might need to add as a separate task: the possibility of using the WSConnection as a context (maybe for us the one providing the context would be WSAssistant). There are many connectors that currently use the websocket context as a way to ensure the connection is properly closed when the context finishes running. But I would not do this in the context of the current PR.

aarmoa (dismissed):
I agree with the rename from `delegate` to `assistant`, and from `api` to `web`.

RC-13:
Reposting issue seen while running the PR for long period. Unexpected clock tick error randomly occurring, sometimes leads to the bot no longer creating order but there are instance that the bot recover but then gets the issue again.
```python
2021-11-08 04:33:52,346 - 2922 - hummingbot.core.clock - ERROR - Unexpected error running clock tick.
  File "hummingbot/connector/connector_base.pyx", line 218, in hummingbot.connector.connector_base.ConnectorBase.c_tick
    self.tick(timestamp)
  File "/home/ubuntu/gate_io/hummingbot/connector/exchange/gate_io/gate_io_exchange.py", line 879, in tick
    if now - self._user_stream_tracker.last_recv_time > 120.0
  File "/home/ubuntu/gate_io/hummingbot/core/data_type/user_stream_tracker.py", line 40, in last_recv_time
    return self.data_source.last_recv_time
  File "/home/ubuntu/gate_io/hummingbot/connector/exchange/gate_io/gate_io_api_user_stream_data_source.py", line 44, in last_recv_time
    recv_time = self._ws.last_recv_time
  File "/home/ubuntu/gate_io/hummingbot/connector/exchange/gate_io/gate_io_websocket.py", line 37, in last_recv_time
    return self._ws_assistant.last_recv_time
AttributeError: 'NoneType' object has no attribute 'last_recv_time'
```

hummingbot/connector/exchange/gate_io/gate_io_websocket.py
```diff
@@ -34,7 +34,9 @@ def __init__(self,
 
     @property
     def last_recv_time(self) -> float:
-        return self._ws_assistant.last_recv_time
+        if self._ws_assistant is not None:
+            self._last_recv_time = self._ws_assistant.last_recv_time
+        return self._last_recv_time
```
aarmoa, line 39: It seems like the local variable `_last_recv_time` is only used for this. I think it would be better if we just remove it and return 0 when the _ws_assistant is not yet initialized.

petioptrv: I didn't go for that approach because it will change the current behaviour, i.e. it will cause the exchange to use the short poll interval for the polling loop, but given that we're likely reconnecting when this happens, I think that's fine

RC-13 (approved):
Confirming that the previous reported issue did not appear anymore on the test bot we are running for more than 24 hours now. LGTM!

aarmoa (approved):
Changes look good to me

PR #4678, opened 2021-11-09
refactor/order collateral token

petioptrv:

This PR introduces the functionality proposed by @Nullably to allow more accurate sizing of the collateral required for perpetual orders. It adopts the approach, proposed by @aarmoa, of implementing a dedicated component that the exchange holds a reference to and the strategies use, as an attempt to avoid the need to introduce concepts specific to perpetual trading in the `ExchangeBase` class.

- Tests for the new components.
- Tests for the modified parts of the perpetual exchanges.
- Tests for the relevant part of the perpetual mm strategy.
- All tests for spot perp arb pass.

**Tips for QA testing**:
This will be a tricky one. Essentially, this PR modifies the perpetual mm strategy, the spot perp arb strategy, and all of the spot connectors (binance, bybit, dydx, perpetual finance / gateway) in the parts of the code that handle adjusting order proposals based on available budget.

To see that the functionality is as expected, the test must make the strategies try to place orders that they don't have sufficient balance for and some of those orders must get rejected by the strategy (not by the exchange), while at the same time, the strategy should use the max balance available (i.e. should not reject an order if there is sufficient balance for it).

In addition, the sizing must consider fees. To test that, you override the fee to something large and observe an order be rejected.

The sizing should also consider leverage (i.e. open a position for which there is not enough real collateral, but there is enough when taking into consideration leverage).

The intricacies of the sizing don't need to be tested on each exchange. That logic lives elsewhere, so as long as it works on one exchange, and the other exchanges work for a small subset of those, this should be sufficient.

### DIFF
--- a/hummingbot/connector/budget_checker.py
+++ b/hummingbot/connector/budget_checker.py
@@ -1,13 +1,15 @@
+import typing
 from collections import defaultdict
+from copy import copy
 from dataclasses import dataclass
 from decimal import Decimal
 from typing import Dict, List, Optional
-from copy import copy
 
-from hummingbot.connector import exchange_base
-from hummingbot.connector.perpetual_trading import PerpetualTrading
 from hummingbot.connector.utils import split_hb_trading_pair
-from hummingbot.core.event.events import OrderType, TradeType
+from hummingbot.core.event.events import OrderType, TradeType, TradeFee
+
+if typing.TYPE_CHECKING:  # avoid circular import problems
+    from hummingbot.connector.exchange_base import ExchangeBase
 
 
 @dataclass
@@ -23,7 +25,7 @@ class OrderCandidate:
 
 
 class BudgetChecker:
-    def __init__(self, exchange: exchange_base.ExchangeBase):
+    def __init__(self, exchange: "ExchangeBase"):
         """
         Provides utilities for strategies to check if the required assets for trades are available.
 
@@ -35,6 +37,10 @@ class BudgetChecker:
         amount needed for that order and makes it unavailable for the following hypothetical orders.
         Once the orders are sent to the exchange, the strategy must call `reset_locked_collateral` to
         free the hypothetically locked assets for the next set of checks.
+
+        The default implementation covers the general case, with the exception of `_get_collateral_token`
+        and `_get_fee` (and perhaps `_get_collateral_adjustment_for_fees`) methods which are spot-specific
+        and may need to be overridden for the specific exchange configuration.
 
         :param exchange: The exchange against which available collateral assets will be checked.
         """
@@ -79,13 +85,15 @@ class BudgetChecker:
         See the doc string for `adjust_candidate` to learn more about how the adjusted order
         amount is derived.
 
+        This method also locks in the collateral amount for the given collateral token and makes
+        it unavailable on subsequent calls to this method until the `reset_locked_collateral`
+        method is called.
+
         :param order_candidate: The candidate order to check and adjust.
         :param all_or_none: Should the order amount be set to zero on insufficient balance.
         :return: The adjusted order candidate.
         """
-        adjusted_candidate = self.adjust_candidate(
-            order_candidate, all_or_none
-        )
+        adjusted_candidate = self.adjust_candidate(order_candidate, all_or_none)
         self._lock_available_collateral(adjusted_candidate)
         return adjusted_candidate
 
@@ -106,12 +114,17 @@ class BudgetChecker:
         """
         adjusted_candidate = self.populate_collateral_fields(order_candidate)
         available_collateral = self._exchange.get_available_balance(adjusted_candidate.collateral_token)
-        if adjusted_candidate.collateral_amount < available_collateral:
+        locked_collateral = self._locked_collateral[adjusted_candidate.collateral_token]
+        available_collateral -= locked_collateral
+        required_collateral_amount = adjusted_candidate.collateral_amount
+
+        if available_collateral < required_collateral_amount:
             if all_or_none:
                 adjusted_candidate.amount = Decimal("0")
                 adjusted_candidate.collateral_amount = Decimal("0")
             else:
-                adjusted_candidate = self._adjust_amounts(adjusted_candidate, available_collateral)
+                adjusted_candidate = self._reduce_order_and_collateral_amounts(adjusted_candidate, available_collateral)
+
         return adjusted_candidate
 
     def populate_collateral_fields(self, order_candidate: OrderCandidate) -> OrderCandidate:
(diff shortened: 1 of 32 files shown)


hummingbot/connector/exchange_base.pyx
```diff
@@ -283,3 +281,43 @@ cdef class ExchangeBase(ConnectorBase):
...
+    async def get_required_collateral(
+        self,
+        trading_pair: str,
+        order_type: OrderType,
+        order_side: TradeType,
+        amount: Decimal, price: Decimal,
+        leverage: Decimal = Decimal("1"),
+    ) -> Tuple[Decimal, str]:
+        """
+        Returns the required collateral value and the collateral token for the given trade.
```
aarmoa, line 294:
```suggestion
        Returns the required collateral value and the collateral token for the given order.
```

hummingbot/connector/exchange_base.pyx
```diff
@@ -283,3 +281,43 @@ cdef class ExchangeBase(ConnectorBase):
         required volume.
         """
         return Decimal(str(self.get_price_for_volume(trading_pair, is_buy, amount).result_price))
+
+    async def get_required_collateral(
```
aarmoa, line 285: If we make the function receive the tokens individually (base and quote) and also the fee as parameters, we could move it to the `PerpetualTrading` class.

aarmoa (commented):
I have added a comment we can discuss

hummingbot/connector/exchange_base.pyx
petioptrv: That was my original approach, but by placing the function in the `ExchangeBase`, we have a unified way for perpetual as well as spot exchanges to calculate the true size of a proposed position

hummingbot/connector/budget_checker.py
```diff
@@ -0,0 +1,290 @@
+from collections import defaultdict
+from dataclasses import dataclass
+from decimal import Decimal
+from typing import Dict, List, Optional
+from copy import copy
```
aarmoa, line 5: This import statement is not correctly sorted

hummingbot/connector/budget_checker.py
```diff
@@ -0,0 +1,290 @@
...
+
+        :param order_candidate: The candidate order to check and adjust.
+        :return: The adjusted order candidate.
+        """
+        trading_pair = order_candidate.trading_pair
+
+        base, quote = split_hb_trading_pair(trading_pair)
+        collateral_token = quote if TradeType.BUY else base
+
+        required_collateral = order_candidate.amount * order_candidate.price
```
aarmoa, line 133: Shouldn't this be different for buys and sells? If I want to buy 2 BTC with order price 60000 USDT, that means the collateral for the order is 120000 USDT. But if I want to sell 2 BTC with order price 60000 USDT the collateral for the order is 2 BTC, not 2 BTC * 60000 USDT/BTC

hummingbot/connector/budget_checker.py
```diff
@@ -0,0 +1,290 @@
...
+                order_candidate.order_side,
+                order_candidate.amount,
+                order_candidate.price,
+            )
+            adjusted_amount = fee.order_amount_from_quote_with_fee(
+                trading_pair, order_candidate.price, available_collateral
+            )
+            adjusted_amount = self._exchange.quantize_order_amount(trading_pair, adjusted_amount)
+        else:
+            adjusted_amount = available_collateral / order_candidate.price
```
aarmoa, line 203: I believe we always need to quantize the amount, both for sells and buys.

hummingbot/connector/budget_checker.py
```diff
@@ -0,0 +1,290 @@
...
+
+        adjusted_candidate = copy(order_candidate)
+        adjusted_candidate.amount = adjusted_amount
+        adjusted_candidate.collateral_amount = available_collateral
+
+        return adjusted_candidate
+
+
+class PerpetualBudgetChecker(BudgetChecker):
```
aarmoa, line 212: It might be better to move the PerpetualBudgetChecker to its own module, and place it inside the perpetual connectors folder.

hummingbot/connector/budget_checker.py
```diff
@@ -0,0 +1,290 @@
...
+
+    def populate_collateral_fields(self, order_candidate: OrderCandidate) -> OrderCandidate:
+        trading_pair = order_candidate.trading_pair
+
+        if order_candidate.order_side == TradeType.BUY:
+            collateral_token = await self._exchange.get_buy_collateral_token(trading_pair)
+        else:
+            collateral_token = await self._exchange.get_sell_collateral_token(trading_pair)
+
+        order_size = order_candidate.amount * order_candidate.price
```
aarmoa, line 230: The size to consider for the collateral will depend on the order type and the token the collateral should be provided in. It will not always be the same calculation.

aarmoa (changes requested):

hummingbot/connector/derivative/binance_perpetual/binance_perpetual_derivative.py
```diff
@@ -1,48 +1,54 @@
...
 from enum import Enum
+from typing import Any, AsyncIterable, Dict, List, Optional
 from urllib.parse import urlencode
-from typing import Optional, List, Dict, Any, AsyncIterable
 
-from hummingbot.connector.derivative.binance_perpetual.binance_perpetual_in_flight_order import BinancePerpetualsInFlightOrder
-from hummingbot.connector.derivative.binance_perpetual.binance_perpetual_order_book_tracker import BinancePerpetualOrderBookTracker
-from hummingbot.connector.derivative.binance_perpetual.binance_perpetual_user_stream_tracker import BinancePerpetualUserStreamTracker
+import aiohttp
+from async_timeout import timeout
```
aarmoa, line 13: These two dependencies are not in the correct order

hummingbot/connector/derivative/bybit_perpetual/bybit_perpetual_derivative.py
```diff
@@ -1,41 +1,41 @@
...
 import logging
 import math
-import pandas as pd
 import time
+from decimal import Decimal
+from typing import Any, AsyncIterable, Dict, List, Optional
+
+import aiohttp
+import pandas as pd
 import ujson
```
aarmoa, line 11: These three import statements are out of order

hummingbot/connector/exchange/paper_trade/paper_trade_exchange.pyx
```diff
@@ -168,7 +170,8 @@ cdef class PaperTradeExchange(ExchangeBase):
     def __init__(self, order_book_tracker: OrderBookTracker, config: MarketConfig, target_market: type):
         order_book_tracker.data_source.order_book_create_function = lambda: CompositeOrderBook()
         self._order_book_tracker = order_book_tracker
-        super(ExchangeBase, self).__init__()
+        self._budget_checker = BudgetChecker(exchange=self)
+        super(ExchangeBase, self).__init__()  # super(ExchangeBase, self).__init__() does not work
```
aarmoa, line 174: The comment says that the code line does not work. Is that still the case?

hummingbot/strategy/spot_perpetual_arbitrage/spot_perpetual_arbitrage.py
```diff
@@ -292,39 +293,79 @@ def check_budget_constraint(self, proposal: ArbProposal) -> bool:
...
-        spot_token = spot_side.market_info.quote_asset if spot_side.is_buy else spot_side.market_info.base_asset
-        spot_avai_bal = spot_side.market_info.market.get_available_balance(spot_token)
-        if spot_side.is_buy:
-            fee = spot_side.market_info.market.get_fee(
-                spot_side.market_info.base_asset,
-                spot_side.market_info.quote_asset, OrderType.LIMIT, TradeType.BUY, s_decimal_zero, s_decimal_zero
+        possible = True
+        possible = possible and self.check_spot_budget_constraint(proposal)
+        possible = possible and self.check_perpetual_budget_constraint(proposal)
+        return possible
```
aarmoa, line 299:
```suggestion
        return (self.check_spot_budget_constraint(proposal) and self.check_perpetual_budget_constraint(proposal))
```

aarmoa (commented):
Added a couple minor comments that can be disregarded by @petioptrv if it does not make sense to make a new commit for them.

aarmoa (approved):
Changes look good to me

RC-13:
Re-posting issue here: Parsing error are seen on the Beaxy and HitBTC connector when running on this feature branch
image
[logs_conf_pure_mm_5.log](https://github.com/CoinAlpha/hummingbot/files/7559660/logs_conf_pure_mm_5.log)

petioptrv:
> Re-posting issue here: Parsing error are seen on the Beaxy and HitBTC connector when running on this feature branch !

Sorry for the back-and-forth @RC-13, but you were originally right: this is not an issue introduced by the current PR; it's simply made evident by the changes in the current PR. I've created a bug issue for it [here](https://app.zenhub.com/workspaces/hummingbot-5f340755c5e17300167a8f30/issues/coinalpha/hummingbot/4750).

RC-13 (approved):
LGTM

Nullably (approved):
LGTM.
Very well written code with good explanation.
I can see BudgetChecker evolves into BudgetManager in the future for extra features like budget allocation and limit.

PR #4691, opened 2021-11-11
Add sample configs for the V2 config manager

james-hummingbot:


gateway/test/services/config-manager-v2.test.ts
```diff
@@ -146,3 +146,11 @@ describe('Configuration manager v2 tests', () => {
     done();
   });
 });
+
+describe('Sample configurations', () => {
+  it('Read sample schemas', (done) => {
+    const sampleConfigManager = new ConfigManagerV2('./conf/samples/root.yml');
```
james-hummingbot, line 152: This is the main test, that there is not a runtime error when loading all of the sample configs.

gateway/test/services/config-manager-v2.test.ts
```diff
@@ -146,3 +146,11 @@ describe('Configuration manager v2 tests', () => {
...
+    expect(sampleConfigManager.get('ssl.caCertificatePath')).toBeDefined();
```
james-hummingbot, line 153: This makes the compiler happy that we are not creating sampleConfigManager for nothing.

gateway/conf/samples/ssl.yml
```diff
@@ -0,0 +1,13 @@
+# need more details...
```
james-hummingbot, line 1: I know we need these files for https, but I don't know the details.

martinkou (approved):

martinkou:
I think I'll prefer to put the schema files under `src/services/schema/` - since they're not expected to be changed by the user. But currently, there's no effective way for the `root.yml` file to refer to the files there. So I'm gonna create a task for that later.

LGTM for the code here.

PR #4720, opened 2021-11-15
Crash if the conf/gateway-passphrase.yml does not exist or does not have expected permissions

james-hummingbot:


gateway/src/services/config-manager-cert-passphrase.ts
```diff
@@ -0,0 +1,46 @@
+import fs from 'fs';
+import yaml from 'js-yaml';
+import { logger } from './logger';
+
+export namespace ConfigManagerCertPassphrase {
+  export const passphraseFliePath: string = './conf/gateway-passphrase.yml';
```
james-hummingbot, line 6: Had to move this code out of config-manager.ts for two reasons: 
1. logger depends on config-manager, it would create a circular dependency.
2. it breaks some unit testing to have this not working because of the process.exit, better to have config-manager without it so it is easier to test.

gateway/jest.config.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
     'src/app.ts',
     'src/https.ts',
     'src/services/ethereum-base.ts',
+    'src/services/config-manager-cert-passphrase.ts',
```
james-hummingbot, line 9: We have some unit tests, but not complete coverage because of process.exit

gateway/src/services/config-manager-cert-passphrase.ts
```diff
@@ -0,0 +1,46 @@
...
+        const x = yaml.load(
+          fs.readFileSync(passphraseFliePath, 'utf8')
+        ) as PassphraseConfig;
+        if ('CERT_PASSPHRASE' in x) {
+          return x.CERT_PASSPHRASE;
+        } else {
+          logger.error(
+            passphraseFliePath + ' does not have CERT_PASSPHRASE set.'
+          );
+          process.exit(1);
```
martinkou, line 29: Since this is the very point of this module, we should add a way to make this testable. We can't test the process exit itself of course, but we can test whether it's place in the code has been triggered.

We can do this by adding a layer of indirection. e.g.

```
export const bindings = {
   _exit: process.exit,
};

        ...
        logger.error(
          passphraseFliePath +
            ' file does not strictly have mode set to user READ_WRITE only.'
        );
        bindings._exit(1);
        ...
```

So when it comes to unit testing, we can then use the `patch()` function to test the exit functionality.

```
  ...
  patch(ConfigManagerCertPassphrase.bindings, "_exit", () => { /* record something here */ });
  ...
```

martinkou (changes requested):

gateway/src/services/config-manager-cert-passphrase.ts
james-hummingbot: @martinkou Good idea, I've implemented this change.

martinkou (approved):

PR #4736, opened 2021-11-16
Refactor/calculation of the greek parameters (kappa, gamma and eta)

mhrvth:

Automatic estimation of order book liquidity parameters (alpha, kappa) as a new indicator "trading intensity"
Change in greeks calculation
Removal of artificial helper parameters altering the strategy behavior
Instant volatility calculation fix
Fixes of multiple equations and variables to achieve unit consistency and calculation correctness
Removal of min-max spread logic and logspaced levels
Implementation of new log levels based on optimal spread
Implementation of multi-tick deep order books for order book liquidity estimation in unit tests
Change of the "algorithm ready" condition to account for the order book liquidity estimation
Added scipy as a new package dependency

Tested order book liquidity estimation logic and strategy performance on various pairs with the binance paper connector
Ran the unit tests

### DIFF
--- a/hummingbot/strategy/__utils__/trailing_indicators/trading_intensity.pyx
+++ b/hummingbot/strategy/__utils__/trailing_indicators/trading_intensity.pyx
@@ -24,6 +24,23 @@ cdef class TradingIntensityIndicator():
         warnings.simplefilter("ignore", OptimizeWarning)
 
     def _simulate_execution(self, bids_df, asks_df):
+        self.c_simulate_execution(bids_df, asks_df)
+
+    cdef c_simulate_execution(self, new_bids_df, new_asks_df):
+        cdef:
+            object _bids_df = self._bids_df
+            object _asks_df = self._asks_df
+            object bids_df = new_bids_df
+            object asks_df = new_asks_df
+            int _sampling_length = self._sampling_length
+            object bid
+            object ask
+            object price
+            object bid_prev
+            object ask_prev
+            object price_prev
+            list trades
+
         # Estimate market orders that happened
         # Assume every movement in the BBO is caused by a market order and its size is the volume differential
 
@@ -31,15 +48,15 @@ cdef class TradingIntensityIndicator():
         ask = asks_df["price"].iloc[0]
         price = (bid + ask) / 2
 
-        bid_prev = self._bids_df["price"].iloc[0]
-        ask_prev = self._asks_df["price"].iloc[0]
+        bid_prev = _bids_df["price"].iloc[0]
+        ask_prev = _asks_df["price"].iloc[0]
         price_prev = (bid_prev + ask_prev) / 2
 
         trades = []
 
         # Higher bids were filled - someone matched them - a determined seller
         # Equal bids - if amount lower - partially filled
-        for index, row in self._bids_df[self._bids_df['price'] >= bid].iterrows():
+        for index, row in _bids_df[_bids_df['price'] >= bid].iterrows():
             if row['price'] == bid:
                 if bids_df["amount"].iloc[0] < row['amount']:
                     amount = row['amount'] - bids_df["amount"].iloc[0]
@@ -52,7 +69,7 @@ cdef class TradingIntensityIndicator():
 
         # Lower asks were filled - someone matched them - a determined buyer
         # Equal asks - if amount lower - partially filled
-        for index, row in self._asks_df[self._asks_df['price'] <= ask].iterrows():
+        for index, row in _asks_df[_asks_df['price'] <= ask].iterrows():
             if row['price'] == ask:
                 if asks_df["amount"].iloc[0] < row['amount']:
                     amount = row['amount'] - asks_df["amount"].iloc[0]
@@ -65,10 +82,18 @@ cdef class TradingIntensityIndicator():
 
         # Add trades
         self._trades += [trades]
-        if len(self._trades) > self._sampling_length:
+        if len(self._trades) > _sampling_length:
             self._trades = self._trades[1:]
 
     def _estimate_intensity(self):
+        self.c_estimate_intensity()
+
+    cdef c_estimate_intensity(self):
+        cdef:
+            dict trades_consolidated
+            list lambdas
+            list price_levels
+
         # Calculate lambdas / trading intensities
         lambdas = []
 
--- a/hummingbot/strategy/__utils__/trailing_indicators/trading_intensity.pyx
+++ b/hummingbot/strategy/__utils__/trailing_indicators/trading_intensity.pyx
@@ -1,0 +1,149 @@
+from math import floor, ceil
+from decimal import Decimal
+import numpy as np
+import pandas as pd
+from scipy.optimize import curve_fit
+from scipy.optimize import OptimizeWarning
+import csv
+import warnings
+
+from typing import (
+    Tuple,
+)
+
+
+cdef class TradingIntensityIndicator():
+
+    def __init__(self, sampling_length: int = 30, order_amount: float = 1, price_levels: Tuple[float] = (0)):
+        self._alpha = 0
+        self._kappa = 0
+        self._trades = []
+        self._bids_df = None
+        self._asks_df = None
+        self._order_amount = order_amount
+        self._sampling_length = sampling_length
+        self._samples_length = 0
+
+        warnings.simplefilter("ignore", OptimizeWarning)
+
+    def _simulate_execution(self, timestamp, bids_df, asks_df):
+        # Estimate market orders that happened
+        # Assume every movement in the BBO is caused by a market order and its size is the volume differential
+
+        bid = bids_df["price"].iloc[0]
+        ask = asks_df["price"].iloc[0]
+        price = (bid + ask) / 2
+
+        bid_prev = self._bids_df["price"].iloc[0]
+        ask_prev = self._asks_df["price"].iloc[0]
+        price_prev = (bid_prev + ask_prev) / 2
+
+        trades = []
+
+        # Higher bids were filled - someone matched them - a determined seller
+        # Equal bids - if amount lower - partially filled
+        for index, row in self._bids_df[self._bids_df['price'] >= bid].iterrows():
+            if row['price'] == bid:
+                if bids_df["amount"].iloc[0] < row['amount']:
+                    amount = row['amount'] - bids_df["amount"].iloc[0]
+                    price_level = abs(row['price'] - price_prev)
+                    trades += [{'price_level': price_level, 'amount': amount}]
+            else:
+                amount = row['amount']
+                price_level = abs(row['price'] - price_prev)
+                trades += [{'price_level': price_level, 'amount': amount}]
+
+        # Lower asks were filled - someone matched them - a determined buyer
+        # Equal asks - if amount lower - partially filled
+        for index, row in self._asks_df[self._asks_df['price'] <= ask].iterrows():
+            if row['price'] == ask:
+                if asks_df["amount"].iloc[0] < row['amount']:
+                    amount = row['amount'] - asks_df["amount"].iloc[0]
+                    price_level = abs(row['price'] - price_prev)
+                    trades += [{'price_level': price_level, 'amount': amount}]
+            else:
+                amount = row['amount']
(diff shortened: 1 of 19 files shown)


hummingbot/strategy/__utils__/trailing_indicators/trading_intensity.pyx
```diff
@@ -0,0 +1,146 @@
...
+import pandas as pd
+from scipy.optimize import curve_fit
+from scipy.optimize import OptimizeWarning
+from typing import (
+    Tuple,
+)
+import warnings
+
+
+cdef class TradingIntensityIndicator():
```
aarmoa, line 13: Why are we creating this class in Cython if all functions are pure Python?

hummingbot/strategy/avellaneda_market_making/avellaneda_market_making_config_map.py
```diff
@@ -224,14 +144,26 @@ def on_validated_parameters_based_on_spread(value: str):
...
-                  validator=lambda v: validate_decimal(v, 5, 600),
-                  default=60),
+                  validator=lambda v: validate_decimal(v, 1, 10000),
+                  default=200),
+    "trading_intensity_buffer_size":
+        ConfigVar(key="trading_intensity_buffer_size",
+                  prompt="Enter amount of ticks that will be stored to estimate order book liquidity >>> ",
+                  type_str="int",
+                  validator=lambda v: validate_decimal(v, 1, 10000),
+                  default=200),
```
aarmoa, line 154: The type is defined as `int` but the validator used is `validate_decimal`.

setup.py
```diff
@@ -1,11 +1,10 @@
 import numpy as np
 import os
+from setuptools import find_packages, setup
+from setuptools.command.build_ext import build_ext
```
aarmoa, line 4: Imports with `from` should come after direct imports.

test/hummingbot/strategy/avellaneda_market_making/test_avellaneda_market_making.py
```diff
@@ -174,14 +172,123 @@ def simulate_high_volatility(strategy: AvellanedaMarketMakingStrategy):
...
+        # Update the trading intensity indicator
+        strategy.trading_intensity = trading_intensity_indicator
+
+        # # Simulates change in mid price to reflect last sample added
+        # strategy.market_info.market.set_balanced_order_book(trading_pair=strategy.trading_pair,
+        #                                                     mid_price=samples[-1],
+        #                                                     min_price=1,
+        #                                                     max_price=200,
+        #                                                     price_step_size=1,
+        #                                                     volume_step_size=10)
```
aarmoa, line 208: Is this code no longer required?

test/hummingbot/strategy/avellaneda_market_making/test_avellaneda_market_making.py
```diff
@@ -174,14 +172,123 @@ def simulate_high_volatility(strategy: AvellanedaMarketMakingStrategy):
...
```
aarmoa, line 243: These lines are also commented

test/hummingbot/strategy/avellaneda_market_making/test_avellaneda_market_making.py
```diff
@@ -514,90 +603,44 @@ def test_calculate_target_inventory(self):
...
-
-        # Initialize strategy with high vol_to_spread_multiplier config
-        self.strategy.vol_to_spread_multiplier = Decimal("10")
-
-        curr_price: Decimal = self.strategy.get_price()
-        curr_vol: Decimal = self.strategy.get_volatility()
-        expected_min_spread: Decimal = self.strategy.vol_to_spread_multiplier * curr_vol
-        expected_max_spread: Decimal = self.max_spread * curr_price * (expected_min_spread / (self.min_spread * curr_price))
+        self.assertAlmostEqual(100.58688318883304, alpha, 5)
+        self.assertAlmostEqual(0.03153874736059721, kappa, 5)
```
aarmoa, line 614: It would be useful for developers in general to have a calculation here indicating how we get the expected alpha and kappa, instead of just hardcoding the expected numbers in the test.

aarmoa (commented):
I have added a couple of comments to be validated by @mhrvth.
In general I think it would be better in the unit tests, when we assert a result against an expected value, to have a calculation in the test providing the expected value, instead of a hardcoded number. In that way the test also helps any developer understand the calculation being done by our components (in case it is possible to include the calculation in the tests)

mhrvth:
scipy had to be downgraded because the newest package seems to be broken on windows

RC-13:
- Level_distance doesn't seems to work. In this screenshot, the level_distance is set to 1% and I'm expecting to see 1% different between each orders however the spread of each orders are the same.
level-distance-1
- The buffer size as well seems to be defaulted to 195 ticks, in the clip, we set the volatility and trading intensity buffer size to 5 to allow the strategy to start faster however, the bot still started with the default value. Increasing the buffer size  higher than 200 doesn't have any issue.

https://user-images.githubusercontent.com/57189990/142137015-8400d09a-bfbe-4238-b1ef-4970b4471b5a.mov

- Also need clarification on how does `min_spread` work, during testing, when I tried increasing the min_spread from 0 to 1, the order spread created reached exceeded 200%

https://user-images.githubusercontent.com/57189990/142138440-4b3c65b6-1171-43b7-a5b1-052a129e7cf7.mov

mhrvth:
level_distance are % of the current spread, therefore 1% for DOGE-USDT is a small distance in absolute values, not visible in the printout.

mhrvth:
The values cannot be verified through a calculation because they're based on an estimation of a stochastic process. They can be estimated as valid / invalid considering the liquidity of the pair. For low liquidity pairs the orders should be placed further away from the mid price than for high liquidity pairs, given that the volatility is the same, the portfolio is balanced, and the time to end is the same.

aarmoa (changes requested):

aarmoa:
Even with the latest changes in `bot.spec` made by @mhrvth the binary version in Windows is failing. I have found a way to configure the file that solves the issue. I paste following the contents from the hidden imports definition until the creation of the Analysis instance. In particular I added two elements in the array used to extend the hidden imports, and I also an element in the array used to extend binaries list when the distribution is Windows.

```
hidden_imports: List[str] = list(enumerate_modules(os.path.join(project_path(), "hummingbot")))
    hidden_imports.extend([
        "aiokafka",
        "pkg_resources.py2_warn",
        "scipy",
        "scipy.spatial.transform._rotation_groups",
    ])
    hidden_imports.extend(collect_submodules('scipy'))

    import _strptime

    datas: List[Tuple[str, str]] = list(enumerate_data_files(
        os.path.join(project_path(), "hummingbot"),
        r"(.+\.json|(?:\/|\\)VERSION|templates(?:\/|\\).+\.yml)$"
    ))
    datas.extend([(_strptime.__file__, ".")])
    datas.extend([(os.path.join(project_path(), "bin/path_util.py"), ".")])
    datas.extend(collect_data_files('scipy'))

    binaries: List[Tuple[str, str]] = []
    if system_type == "Windows":
        import coincurve
        conda_env_path = os.environ['CONDA_PREFIX']
        binaries.extend([
            (os.path.realpath(os.path.join(coincurve.__file__, "../libsecp256k1.dll")), "coincurve"),
            (os.path.realpath(os.path.join(conda_env_path, "Library", "bin", "libiomp5md.dll")), ".")])
        datas.extend([(os.path.realpath(os.path.join(project_path(), "redist/VC_redist.x64.exe")), "redist")])
```

aarmoa:
We also have to change back the scipy dependency to be a conda dependency

mhrvth:
For me the binary version did work correctly. Not sure why you'd want to specify the scipy hidden imports when all are being collected just below. The same for the OpenMP dll library.

RC-13:
- The `inventory_target_base_pct`parameter appears to be ignored, in this example screenshot, the inventory_target_base_pct is set to 5% however the bot still created both side of the order:
ave-inventory-pct-pr
Checked the development branch with the same configuration and the parameter behave as intended.
ave-inventory-pct-dev

- We are getting random `Unexpected error running clock tick` usually when changing the inventory target base pct, although the bot recover however there are instances that it may occur during strategy startup and prevent the creation of order.
image
```python
2021-11-22 12:43:32,019 - 14097 - hummingbot.core.clock - ERROR - Unexpected error running clock tick.
  File "hummingbot/strategy/avellaneda_market_making/avellaneda_market_making.pyx", line 602, in hummingbot.strategy.avellaneda_market_making.avellaneda_market_making.AvellanedaMarketMakingStrategy.c_tick
    self.c_calculate_reserved_price_and_optimal_spread()
  File "hummingbot/strategy/avellaneda_market_making/avellaneda_market_making.pyx", line 710, in hummingbot.strategy.avellaneda_market_making.avellaneda_market_making.AvellanedaMarketMakingStrategy.c_calculate_reserved_price_and_optimal_spread
    self._optimal_spread += 2 * Decimal(1 + self._gamma / self._kappa).ln() / self._gamma
decimal.InvalidOperation: [<class 'decimal.InvalidOperation'>]
```

rapcmia:
Adding to RC's comment on the `unexpected error running clock tick`, mine does not recover from the error.  In this case, I set `inventory_target_base_pct` to 10 with an amount enough for base assets to start creating order but no go. 

image

Attach logs and config file:
[ave-test.zip](https://github.com/CoinAlpha/hummingbot/files/7578613/ave-test.zip)

mhrvth:
Creating orders on both sides is normal. The balancing is done by skewing their prices relative to the mid price. If the strategy needs to buy more of an asset, buy orders are placed closer to the mid price and sell orders further away, so that buy orders are more likely to be filled and sell orders less likely.

mhrvth:
Creating orders on both sides is normal. The balancing is done by skewing their prices relative to the mid price. If the strategy needs to buy more of an asset, buy orders are placed closer to the mid price and sell orders further away, so that buy orders are more likely to be filled and sell orders less likely.

mhrvth:
Creating orders on both sides is normal. The balancing is done by skewing their prices relative to the mid price. If the strategy needs to buy more of an asset, buy orders are placed closer to the mid price and sell orders further away, so that buy orders are more likely to be filled and sell orders less likely.

mhrvth:
Regarding the exception: are you running it with the latest changes? I've fixed this problem, but it's possible it's not in github yet.

mhrvth:
Can you also please check if you're able to build a binary version on Windows and run the strategy?

RC-13:
Hello @mhrvth the last commit `60fa015` fix the windows binary build issue introduce on commit `3b86212` however it broke the compiling process on mac:
image

mhrvth:
So, in fact, @aarmoa was right, it did not build well from an absolute scratch. The problem was that installation of scipy from conda-forge fails to build a package, and fails silently, therefore the fortran dll was missing each time installation was done with conda, and that's why reinstallation with pip fixed it. From the defaults channel the scipy package just installs and is not build, that works well. In the environment scripts conda-forge has a priority over defaults. If a package is found in conda-forge, defaults is not searched. The solution was to specify the channel from where the scipy should be installed. Another issue was the collecting the dll libraries in the spec file didn't really work. Collection of modules did work. Because of that the OpenMP dll was missing if done from a complete scratch. The solution, as in Abel's script was to specify it exactly. On top of that it's unnecessary to collect and import all scipy submodules, therefore an exact module that is needed is specified as well.

pyinstaller/bot.spec
```diff
@@ -104,7 +101,7 @@ if "SPEC" in globals():
     exe = EXE(pyz,
               a.scripts,
               [],
-	      exclude_binaries=True,
+          exclude_binaries=True,
```
aarmoa, line 104: There seems to be an indentation difference here

aarmoa (commented):
I have added a note regarding an indentation difference. Besides that and the integration issue that needs to be solved (the file setup/requirements-arm.txt has been removed by other PR) I agree with the rest of the changes. I will approve the code review when the last fixes are included.

aarmoa (approved):
All changes look good to me

RC-13 (approved):
LGTM

PR #4747, opened 2021-11-17
Feat/gateway docker file

Nullably:

- create a docker file for Gateway v2
- add hook for auto build to docker hub 

- docker image can be created from docker file
- docker image is created on docker hub

**Tips for QA testing**:
- docker image and container created from this docker file cannot be used yet as it's still required further process to update environment variables and create config file

### DIFF
--- a/gateway/Dockerfile
+++ b/gateway/Dockerfile
@@ -14,14 +14,11 @@ ENV BUILD_DATE=${DATE}
 # app directory
 WORKDIR /usr/src/app
 
-# install dependancies
-COPY package*.json ./
-# COPY yarn.lock ./
-
-RUN yarn
-
 # copy pwd file to container
 COPY . .
+
+# install dependencies
+RUN yarn
 
 EXPOSE 5000
 
--- a/gateway/Dockerfile
+++ b/gateway/Dockerfile
@@ -18,6 +18,8 @@ WORKDIR /usr/src/app
 COPY package*.json ./
 COPY yarn.lock ./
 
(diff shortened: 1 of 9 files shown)


gateway/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+WORKDIR /usr/src/app
+
+# install dependancies
+COPY package*.json ./
+# COPY yarn.lock ./
+
+RUN yarn
+
+# copy pwd file to container
+COPY . .
```
vic-en, line 24:
```suggestion
# copy pwd file to container
COPY . .

# install dependencies
RUN yarn
```

vic-en: What do you this about this approach of copying files first then going on with the rest?

vic-en (changes requested):
Also I believe `.gitignore` file should be updated to ignore some sensitive gateway files.

gateway/Dockerfile
Nullably: I am fine with that.

Nullably:
> Also I believe `.gitignore` file should be updated to ignore some sensitive gateway files.

Which files are you referring to?

vic-en:
> > Also I believe `.gitignore` file should be updated to ignore some sensitive gateway files.
> 
> Which files are you referring to?

Nvm, I wanted to mention `.gitignore`. But I see it doesn't need at update after checking again.

vic-en (approved):
LGTM

PR #4786, opened 2021-11-22
Feat/make configuration manager v2 upgradable

vic-en:


gateway/src/services/config-manager-v2.ts
```diff
@@ -267,10 +304,35 @@ export class ConfigManagerV2 {
...
     ) as ConfigurationRoot;
+    const configRootTemplate: ConfigurationRoot = yaml.load(
+      fs.readFileSync(configRootTemplateFullPath, 'utf8')
+    ) as ConfigurationRoot;
+
+    // version control to only handle upgrades
+    if (configRootTemplate.version > configRoot.version) {
+      // latest root config missing, copying from template
+      fs.copyFileSync(configRootTemplateFullPath, configRootFullPath);
+      Object.assign(configRoot, configRootTemplate);
```
vic-en, line 323:
```suggestion
```

gateway/src/services/config-manager-v2.ts
```diff
@@ -76,9 +89,14 @@ export class ConfigurationNamespace {
       JSON.parse(fs.readFileSync(schemaPath).toString())
     );
 
-    if (fs.existsSync(configurationPath)) {
-      this.loadConfig();
+    if (!fs.existsSync(configurationPath)) {
+      // copy from template
+      fs.copyFileSync(
+        path.join(ConfigTemplatesDir, path.basename(configurationPath)),
```
martinkou, line 95: As discussed during the 1:1 video chat, you are adding an undocumented design constraint here by assuming every configuration file has to be located immediately under the configuration folder, with no subdirectories in between. The configuration v2 specs never mentioned this constraint - you're in effect changing the design in the code here.

You need to ask an architectural question here - whether the configuration namespace, or the configuration manager, is the best place to manage the mapping from the configuration paths to the template paths. In this particular case, I'd say it's better for the configuration manager to manage that relation. So, if the configuration namespace doesn't manage the mapping, then it most likely should just receive the template file path from the manager via its constructor.

gateway/src/services/config-manager-v2.ts
```diff
@@ -98,7 +116,26 @@ export class ConfigurationNamespace {
...
+      // merge with template file and try validating again
+      const configTemplateCandidate: Configuration = yaml.load(
+        fs.readFileSync(
+          path.join(ConfigTemplatesDir, path.basename(this.#configurationPath)),
+          'utf8'
+        )
+      ) as Configuration;
+      if (
+        !this.#validator(
+          Object.assign(configTemplateCandidate, configCandidate)
```
martinkou, line 128: As discussed in the 1:1 video conf, you need a deep copy here. A shallow copy like `Object.assign()` would overwrite the whole config tree as a single property - which will break new properties added in the template that are not defined at the top level.

martinkou (changes requested):

gateway/src/services/config-manager-v2.ts
```diff
@@ -267,10 +323,31 @@ export class ConfigManagerV2 {
...
       fs.readFileSync(configRootFullPath, 'utf8')
     ) as ConfigurationRoot;
+    const configRootTemplate: ConfigurationRoot = yaml.load(
+      fs.readFileSync(configRootTemplateFullPath, 'utf8')
+    ) as ConfigurationRoot;
+
+    // version control to only handle upgrades
+    if (configRootTemplate.version > configRoot.version) {
+      // run migration in order if available
+      for (
```
martinkou, line 341: Have you thought about the case where the current config is at version 10, the newest version is version 13 - and you've got migration functions defined for versions 1, 3, 6, 10, 12, 13? Simply running everything from version 1 doesn't work.

gateway/src/services/config-manager-v2.ts
```diff
@@ -3,29 +3,53 @@ import { ValidateFunction } from 'ajv';
...
+};
 interface _ConfigurationNamespaceDefinition {
   configurationPath: string;
   schemaPath: string;
 }
 type ConfigurationNamespaceDefinition = _ConfigurationNamespaceDefinition & {
   [key: string]: string;
 };
-type ConfigurationRoot = {
+type ConfigurationPattern = {
```
martinkou, line 21: Bad name. Saying something is a "Pattern" usually means it's something like a regex expression. Rename this to "ConfigurationNamespaceDefinitions".

gateway/src/services/config-manager-v2.ts
```diff
@@ -281,17 +358,31 @@ export class ConfigManagerV2 {
...
+        if (!path.isAbsolute(filePath)) {
           namespaceDefinition[key] = path.join(configRootDir, filePath);
+          if (key === 'configurationPath') {
+            namespaceDefinition['templatePath'] = path.join(
+              ConfigTemplatesDir,
+              filePath
+            );
+          }
+        } else {
+          //assumption is made that config file is relative to root config file
```
martinkou, line 379: There should be a space after //.

gateway/src/services/config-manager-v2.ts
```diff
@@ -281,17 +358,31 @@ export class ConfigManagerV2 {
...
+          if (key === 'configurationPath') {
+            namespaceDefinition['templatePath'] = path.join(
+              ConfigTemplatesDir,
+              path.relative(configRootDir, filePath)
```
martinkou, line 383: If the file path is absolute, and yet we want to map it to a template - then this mapping doesn't make sense anymore. We should simply raise an error instead and disallow any absolute paths in the configuration definitions.

gateway/src/services/config-manager-v2.ts
martinkou: Also, if you're executed any migrations, you should save the configurations after `addNamespace()`.

gateway/src/services/config-manager-v2.ts
```diff
@@ -93,12 +126,33 @@ export class ConfigurationNamespace {
...
+      // merge with template file and try validating again
+      const configTemplateCandidate: Configuration = yaml.load(
+        fs.readFileSync(this.#templatePath, 'utf8')
+      ) as Configuration;
+      deepCopy(configCandidate, configTemplateCandidate);
+      if (!this.#validator(configTemplateCandidate)) {
+        throw new Error(
+          `Configuration for namespace ${this.id} seems to be outdated/broken. Kindly fix manually.`
+        );
+      } else {
```
martinkou, line 151: Since you're throwing an error and exiting on the if block above - you do not need to else block here. You should just put the rest of it outside the if block. Flat is better than nested.

martinkou (changes requested):

gateway/conf/migration/migrations.js
```diff
@@ -0,0 +1,10 @@
+/*
+ * This file defines migration functions for each version.
+ * Note: type Migration = () => void;
+ */
+
+module.exports = {
+  updateToVersion1() {
```
martinkou, line 7: If somebody is writing a migration function here, does he have all the needed information?

This function's type def has no arguments. So, where do I find the configuration to upgrade? Where do I find the template? etc.

While I didn't write the specs down to this detail - the expectation is you should be able to think about what's needed here and explain why that's useful / easy to use / simple for the engineers who're going to use your design.

gateway/src/services/config-manager-v2.ts
```diff
@@ -3,29 +3,53 @@ import { ValidateFunction } from 'ajv';
...
   schemaPath: string;
 }
 type ConfigurationNamespaceDefinition = _ConfigurationNamespaceDefinition & {
   [key: string]: string;
 };
-type ConfigurationRoot = {
+type ConfigurationPattern = {
   [namespaceId: string]: ConfigurationNamespaceDefinition;
 };
+type ConfigurationRoot = {
```
martinkou, line 24: Follow the convention we have... if this has fixed members, then it's an interface.

gateway/src/services/config-manager-v2.ts
```diff
@@ -3,24 +3,37 @@ import { ValidateFunction } from 'ajv';
...
+  configurations: configurationPattern;
+};
 const NamespaceTag: string = '$namespace ';
 const ConfigRootSchemaPath: string = path.join(
   __dirname,
   'schema/configuration-root-schema.json'
 );
+const ConfigTemplatesDir: string = path.join(
+  __dirname,
+  '../../conf/templates/'
```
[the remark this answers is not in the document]
martinkou: You have not fixed my comment above. You need to put your templates in the source directories, not under `conf/`.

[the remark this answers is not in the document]
vic-en: ok, I missed the part of copying them into the `/src` directory. I'll address now.

gateway/src/services/config-manager-v2.ts
vic-en: We may have to think about this again. saving after call to `addNamespace()` wouldn't work because configurations are loaded during call to `addNamespace()`.

vic-en: The way I imagined it is that saving/updating will be done during migration. I was thinking that was the major reason you wanted functions like `initiateWithTemplate` exported for use in the migration file.

gateway/src/services/config-manager-v2.ts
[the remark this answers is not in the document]
vic-en: Done!

gateway/conf/migration/migrations.ts
```diff
@@ -0,0 +1,17 @@
+import * as utils from '../../src/services/config-manager-v2';
+
+/*
+ * This file defines migration functions for each version.
+ * Note: type utils.Migration = (configRootFullPath: string,
+ *                         configRootTemplateFullPath: string) => void;
+ */
+
+export const updateToVersion1: utils.Migration = (
```
martinkou, line 9: Since this file is not supposed to be changed by the user, but by gateway developers - it should not be placed inside the conf folder. Lets move it to `gateway/src/services/config-migration/migrations.ts`.

gateway/src/services/config-manager-v2.ts
```diff
@@ -3,29 +3,60 @@ import { ValidateFunction } from 'ajv';
 import fs from 'fs';
 import path from 'path';
 import yaml from 'js-yaml';
+import * as migrations from '../../conf/migration/migrations';
```
martinkou, line 6: Change this line after moving the config migration file. Besides, as a reminder, importing the conf directory directly doesn't work.

gateway/src/services/config-manager-v2.ts
```diff
@@ -3,29 +3,60 @@ import { ValidateFunction } from 'ajv';
...
   configPath: string;
 }
 
+export function deepCopy(srcObject: any, dstObject: any): any {
+  for (const key in srcObject) {
+    if (typeof srcObject[key] !== 'object') {
+      if (typeof dstObject[key] !== 'object') dstObject[key] = srcObject[key];
+    } else {
+      if (!dstObject[key]) {
+        if (srcObject[key] instanceof Array) dstObject[key] = [];
```
martinkou, line 48: WTF? Have you actually tested this function yourself?? This doesn't work for arrays since `for (const key in srcObject)` with an array would iterate the values instead of keys.

martinkou (changes requested):

gateway/src/services/config-manager-v2.ts
martinkou: Change the structure of this function to this, to make it easier to read.

```
if (srcObject instanceof Array) {
  ...
} else if (srcObject instanceof Object) {
  ...
} else {
  ...
}
```

.gitignore
```diff
@@ -81,9 +81,9 @@ coverage.xml
 # Gateway-V2 repo
 # configuration files
 /gateway/conf/*
-# except the example files in /gateway/conf/ and /gateway/conf/samples
+# except the examples, template and migration files in /gateway/conf/
 !/gateway/conf/*-example.yml
-!/gateway/conf/samples
+!/gateway/conf/migration
```
martinkou, line 86: This can be removed after moving the migrations directory inside `src/`.

gateway/src/services/config-manager-v2.ts
martinkou: And make sure to include a unit test with some nested objects and arrays. e.g.

```
{
  key1: "string value",
  key2: {
    ...
  },
  key3: [
    { ... },
    { ... },
    ...
  ],
}
```

gateway/src/services/config-manager-v2.ts
vic-en: I thought `"copy-files": "copyfiles -a 'src/**/schema/*.json' 'test/services/data/**/*.*' 'src/templates/*.*' dist",` in the package.json handles that. Anyway, I'll update it not to copy over files from the conf directory since we've moved the template and schema to the project src

gateway/src/services/config-manager-v2.ts
vic-en: A unit test is already included in this pr to test similar case - https://github.com/CoinAlpha/hummingbot/blob/6efa89f5662419b87c594078c88a48d16b1a49fa/gateway/test/services/config-manager-v2.test.ts#L163
Can you take a look?

vic-en: Also, Typescript seems to have issue with this `for (const [key, value] in Object.entries(srcObject)) {`. I'll try to figure a solution to that.

martinkou: You have a unit test for having objects and arrays, but not nested objects and arrays. What I mean is you don't have cases like "key3" above - where you have objects inside arrays. The converse (arrays inside something like key2) should also be tested.

vic-en: Done!

vic-en: One thing to note is that the order of entries in arrays are sensitive during deepcopy. Inserting new element at the beginning or in between an array list will cause entries after that insertion to be overwritten.

vic-en: I also think there's really no need to cater for such cases as that will make the  logic very complex.

martinkou (approved):

martinkou:
LGTM, but please fix the unit test failures after the merge.

PR #4790, opened 2021-11-23
(feature) Adds slippage buffer setting for XEMM strategy.

petioptrv:

This PR adds the changes proposed by @jellebuth [here](https://github.com/CoinAlpha/hummingbot/pull/4557).

The strategy compiles and runs with the new changes.

**Tips for QA testing**:
- To test the main feature of the changes, set the slippage buffer to something high, like 10% and make sure that the taker orders are placed at 10% away from the ask or bid respectively (ask_price * 1.1 for buy orders and bid_price * 0.9 for sell orders).
- To test the safety feature for maker sell orders, set the slippage to something even higher, like 50%, and check that the strategy rejects maker orders for which it has enough absolute balance on the taker exchange, but not enough if we factor in the 50% slippage. E.g. if we want to place a maker sell order to sell 10 USDT worth of COINX, we will need around 15 USDT on the taker market account so that the strategy goes through with it. 
    - More precisely, we need the VWAP * order_size * 1.5 USDT. The VWAP is calculated like this: if we place a sell order of, say 2 COINX on the maker (because COINX 1 = 5 USDT), and the order book on the taker market has one ask for 1 COINX at 5 USDT and one ask for 2 COINX at 6 USDT, then the VWAP for us to place the hedge is (1 * 5 + 1 * 6) / 2 = 5.5 USDT. We do 1 * 6 because we only need to buy 2 COINX, not 3. So we need VWAP * order_size * slippage_buffer = 5.5 * 2 * 1.5 = 16.5 USDT on the taker market.
    - This would be very hard to verify though, so as long as we have anything above 10 USDT and below 15 USDT on the taker market and the strategy refuses to place the maker order, this should probably be sufficient.


CrimsonJacket (dismissed):
Changes LGTM. 

Note: Additional logic as proposed by @jellebuth will be voted on before being implemented.

mhrvth (dismissed):
LGTM

RC-13 (approved):
LGTM

CrimsonJacket (dismissed):
Additional tests LGTM.

CrimsonJacket (approved):
Changes LGTM.

RC-13 (approved):
Re-approving PR

PR #4792, opened 2021-11-23
Fix / Wazirx Connector Trade Events

### LINKED ISSUE #4787: WazirX failing to update order status without stop/start of bot
**Describe the bug**
// A clear and concise description of what the bug is.
Like posted in https://github.com/CoinAlpha/hummingbot/pull/4563:

I just compiled the approved version (commit 035b44a merged into CoinAlpha:development), still the same issue:

Trade happens
binance leg confirms immediately
wazirx leg doesn't confirm at all sits as "open order" forever, only fix is stop / start and then it recognises the filled order

tested in cross-exchange market-making and arbitrage strategy - same result

**Steps To Reproduce**
// A concise description of the **steps to reproduce the buggy behavior**:

1. Set up your configuration file with these parameters: ...
2. Start the bot using this command: ...
...

just simply use it for any trade in cross-exchange market making or arbitrage wazirx/binance 
easiest to test: set profit to 0.01 to immediately trigger most of the time

orders will open correctly (arbitrage both, cross-money-making both wazirx legs)

arbitrage:
the binance leg will fill and bot will report it as filled
wazirx leg will fill and bot will not report it at all but have it sit as pending forever (I waited 15min after order was filled to test, nothing)

only when using stop and then start it recognises the fulfilled order and starts with empty order book 

cross-exchange market-making:
wazirx will post orders correctly
once wazirx order is filled on exchange, nothing happens in bot, it won't recognise it as filled
--- result: only 1 side of the leg will be traded, the binance leg will never be opened
also doesn't recognised partially filled orders obviously, so those are also just filled on wazirx without the complementing order on binance

**Screenshots**
// If applicable, add screenshots to help explain your problem.

**Release version**
// Include your bot's version number (Can be found at the upper left corner of your CLI).
dev-0.46.0

**Attachments**
// Please attach your config file and log file to help us debug. It would be difficult for us to help you without those! 

lijoppans:

Parsing order id and trade id in events to string to fix #4787.


aarmoa (approved):
Changes look good to me

aarmoa:
Even when the code change looks good to be applied, I am not sure the code that was changed is really related to the issue being addressed. The issue specifies that there are orders being filled but the client does not detect them. That is information the client should be receiving from the "trades update" user stream websocket channel. The code that processed those updates is in `_user_stream_event_listener' and more specifically for trade updates in `_process_trade_message`

lijoppans:
@aarmoa This was causing HB not to detect "partially filled cancelled orders". @davyboy confirmed that this fix solves his problem.

aarmoa:
@lijoppans Yes, I understand the problem. And again, both partial fills and full fills arrive to the client through the trade events in the user data source. It has nothing to do with the orders status update.

RC-13 (approved):
Tested and confirmed working by QA

davyboy:
I can confirm this solved this issue, thanks.

PR #4796, opened 2021-11-23
Make EVMNonceManager a normal class

james-hummingbot:


gateway/src/services/evm.nonce.ts
```diff
@@ -18,14 +17,6 @@ export class EVMNonceManager {
...
-    if (!EVMNonceManager._instance) {
-      EVMNonceManager._instance = new EVMNonceManager();
-    }
-
-    return EVMNonceManager._instance;
-  }
-
   // init can be called many times and generally should always be called
   // getInstance, but it only applies the values the first time it is called
   public async init(
```
martinkou, line 22: Since we're now giving different networks (e.g. Avalanche) their own `EVMNonceManager` - we should add the network name to the nonce manager init function.

martinkou (changes requested):

gateway/test/services/ethereum.base.test.ts
```diff
@@ -25,15 +25,16 @@ describe('Eth block listener test', () => {
...
+  //     expect(msg.action).toEqual('request');
+  //     done();
+  //   }
 
-    eth.onDebugMessage(processDebugMsg);
-    // this is the second request
-    eth.provider.emit('debug', { action: 'request' });
-    expect(eth.requestCount).toEqual(2);
-  });
+  //   eth.onDebugMessage(processDebugMsg);
```
james-hummingbot, line 35: This randomly fails. My guess there is some sort of time element. Best to fix it in another PR.

gateway/src/services/evm.nonce.ts
james-hummingbot: Good idea, I've made the change

martinkou (approved):

PR #4799, opened 2021-11-24
Feat/refactor location of config schemas

vic-en:


gateway/conf/samples/root.yml
```diff
@@ -1,35 +1,35 @@
 $namespace avalanche:
   configurationPath: avalanche.yml
-  schemaPath: ethereum-schema.json
+  schemaFileName: ethereum-schema.json
```
martinkou, line 3: Keep the "schemaPath" name. By renaming it to "xxxFileName" you're giving an illusion that this cannot be a path.

gateway/src/services/config-manager-v2.ts
```diff
@@ -290,8 +290,17 @@ export class ConfigManagerV2 {
     // Rebase the file paths in config root if they're relative paths.
     for (const namespaceDefinition of Object.values(namespaceMap)) {
       for (const [key, filePath] of Object.entries(namespaceDefinition)) {
-        if (filePath.charAt(0) !== '/') {
-          namespaceDefinition[key] = path.join(configRootDir, filePath);
+        if (key === 'configurationPath') {
+          if (!path.isAbsolute(filePath)) {
+            namespaceDefinition[key] = path.join(configRootDir, filePath);
+          }
+        } else {
```
martinkou, line 297: Use an `else if` here. Just an `else` works but is difficult to read, and more error prone when there are more schema changes.

gateway/src/services/config-manager-v2.ts
```diff
@@ -290,8 +290,17 @@ export class ConfigManagerV2 {
...
+          // create schemaPath and del schemaFileName
+          namespaceDefinition['schemaPath'] = path.join(
+            path.dirname(ConfigRootSchemaPath),
+            filePath
+          );
+          delete namespaceDefinition[key];
```
martinkou, line 303: Don't create a new name to begin with.

martinkou (changes requested):

gateway/conf/samples/root.yml
```diff
@@ -1,35 +1,35 @@
 $namespace avalanche:
   configurationPath: avalanche.yml
-  schemaPath: ethereum-schema.json
+  schemaPath: ../../src/services/schema/ethereum-schema.json
```
martinkou, line 3: The relative path should start from the schema directory, so there should be no need to change any of these.

When you're given the flexibility to design something for users, you should strive to make things simple. If you came up with a design that requires adding the same path prefix to entry - it's a good sign that the path prefix should not be there.

Also, make sure to test your updated config samples to make sure it works.

martinkou (changes requested):

martinkou (approved):

PR #4803, opened 2021-11-25
fix/ascendex auth timestamp

### LINKED ISSUE #4800: Ascendex connector should use current time as parameter for authentication
## Why
Ascendex connector is currently using a nonce as the timestamp parameter when generating the auth information. It is also using the nonce as the timestamp parameter for order creation and cancellation.

## What
The connector needs to use the real timestamp, even if the value is repeated when two request are generated at the same time.
This change will prepare the connector to use a synchronized timestamp once Ascendex implements a new endpoint to provide the server timestamp.

## DoD
The connector uses the real timestamp (expressed as milliseconds) to generate the auth header and for the timestamp parameter in both the create order and the cancel order requests.

aarmoa:

Change in the logic that generates the auth headers for Ascendex. The code now uses the real timestamp instead of a nonce. The same change is applied in the timestamp parameter used when creating and cancelling orders.

Added a unit test

**Tips for QA testing**:
Check private websocket events are received correctly (order status updates, balance updates and trade events). Also test that orders are created and cancelled without issues.

Closes #4800 4800

### DIFF
--- a/test/hummingbot/connector/exchange/ascend_ex/test_ascend_ex_utils.py
+++ b/test/hummingbot/connector/exchange/ascend_ex/test_ascend_ex_utils.py
@@ -50,16 +50,16 @@ class AscendExUtilsTests(TestCase):
         url = utils.get_ws_url_private(account_id=account_id)
         self.assertEqual(f"wss://ascendex.com/{account_id}/api/pro/v1", url)
 
-    def test_get_ms_timestamp(self):
-        with mock.patch('hummingbot.connector.exchange.ascend_ex.ascend_ex_utils._time') as get_tracking_nonce_low_res_mock:
-            get_tracking_nonce_low_res_mock.return_value = 1234567891.23456
+    @patch("hummingbot.connector.exchange.ascend_ex.ascend_ex_utils._time")
+    def test_get_ms_timestamp(self, time_mock):
+        time_mock.return_value = 1234567891.23456
 
-            timestamp = utils.get_ms_timestamp()
-            self.assertEqual(1234567891234, timestamp)
-            # If requested two times at the same moment it should return the same value (writen this way to ensure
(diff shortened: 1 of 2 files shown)


test/hummingbot/connector/exchange/ascend_ex/test_ascend_ex_utils.py
```diff
@@ -51,11 +51,15 @@ def test_get_ws_url_private(self):
         self.assertEqual(f"wss://ascendex.com/{account_id}/api/pro/v1", url)
 
     def test_get_ms_timestamp(self):
-        with mock.patch('hummingbot.connector.exchange.ascend_ex.ascend_ex_utils.get_tracking_nonce_low_res') as get_tracking_nonce_low_res_mock:
-            get_tracking_nonce_low_res_mock.return_value = 123456789
+        with mock.patch('hummingbot.connector.exchange.ascend_ex.ascend_ex_utils._time') as get_tracking_nonce_low_res_mock:
```
CrimsonJacket, line 54: Maybe it will be good to be more consistent with our unit tests and use `@patch` instead of using a `with` statement. Also since we're no longer using `get_tracking_nonce_low_res`, lets change that mock name as well.

CrimsonJacket, line 54: Maybe it will be good to be more consistent with our unit tests and use `@patch` instead of using a `with` statement. Also since we're no longer using `get_tracking_nonce_low_res`, lets change that mock name as well.

CrimsonJacket (changes requested):
Minor changes requested. Other than that, logic looks good to me.

CrimsonJacket (approved):
LGTM.

RC-13 (approved):
LGTM

PR #4815, opened 2021-11-26
Fix/trading pairs parsing

petioptrv:

This PR adds a couple of fixes for Beaxy and HitBTC to **avoid** the parsing failures exposed by the PR #4678 . It does not fix the broken parsing, but simply makes it explicit in the logs and re-enables the connectors to use the correctly parsed pairs.

Ran PMM on both exchanges successfully.

**Tips for QA testing**:
Simply checking that the connectors are usable in the client is sufficient.


petioptrv:
Addressed in #4818

PR #4829, opened 2021-11-29
fix/reorganize mexc tests

aarmoa:

I removed order book tests class because the test were running a lot of request querying the real exchange (something that should not be done in a unit test) and they were not validating anything useful in the asserts.
I have also moved test_mexc_market.py to the non unit tests folder (outside of test/hummingbot) because those test are using the mock connector to run and thus they can't be considered unit tests.

Run all unit tests.

**Tips for QA testing**:
Not required. Changes only impact tests.


CrimsonJacket (approved):
LGTM

PR #4832, opened 2021-11-29
fix/ndax test async with timeout

aarmoa:

Added a timeout to all async executions done by NDAX tests to avoid the unit tests runs from hanging git validation workflow.

Run all unit tests

**Tips for QA testing**:
No test required. Changes only impact unit tests.

### DIFF
--- a/hummingbot/connector/exchange/ndax/ndax_exchange.py
+++ b/hummingbot/connector/exchange/ndax/ndax_exchange.py
@@ -845,6 +845,9 @@ class NdaxExchange(ExchangeBase):
         for order_status in parsed_status_responses:
             self._process_order_event_message(order_status)
 
+    def _reset_poll_notifier(self):
+        self._poll_notifier = asyncio.Event()
+
     async def _status_polling_loop(self):
         """
         Periodically update user balances and order status via REST API. This serves as a fallback measure for web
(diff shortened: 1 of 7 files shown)


petioptrv (changes requested):
I think this is a good opportunity to get rid of the `0.3` second async waits. One, neater imo, way would be to add a `self._user_stream_tracker.user_stream.task_done()` at the end of `_iter_user_event_queue` method and then just await on `...user_stream.join()` in the tests. Or maybe, using the `dummy_user_stream.get.side_effect = lambda: ...` approach, the side effect can set an `asyncio.Event` that the test awaits. Something like:
```Python
    def test_exchange_logs_unknown_event_message(self):
        payload = {}
        message = {"m": 3,
                   "i": 99,
                   "n": 'UnknownEndpoint',
                   "o": json.dumps(payload)}

        self.exchange_task = asyncio.get_event_loop().create_task(
            self.exchange._user_stream_event_listener())

        done_event = asyncio.Event()

        async def return_and_set(*args, **kwargs):
            if done_event.is_set():
                raise asyncio.CancelledError
            done_event.set()
            return message

        dummy_user_stream = AsyncMock()
        dummy_user_stream.get.side_effect = return_and_set
        self.exchange._user_stream_tracker._user_stream = dummy_user_stream

        self.exchange._user_stream_tracker.user_stream.put_nowait(message)

        self.async_run_with_timeout(done_event.wait())

        self.assertTrue(self._is_logged('DEBUG', f"Unknown event received from the connector ({message})"))
```

petioptrv (approved):
Changes look good to me!

PR #4841, opened 2021-11-30
Check nonces from the networks on start up

james-hummingbot:


gateway/test/chains/ethereum/evm.nonce.test.ts
```diff
@@ -81,8 +84,10 @@ describe('unitiated EVMNodeService', () => {
 
 describe('EVMNodeService', () => {
   let nonceManager: EVMNonceManager;
+  const dbPath = '/tmp/evm-nonce2.test.level';
```
martinkou, line 87: You shouldn't expect `/tmp/` to always exist in UNIX systems. Also, the temporary files should be cleaned up after use.

Try `fs.mkdtemp()` and `fs.remove()` way as shown in `config-manager-v2.test.ts`. That ensures the generated paths are really yours and that they're cleaned up after use.

martinkou (approved):

martinkou:
Since I'm seeing the same `/tmp` problem in another patch, and the problem existing in unit tests is rather minor - I think I'm going to just merge this first, and then just create another task to fix the `/tmp` path problem.

PR #4851, opened 2021-12-02
Transaction status when in mempool

james-hummingbot:
I refactored local-storage to be a more general solution. There is a child class specifically for the nonce manager and one for the transactions. The transaction one stores the time the transaction is performed and the gas price for the transaction at the time. This information is later used to calculate the heuristic.


gateway/test/services/evm.tx-storage.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  it('save, get and delete a key value pair in the local db', async () => {
+    const testChain1 = 'ethereum';
+    const testChain1Id = 423;
+    const testChain1Tx1 =
+      '0xadaef9c4540192e45c991ffe6f12cc86be9c07b80b43487e5778d95c964405c7';
+    const testChain1GasPrice1 = 200000;
+    const testChain1Tx2 =
+      '0xadaef9c4540192e45c991ffe6f12cc86be9c07b80b43487edddddddddddddddd';
+    const testChain1GasPrice2 = 200300;
+    const dbPath = '/tmp/ethereum.tx-storage.test.level';
```
martinkou, line 14: Please change to the `fs.mkdtemp()` and `fs.remove()` way used in `config-manager-v2.test.ts` - make sure the temp path is really yours and that it's cleaned up properly after tests.

martinkou (changes requested):

gateway/test/services/evm.tx-storage.test.ts
james-hummingbot: This has been resolved

gateway/test/chains/ethereum/evm.nonce.test.ts
```diff
@@ -17,13 +20,18 @@ afterEach(() => {
...
   let nonceManager: EVMNonceManager;
-  beforeAll(() => {
-    fs.rmSync(dbPath, { recursive: true, force: true });
+  beforeAll(async () => {
+    dbPath = await fsp.mkdtemp(path.join(__dirname, '/evm-nonce1.test.level'));
     nonceManager = new EVMNonceManager('ethereum', 43, 0, dbPath);
   });
 
+  afterAll(async () => {
+    await fse.emptyDir(dbPath);
```
james-hummingbot, line 31: this may seem redundant to use emptyDir and remove, because the documentation says that `remove` is `rm(path, {force: true, recursive: true})`, but the runtime behavior does not match that statement. I keep getting an error `ENOTEMPTY: directory not empty`. Adding `emptyDir` solves it.

martinkou (approved):

PR #4858, opened 2021-12-03
Feat/add gateway create command

vic-en:


hummingbot/client/command/gateway_command.py
```diff
@@ -63,6 +68,104 @@ async def _generate_certs(self,  # type: HummingbotApplication
...
+                self._notify(f"Removing existing gateway container with id {container['Id']}...")
+                self._docker_client.remove_container(container["Id"], force=True)
+        except Exception:
+            pass  # silently ignore exception
+
+        await self._generate_certs()  # create cert if not available
+        self._notify("Pulling Gateway docker image...")
+        await asyncio.sleep(0.5)
+        try:
+            pull_logs = iter(self._docker_client.pull(gateway_docker_name, tag="gateway-v2", stream=True, decode=True))
```
martinkou, line 141: The iterator here is synchronous, which causes the `prompt-toolkit` UI library to be unresponsive during the image pull.

There are two options here:

 1. Wrap this in a thread via `run_in_executor()`, and then use `call_soon_threadsafe()` from the child thread to schedule the message back to the main thread. You can then wrap this all again into an async iterator, s.t. you can do `async for message in thread_pull_wrapper.pull(...): ...`.
 2. Use `aiokafka` instead. They also have a stream option.

martinkou (changes requested):

hummingbot/client/command/gateway_command.py
```diff
@@ -63,6 +68,104 @@ async def _generate_certs(self,  # type: HummingbotApplication
...
+            while True:
+                try:
+                    self.logger().info(json.dumps(next(pull_logs), indent=4))
```
martinkou, line 144: Need to think about what kind of log output format makes the most sense for the user - rather than just dumping the JSON to the user.

martinkou:
Architecture-wise, I think the new patch makes sense - running the docker pull in another thread while logging the lines there. I'd like to see a video of it in action before approving this.

vic-en:
@martinkou Here ==> https://youtu.be/HbnULDrh8Ag

martinkou (approved):

PR #4887, opened 2021-12-08
fix/fees beaxy

### LINKED ISSUE #2084: [BUG] Filled order fees are not recorded correctly
**Describe the bug**

Fees for orders that are filled are not recorded correctly for many connectors, instead of taking fees reported from exchange's API, the connectors call `c_get_fee` (which estimates fee). This results in incorrect fees stored on trades table (sqlite) and incorrect performance reported on `history` command.

This mistake was partly due to Binance connector had this bug which other connectors copied the code from. The bug was fixed on 29 December 2020 on Binance connector so connectors created after the date should not get affected.

Connectors which broadcast fee incorrectly on `OrderFilledEvent`:

- [ ] ~Kucoin~
- [x] Binance Perpetual Derivative
- [x] Bifinex
- [x] Coinbase Pro
- [x] Bittrex
- [ ] ~OkEx~
- [x] Huobi
- [ ] ~Beaxy~
- [x] FTX
- [ ] ~Blocktane~
- [ ] ~Coinzoom~
- [ ] ~DigiFinex~
- [x] Gate.io
- [ ] Liquid

┆Issue is synchronized with this [Clickup task](https://app.clickup.com/t/1pd1e7e) by [Unito](https://www.unito.io)

aarmoa:

The PR include the required changes to consider the fees informed by the exchange when a trade takes place, instead of estimating them.
This is more a proof of concept than a final change. Since Beaxy does not provide an event for trades (or order fills), this PR changes the fill logic to take fees from the order update event. The only problem is that in my tests the fee was always informed as being 0 (or None). But I could not validate that in the exchange because I don't have 2FA code to connect. QA please validate the fees registered by the client are in line with the fees in the exchange.
Partially closes #2084

Added new unit tests for the order fill process and the fees registration

**Tips for QA testing**:
Please check that partial and complete fills are working correctly in Beaxy connector. Also verify if the fees are now being registered correctly.
Verify if the fees registered by the client are the same fees registered by the exchange (I could not validate this because I don't have Beaxy 2FA code)

### DIFF
--- a/hummingbot/connector/exchange/beaxy/beaxy_api_user_stream_data_source.py
+++ b/hummingbot/connector/exchange/beaxy/beaxy_api_user_stream_data_source.py
@@ -63,6 +63,7 @@ class BeaxyAPIUserStreamDataSource(UserStreamTrackerDataSource):
 
     async def _listen_for_orders(self, ev_loop: asyncio.BaseEventLoop, output: asyncio.Queue):
         async for msg in self.__listen_ws(BeaxyConstants.TradingApi.WS_ORDERS_ENDPOINT):
+            self.logger().info(f"\n***\n{msg}\n***")
             output.put_nowait([BeaxyConstants.UserStream.ORDER_MESSAGE, msg])
 
     async def listen_for_user_stream(self, ev_loop: asyncio.BaseEventLoop, output: asyncio.Queue):
--- a/hummingbot/connector/exchange/beaxy/beaxy_exchange.pyx
+++ b/hummingbot/connector/exchange/beaxy/beaxy_exchange.pyx
@@ -2,25 +2,27 @@ import asyncio
 import json
 import logging
 from async_timeout import timeout
-from datetime import datetime, timedelta
 from decimal import Decimal
-from libc.stdint cimport int64_t
+from datetime import datetime, timedelta
 from typing import Any, AsyncIterable, Dict, List, Optional
 
 import aiohttp
-
 from aiohttp.client_exceptions import ContentTypeError
+from libc.stdint cimport int64_t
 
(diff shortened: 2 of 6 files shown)


nikspz:
Upon checking
Bot showing `0 USDC fees for the trade`, unmatched with the fees on Beaxy.
Also, the `performance tab shows a huge amount of return (6,69%)` for the single trade.
image
image
Same big amount of return for sell single trade
image
image

aarmoa:
Hello @nikspz. I see in the exchange screenshot the fee is negative. What does it mean? Is the exchange paying you for your operation? If that is the case they might not be reporting it as commission.
Regarding the return % I am not sure I understand the issue. If we divide the PnL by the hold portfolio value it gives that percentage. Do you think the PnL is incorrect or the portfolio value is incorrect?

RC-13:
> Hello @nikspz. I see in the exchange screenshot the fee is negative. What does it mean? Is the exchange paying you for your operation? If that is the case they might not be reporting it as commission.

The negative symbol indicates the amount deducted from the transaction.

> Regarding the return % I am not sure I understand the issue. If we divide the PnL by the hold portfolio value it gives that percentage. Do you think the PnL is incorrect or the portfolio value is incorrect?

The return is correct, there was just confusion on this part and I have clarified this to Nikita.

nikspz:
Looks like the `current BTC price's not updating`
Still showing `56107` as on prev. screenshots.
And that's could be the reason for `incorrect PnL and return %` which is bad because the bot can trigger the killswitch incorrectly.

image
image

aarmoa:
The exchange confirmed that they don't have the functionality to include the fees as part of the order update notification. We won't be able to stop estimating fees for Beaxy. I am closing this PR.

PR #4897, opened 2021-12-10
Feat/gateway v2 certs

willyPat:
Adds a bash file to create certs to use gateway as a standalone.

### DIFF
--- a/gateway/setup/create_certs.sh
+++ b/gateway/setup/create_certs.sh
@@ -1,0 +1,11 @@
+if ! command -v openssl &> /dev/null
+then
+    echo "openssl could not be found"
+    exit
+fi
+mkdir -p ../certs
+cd ../certs
+openssl genrsa -des3 -out server_key.pem
+openssl req -new -key server_key.pem -out csr.pem -subj "/C=/ST=/L=/O=/CN=localhost"
+openssl x509 -req -days 9999 -in csr.pem -signkey key.pem -out server_cert.pem
+rm csr.pem
--- a/gateway/setup/create_certs.sh
+++ b/gateway/setup/create_certs.sh
@@ -5,7 +5,15 @@ then
 fi
 mkdir -p ../certs
 cd ../certs
-openssl genrsa -des3 -out server_key.pem
+# create CA private key
+openssl genrsa -des3 -out ca_key.pem 2048
+# create CA certificate
+openssl req -x509 -new -nodes -key ca_key.pem -sha256 -days 365 -out ca_cert.pem
+# create server private key
+openssl genrsa -des3 -out server_key.pem 
+# create CSR
 openssl req -new -key server_key.pem -out csr.pem -subj "/C=/ST=/L=/O=/CN=localhost"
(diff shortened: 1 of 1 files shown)


gateway/setup/create_certs.sh
```diff
@@ -0,0 +1,19 @@
+if ! command -v openssl &> /dev/null
+then
+    echo "openssl could not be found"
+    exit
+fi
+mkdir -p ../certs
+cd ../certs
+# create CA private key
+openssl genrsa -des3 -out ca_key.pem 2048
```
martinkou, line 9: 3DES is a deprecated encryption algorithm now - even though a lot of OpenSSL guidelines still have it via copy & paste. AES128 is the recommended cipher for most applications these days. So replace the `-des3` with `-aes128`.

gateway/setup/create_certs.sh
```diff
@@ -0,0 +1,19 @@
...
+# create CA certificate
+openssl req -x509 -new -nodes -key ca_key.pem -sha256 -days 365 -out ca_cert.pem
```
martinkou, line 11: This will generate a CA certificate that expires after 1 year - which may give QA some funny issues if they keep using it for 1 year. In general, if a cert or CA is not intended to require renewal, it's better to just put `-days 9999` there.

Also, you can skip the unnecessary subject question prompts by adding `-subj "/CN=Local Hummingbot CA"`.

gateway/setup/create_certs.sh
```diff
@@ -0,0 +1,19 @@
...
+# create server private key
+openssl genrsa -des3 -out server_key.pem
```
martinkou, line 13: Ditto, use AES128 instead.

gateway/setup/create_certs.sh
```diff
@@ -0,0 +1,19 @@
...
+mkdir -p ../certs
+cd ../certs
+# create CA private key
+openssl genrsa -des3 -out ca_key.pem 2048
+# create CA certificate
+openssl req -x509 -new -nodes -key ca_key.pem -sha256 -days 365 -out ca_cert.pem
+# create server private key
+openssl genrsa -des3 -out server_key.pem 
+# create CSR
+openssl req -new -key server_key.pem -out csr.pem -subj "/C=/ST=/L=/O=/CN=localhost"
```
martinkou, line 15: You can just put `"-subj "/CN=Local Hummingbot Gateway"` and avoid a bunch of "req: No value provided for xxx" error messages.

gateway/setup/create_certs.sh
```diff
@@ -0,0 +1,19 @@
...
+# create certificate signed by CA
+openssl x509 -req -days 365 -in csr.pem -CA ca_cert.pem -CAkey ca_key.pem -CAcreateserial -out server_cert.pem -sha256
+rm csr.pem
+rm ca_cert.srl
```
martinkou, line 19: It's usually OK to keep the serial file - esp. if you intend to use the CA cert to generate more certificates later.

martinkou (changes requested):

PR #4983, opened 2022-01-08
feat/Tri Arbitrage Strategy

bsmeaton:

The Triangular Arbitrage Strategy is used to trade between three linked currency pairs on a single exchange. The strategy starts and ends with the same currency holdings and seeks to profit through the difference of exchange rates amongst the three pairs traded. For example using the three pairs FRONT/BTC, BTC/USDT and FRONT/USDT if you started with 1000 FRONT this could be exchanged for BTC which could then be exchanged for USDT and then back to FRONT. If the price of FRONT to BTC in dollar terms were offset from the price to USDT this could resultin a profit. The major hurdle here is that three trades also attract three lots of exchange fees which would greatly reduce the profitability so these must be taken into consideration within the calculation and the offset in value between various pairs must be quite extreme to profit.

 The strategy always involves three separate currencies and it allows for a starting balance in any one of these currencies or multiple. The strategy tests the resulting profitability from any of the six possible combinations of trades (starting from any of the three pairs and either going clockwise or anticlockwise around the process). Taking into account available balance the most profitable option is found. The order amount is then optimised from the vwap using the order book and if resulting profitability is above min profitability trading mode is activated. this is done in a progressive order in which the second trade is not started until the first trade has completed and so on to ensure enough balance is available. this is completed over several ticks. when the third trade is completed the system is reset and starts to monitor profitability again.

Strategy borrows and builds off existing arbitrage strategy

The strategy should operate correctly with three pairs of interlinked currencies as per example above on any exchange. 

**Tips for QA testing**:
Set min profitability to -0.5 for example to activate a trade (its not easy to find a profitable trade quickly as the strategy only sets market orders rather than limit orders when an opportunity is found).

QA testing has been carried out but essentially the outputs of the functions: (c_calculate_arbitrage_top_order_profitability) should be monitored which provides a first pass of profitability and constructs the order instructions (the three trades and sidees) and allows the strategy to progress to (c_find_best_profitable_amount) which refines the profitability and constructs the order sizes and prices for the three trades. Lastly (c_maketrade) is called in sequence during the trading mode which calls all three trades in order.

Some cleaning up around the use of the rate oracle and existing variables from the arbitrage strategy is needed at some point.

### DIFF
--- a/hummingbot/strategy/cross_exchange_market_making/Analyse_Orders.py
+++ b/hummingbot/strategy/cross_exchange_market_making/Analyse_Orders.py
@@ -1,0 +1,69 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Thu Mar 25 09:29:51 2021
+
+@author: BenjaSmea
+"""
+
+import pandas as pd
+pd.options.display.max_columns = None
+pd.options.mode.chained_assignment = None 
+
+import ccxt
+from os.path import exists
+
+
+api_keybinance = 'XXX'
+sec_keybinance = 'XXX'
+api_keykucoin = 'XXX'
+sec_keykucoin = 'XXX'
+api_keyascendex = 'XXX'
+sec_keyascendex = 'XXX'
+api_keygateio = 'XXX'
+sec_keygateio = 'XXX'
+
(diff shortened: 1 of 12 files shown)


aarmoa:
@bsmeaton please remember to follow the required steps for the community to be able to vote for your PR. If not it won't be possible to merge it (https://hummingbot.org/governance/proposals/prp/)

PR #4986, opened 2022-01-10
Feature/trade fee enhacements

petioptrv:

This PR enhances the trade fee logic to be able to more accurately estimate trade fees and model the different exchange's schemas. This, in turn, will lead to more accurate sizing of the trades by the `BudgetChecker`. One notable example is to be able to properly factor in Binance fees, which are deducted from the returns when making a buy order while not using BNB (i.e. a buy order to buy 100 USDT worth of BTC will return only 99.9 USDT worth of BTC). Another is the case of using BNB where the fee is charged in a third token.

I initially added the logic of building the `TradeFee` to the `TradeFeeSchema` itself, but this caused a circular dependency that required very hacky solutions (the fees override requires the connector settings, the connector settings requires the trade fee schema, the trade fee schema requires the fee override; plus, the fee override is built on import which requires that the connector settings are also built). This is why the build logic still lives in a separate method `build_trade_fee`. Now, that method could have just superimposed the overrides on the particular schema and then the rest of the logic lives in the schema, but I foresee community developers defaulting to directly using the schema itself and in that way not applying the overrides. To force them to use the `build_trade_fee` function, I've kept the entire trade fee building logic there.

Instead of introducing the new `build_trade_fee`, I could have introduced default values for the remainder of the order parameters in `estimate_fee`. However the new `build_trade_fee` explicitly requires the input of all of the order parameters so that this information is readily available for future developments of the trade fee building process.

- Test cases covering the new functionality.

**Tested and confirmed working by CoinAlpha QA**
Test summary:
- Run a pure market-making strategy for each centralized connector.
- Run Perpetual market making strategy to check the transaction fees on each perpetual connector.
- Run an AMM arbitrage strategy to check fee estimates for balancer and Uniswap
- Run Celo-Arb to check the fee estimate on CELO connector
- Run spot-perpetual arbitrage to check the fee transaction on Perp_Fi
- Wait for fill events to take place (full and partial fill)
- Check the fee recorded in the log pane, sqlite, csv file, and in the exchange
- Confirmed that fees charged are all matched except for connectors that use estimate fees
- Re-test PR after some fixes are pushed to address issues found on Blocktane, Kraken, Balancer, and Uniswap connectors.
- Created a Docker image based on the feature branch and start a PMM strategy to check if the intended behavior did not break.
- Test fee override for Binance connector and captured an error
- Re-test PR after new commits have been pushed.
- Confirmed all reported issues are now fixed.


Nullably (approved):
LGTM

Issue #4914, opened 2021-12-15
Avellaneda MM - Randomly appearing RunTimeWarning on the log pane


rxlxrxsx:
**Describe the bug**
When running a bot using the avellaneda market making strategy there are instances that a `RunTimeWarning` appears on the log pane. Though this warning appears on the log pane, I don't see any impact on the bot's performance. Creating a ticket for tracking.

**Steps To Reproduce**
1. Create and run an avellaneda market making strategy

**Screenshots**
<img width="1421" alt="Screen Shot 2021-12-15 at 2 31 06 PM" src="https://user-images.githubusercontent.com/64377055/146136783-109ce942-0e04-4873-99d4-d16e98355539.png">

**Release version**
development-1.0.0

**Attachments**
Logs and config: [ave_mm.zip](https://github.com/CoinAlpha/hummingbot/files/7717030/ave_mm.zip)

PR #4988, opened 2022-01-10
Fix/Avellaneda MM - Randomly appearing RunTimeWarning on the log pane

mhrvth:

- bounds specified for the scipy curve_fit parameters

Possible values of the curve parameters have to be restricted so that invalid values are not supplied to the equations. This also solves another issue where kappa and alpha become negative if the quality of collected order book data is low, either due to short buffer length settings, or other factors. The values need to be restricted to a range so that they make sense, and so that the probability density function is always approximated by an inverted logarithm.

This is a fix for the issue:
https://github.com/hummingbot/hummingbot/issues/4914

- ran the strategy with binance paper and the AXS-USDT
- ran unit tests


Nullably (approved):
LGTM

Issue #4869, opened 2021-12-06
Avellaneda MM - Order refresh tolerance parameter not working


rxlxrxsx:
**Describe the bug**
Using the order refresh tolerance parameter on avellaneda market making is not working.

**Steps To Reproduce**
1. Create and run an avellaneda market making bot
2. Set order_refresh_tolerance_pct to 10 (to easily see if orders will be retained)
3. Notice that even when the spread are still within the range of order refresh tolerance, it'll still cancel the orders and create a new set of orders.

**Screenshots**

https://user-images.githubusercontent.com/64377055/144794570-a320bb9a-588d-48d1-a47f-1715026de158.mp4

**Release version**
dev-1.0.0

**Attachments**
Logs and config: [avellaneda.zip](https://github.com/CoinAlpha/hummingbot/files/7658324/avellaneda.zip)

PR #4989, opened 2022-01-10
Fix/Avellaneda MM - Order refresh tolerance parameter not working

mhrvth:

Orders should be cancelled before their max age is reached only when a new proposal is ready. 
Because of this bug, orders were cancelled each time on order refresh time, regardless of if a new proposal is ready.

This is a fix of the issue: https://github.com/hummingbot/hummingbot/issues/4869

- ran unit tests
- ran the strategy with binance testnet
- printed debug output to confirm that orders were cancelled before their max age is reached only when the spread tolerance condition is fulfilled
- confirmed that with low tolerance values orders get cancelled
- confirmed that with high tolerance values orders don't get cancelled specifically because they don't fulfill the tolerance condition


Nullably (approved):
LGTM

PR #4993, opened 2022-01-11
Fix/Avellaneda - Verify the eta parameter

mhrvth:

While reviewing the implementation and use of the eta parameter I came across this problem.

Currently the order sizes are calculated based on the eta parameter and the amount of base asset. The problem is that it depends on the absolute amount of the base asset. That means if the client trades with less assets, it will calculate a different order size given that the target and the current state of the portfolio allocation are the same. The sizing of orders should not depend on absolute value of assets but on their relative amount in comparison to the target.

* Ran unit tests
* Ran the strategy with binance paper and setting the eta to 0.5


mhrvth:
[sc-21333]

Nullably (approved):
LGTM

PR #4998, opened 2022-01-13
Fix/command option help

mhrvth:

-h, --help options were being described in command help printouts but they didn't actually work. The options were ignored and the command was executed as if there's no additional argument. Also this option wasn't offered in command dropdowns.

This PR fixes that. -h and --help now print help for the particular command, and these options are also being displayed in dropdown menus.

Ran the client and displayed help for multiple different commands with this option.


mhrvth:
[sc-21086]

mhrvth:
[sc-21086]

Nullably (approved):

PR #5021, opened 2022-01-20
fix/Fix docker create script for CentOS/Fedora

lowitea:

On Fedora/CentOS, the `:z` flag must be used to work with selinux. And the log driver don't support the `max-size` and max-file parameters.


lowitea:
@genecheon Hi!
The CI error is not related to my MR, should I fix it anyway?

aarmoa (changes requested):
It does not make sense to change the script file because it does not work for one particular Linux version. In any case it should be adapted to use a different command only for that Linux version.

JeremyKono:

PR #5035, opened 2022-01-22
Refactor/xemm delete unused code

leastchaos:

Delete unused order_size_taker_volume_factor as it is not used in any part of the code. 

```
if len(market_pairs) < 0:
    raise ValueError(f"market_pairs must not be empty.")

if not 0 <= order_size_taker_balance_factor <= 1:
    raise ValueError(f"order_size_taker_balance_factor must be between 0 and 1.")
```

These 2 lines also seem unnecessary since they are checked at configmap but I left it in for now. 
If it's okay, I would delete it also.

run xemm strategy and confirmed working normally

**Tips for QA testing**:


leastchaos:
realized taker_volume_order_factor was intended to be used in another part of the code, but it was not. should be a bug fix and not remove.

PR #5081, opened 2022-02-01
Feature/improving tables formatting

petioptrv:

This task adds formatting to the printout tables in the client. It allows the user to set the format in the configs.

Previous:
image
New:
image

There was an issue discovered with the lexer parsing to add colour to text elements. Now fixed and test case updated.

Affected commands:
- history
- connect
- config
- orderbook
- ticker

- Test for each one of the affected commands
- Updated lexer test case

**Tested and approved by QA team**

Test Summary:
- Confirmed tables-format working as expected using the commands history, connect, config, orderbook and ticker
- Created a strategy and went through initial prompts successfully
- Confirmed wallet and exchange balances are reflecting correct
- Started the strategy without issues, confirmed orders are created and cancelled based on config
- Ran a test bot for several hours without errors
- Confirmed in logs that there is no errors when orders are filled.
- Trade information are correct (price, size) based on history command output
- Confirmed no issue installing and running it on different supported OS (linux (ubuntu), macos, windows and arm
- Confirmed no issue building on source or docker build.
-Tried several [Pretty-print tabular table format](https://github.com/astanin/python-tabulate#table-format) and its working as expected without any issue using `config tables-format`


dennisocana:
Proposal created: https://snapshot.org/#/hbot-prp.eth/proposal/0xfd8c923edaa62d62b51c242ea612c43de46fcf37a5c76a571700785b014a3674

aarmoa (changes requested):
I agree with the changes. The new functionality will allow the client to present more readable and clear tables.
But in my opinion the new `tabulate` dependency should be installed using conda instead of pip (we use pip only for dependencies that are not included in conda)

petioptrv:
> I agree with the changes. The new functionality will allow the client to present more readable and clear tables. But in my opinion the new `tabulate` dependency should be installed using conda instead of pip (we use pip only for dependencies that are not included in conda)

@aarmoa, I agree, and I've addressed that in the latest commit.

PR #5087, opened 2022-02-03
(fix) Adds an initial ping to user stream ws connection to update last recv time.

petioptrv:

Binance perpetual takes some time before it delivers the first ws message to the user stream. This keeps the last recevied timestamp zero for a while, which does not allow the strategy to start execution. In addition, Binance spot does not check if the user-stream is ready before marking itself as being ready.

This PR adds a check to Binance spot to determine if the user-stream is ready. It also adds an initial ping to both spot and perpetual to kick-start the readiness checks.

Added a simple check to existing tests to determine if the ping was issued on user-stream startup.

**Tips for QA testing**:
Check that both binance spot and perpetual start trading with minimal delay (below 5 seconds).


petioptrv:
Closing - opened by mistake first towards foundation repo

PR #5092, opened 2022-02-04
Fix/RingBuffer's c_get_as_numpy_array() doesn't return the correct array

mhrvth:

After adding a sample to the RingBuffer, start and stop indexes are updated. When calling the c_get_as_numpy_array() afterwards, the method doesn't account for the shifted indexes (in case the buffer is already filled).
As a result it returns an array with wrong data.

* Printed c_get_as_numpy_array()'s output and compared it to the actual content of the buffer and the start and stop indices
* Ran unit tests

### DIFF
--- a/hummingbot/strategy/__utils__/ring_buffer.pxd
+++ b/hummingbot/strategy/__utils__/ring_buffer.pxd
@@ -5,13 +5,12 @@ cimport numpy as np
 cdef class RingBuffer:
     cdef:
         np.float64_t[:] _buffer
-        int64_t _start_index
-        int64_t _stop_index
+        int64_t _delimiter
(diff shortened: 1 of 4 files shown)


dennisocana:
Proposal created: https://snapshot.org/#/hbot-prp.eth/proposal/0xc5afb105d02a1e75a1d14131de8121000884347333adad2346cec1876ce70599

Nullably (approved):

rapcmia (approved):
Test performed:
- Run unit tests based on PR's file changes, ok
- Successfully run nosetests for:
    - test_ring_buffer.py
    - test_instant_volatility.py

PR #5128, opened 2022-02-21
hotfix/fixed install from source scripts to match hummingbot repo

phbrgnomo:

Hotfix to change the install-from-source scripts to the correct github repository

**Tips for QA testing**:


RobHBOT (approved):
ok per Paulo, It's the same change on different branches. Since we already had it on staging, if we don't change development, it will override the next time we do development -> staging , and revert back to coinalpha repo. This is necessary code

PR #5138, opened 2022-02-24
Refactor/preventing duplicate client order ids

petioptrv:

Duplicated client order IDs can cause orders to be rejected by the exchange and can cause issues with historical analysis of strategy performance. Currently, the client uses a nonce derived from the current timestamp to generate "unique" client order IDs. The issue is that those IDs can be duplicated across multiple instances of `hummingbot` running in parallel. To prevent this from happening, this PR introduces a new method of generating the IDs that will be unique across instances as well.

The two main changes are:
- Propagates the Binance approach of ensuring unique client order Ids into the rest of the CoinAlpha-maintained exchanges with slight modifications.
- Adds a default function to use for that purpose with options for adding an exchange-specific prefix and limiting the length of the ID string.

### The new order ID schema
- Optional exchange-specific `hbot_order_id_prefix`
- Order `side` ["B", "S"]
- The `base_str`, which is the first and last letter of the base token
- The `quote_str`, which is the first and last letter of the quote token
- Tracking nonce (low-res version, with milliseconds)
- The `client_instance_id`, which are generated so that they differ between instances on the same machine, but also between instances on different machines (eng: see concerns below)

- Tests for the new function
- Tests for all of the exchanges affected:
    - AscendEx
    - Binance spot
    - Binance perp
    - Bybit perp
    - Gate.io
    - Huobi
    - KuCoin
    - OKEX
    - ProBit
- Successfully placed orders on the following exchanges (expired creds for ByBit perp; invalid ones for Huobi and OkEx):
    - AscendEx
    - Binance spot
    - Binance perp (testnet)
    - Gate.io
    - KuCoin
    - ProBit

**Tested and confirmed working QA.**

Tests performed:
- Created a strategy using the affected connectors and went through initial prompts successfully
- Started the strategy without issues, confirmed orders are created and cancelled based on config
- Confirmed that the orders created now uses the new order ID schema
- Confirmed that no duplicate client order IDs were created when running between instances on the same and different machine
- Trade information are correct (price, size and order ID) based on history command output, sqlite and csv file
- Confirmed all issues reported are now fixed


JeremyKono:
Hi @petioptrv. Please resolve branch conflicts in this PR. Thanks!

petioptrv:
@JeremyKono, dues to the fact that this branch was originally branched off of a feature branch, resolving the issues has been a nightmare, so I created a new PR that [branched](https://github.com/hummingbot/hummingbot/pull/5161) directly off of `development` instead.

petioptrv:
Closing this PR, and re-creating it [here](https://github.com/hummingbot/hummingbot/pull/5161)

Issue #5170, opened 2022-03-08
Binance Perpetual InFlightOrder returns wrong executed amount for partial fills


cardosofede:
When filling an order with Binance perpetual connector, partial fills are summed but last partial fill execution_amount in InFlightOrder is equal to the total original amount. This leads to incorrect total executed amount in client order tracker. Hedge Strategy relies in order_filled_event.amount which is wrong because of InflightOrder bug.

<img width="1498" alt="Screen Shot 2022-03-08 at 18 09 22" src="https://user-images.githubusercontent.com/36869960/157326917-abc2aa91-0208-400f-bf29-36cbb5599503.png">

[analisis_hedge.txt](https://github.com/hummingbot/hummingbot/files/8209728/analisis_hedge.txt)

dennisocana:
Our team will start working on a fix for this bug today.

rapcmia:
Fix has been applied version 1.3.0

PR #5183, opened 2022-03-14
Fix/binance perpetual order fill amount bug

aarmoa:

Fix to the issue https://github.com/hummingbot/hummingbot/issues/5170 by correctly parsing the trade fill event from the connector and reading the fill amount for the particular trade instead of the accumulated quantity field.

All unit tests passing in green

**Test summary**:
- Install feature branch and test the connector using perpetual market-making strategy.
- Leave the bot running to capture a partial fill both for open and close positions
- Checked the total order amount of the partially filled orders by running the history verbose command and confirmed that the total amount is equal to the specified order amount value.
- Leave the bot running for more than 24 hours and inspect the filled record after, 4 partial fill took place, and confirm that the total order amount is equal to the specified value in the order_amount parameter.
- Review the logs and confirmed that 
  -  When a partial fill takes place, then the remaining get filled right after, the total amount is correct
   - When partial fill takes place then the remaining gets canceled, the next order created will have a size equal to the remaining unfilled order to make sure that the total amount is still correct. 

**Tips for QA testing**:
To validate the fix is working it is important to check the total amount filled for an order when there is at least one partial fill.

This PR comes from https://github.com/CoinAlpha/hummingbot/pull/135


aarmoa:
Created proposal: https://snapshot.org/#/hbot-prp.eth/proposal/0xe84f7fb19be1708b1e99f3ae4ffe0787aa6bb322a42a8a1b6c15b75cccb47fd3

rapcmia (approved):
Test performed:
- Setup perpetual market-making on Binance, ok 
- Set tight spreads enough for trades and monitor on partial orders 
- Partial orders are recorded correctly and data matched on History command, CSV and exchange's trade history 
- Partial orders are also tested on Binance_perpetual_testnet, ok 
- Review logs and found no errors
- There was an instance where the partial filled order reaches the set order refresh time, the new created order uses the same order price and use the remaining unfilled order amount.
	- Set order_amount is 30 XRP
	- Initial partial fill was 1.1 XRP then order gets cancelled/refreshed
	- The new open order used the same order price and use the remaining unfilled order 28.9 which eventually filled

PR #5222, opened 2022-04-11
fix/xemm decimal invalid operation error when too much quote balance

andrewsiah:

Cross Exchange market making returns a "Decimal Invalid Operation" error when quote balance is too high. This is due to "taker_market.c_get_price_for_quote_volume()" being called with quote balance, and this function returns NaN when quote balance is too high.

This bug can be replicated by using paper trade. Setting balance paper to USDT 1000000, and running xemm on an illiquid market (thin order book).

Tested live.

**Tips for QA testing**:


hummingbot/strategy/cross_exchange_market_making/cross_exchange_market_making.pyx
```diff
@@ -847,6 +847,9 @@ cdef class CrossExchangeMarketMakingStrategy(StrategyBase):
 
             taker_slippage_adjustment_factor = Decimal("1") + self._slippage_buffer
             taker_balance = taker_balance_in_quote / (taker_price * taker_slippage_adjustment_factor)
+            if Decimal.is_nan(taker_price):
+                taker_balance = maker_balance
```
leastchaos, line 851: Hi, sorry I don't think it is a good idea to set taker_balance = maker_balance here just because of too much quote balance. 
For example, if base contains 1k token while quote contain 100 token (converted from quote amount) and the order book depth only has 90 tokens. this would still result in a order size of 1k token even though the taker could only get 100 token. This order would then immediately get canceled when c_check_if_sufficient_balance is called.

This issue is mainly due to the taker_market order depth does not have sufficient depth due to large balance which we can limit by using the user_order instead. we can also change to using an existing function here on line 841.

`self.c_calculate_effective_hedging_price(market_pair, is_bid, user_order)`

hummingbot/strategy/cross_exchange_market_making/cross_exchange_market_making.pyx
```diff
@@ -1163,7 +1166,9 @@ cdef class CrossExchangeMarketMakingStrategy(StrategyBase):
                 taker_trading_pair, True, quote_asset_amount
```
leastchaos, line 1166: Since we actually already have the active_order price, I think we can use that instead of recalculating the price instead 
`taker_price = order_price`

leastchaos, line 1166: `taker_price = order_price`

hummingbot/strategy/cross_exchange_market_making/cross_exchange_market_making.pyx
```diff
@@ -1163,7 +1166,9 @@ cdef class CrossExchangeMarketMakingStrategy(StrategyBase):
...
             ).result_price
             adjusted_taker_price = taker_price * taker_slippage_adjustment_factor
-            order_size_limit = min(base_asset_amount, quote_asset_amount / adjusted_taker_price)
+            order_size_limit = base_asset_amount
```
leastchaos, line 1169: I think setting order_size_limit here as base_asset_amount will remove the intention of this method. This method is to check for sufficient balance on both the maker and taker account. By ignoring due to no taker price, we are not checking for sufficient taker balance.
So it may be better if we use the order price directly on line 1165

leastchaos, line 1169: I think setting order_size_limit here as base_asset_amount will remove the intention of this method. This method is to check for sufficient balance on both the maker and taker account. By ignoring taker balance due to no taker price, we are not checking for sufficient taker balance.
As we already have the order price in this function, it may be better if we use the order price directly on line 1165 to avoid this error.

hummingbot/strategy/cross_exchange_market_making/cross_exchange_market_making.pyx
leastchaos: sorry, my bad, the check_and_hedge function uses the following method when calculating the price to hedge so we should use the same in order not to result in insufficient balance.
```
            order_price = taker_market.get_price_for_volume(taker_trading_pair, True,
                                                            quantized_hedge_amount).result_price
```
so we should probably change this to calculate based on the actual order quantity instead of the quote asset amount which is a much larger sum.

`taker_price = taker_market.get_price_for_volume(taker_trading_pair, True, active_order.quantity).result_price`

hummingbot/strategy/cross_exchange_market_making/cross_exchange_market_making.pyx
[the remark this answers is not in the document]
andrewsiah: Those are great ideas. Would you mind making a new PR that implements your suggestions? I can pull mine down.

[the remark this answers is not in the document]
leastchaos: Hey, sorry I am only acting as a reviewer for this PR and do not intend on creating a new PR on the same issue.

JeremyKono:
Hi @andrewsiah, please make the changes in line with the comments.

hummingbot/strategy/cross_exchange_market_making/cross_exchange_market_making.pyx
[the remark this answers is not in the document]
mifeng: @andrewsiah please make the changes that @leastchaos proposed by April 30, otherwise this fix won't be included in the next release and will be rolled to the next month's release.

mifeng:
@JeremyKono Let's just close this ticket. The discussion is taking longer than the actual code changes which only affect a few lines of code.

If this is a bug that users really care about, they can re-open this ticket.

JeremyKono:

PR #5223, opened 2022-04-11
Fix/TWAP & TWAP Plus: `has_enough_balance()` doesn't consider order amount, but a total `target_asset_amount`

mhrvth:

In the TWAP and TWAP Plus strategy the method `has_enough_balance()` works with the `target_asset_amount` and not an order amount when it's supposed to check whether it's possible to place an order.

Ran the strategy with kucoin paper and SHR-USDT.
Ran unit tests

**Tips for QA testing**:

When specifying a very big `target_asset_amount` but a "normal" order amount the strategy shouldn't thrown an exception, but should place an order.


mhrvth:
Created by accident, closed for now. Will be reopened later.

Issue #5251, opened 2022-04-24
Binance market order without price fail with illegal characters


lt7:
**Describe the bug**
New market orders to Binance without a price (which is optional) fail with illegal characters because price is being submitted as string NaN.

**Steps To Reproduce**
Create a custom strategy that utilises market orders and attempt to submit a market order without a price e.g

`if side == 'buy':
            order_id = self.buy_with_specific_market(
                market_trading_pair_tuple=self._market_info,
                amount=self._order_amount,
                order_type=OrderType.MARKET
            )`

Here are the steps to reproduce the issue (see attachments in the section below):
1. Set up your configuration file with these parameters: ...
2. Start the bot using this command: ...
...

**Release version**
v1.3.0

**Error Message**
`{2022-04-24 03:02:10,118 - 28767 - hummingbot.connector.exchange.binance.binance_exchange - NETWORK - Error submitting SELL MARKET order to Binance for 0.00020000 BTC-AUD NaN.
Traceback (most recent call last):
  File "/opt/hummingbot/hummingbot/connector/exchange/binance/binance_exchange.py", line 526, in _create_order
    order_result = await self._api_request(
  File "/opt/hummingbot/hummingbot/connector/exchange/binance/binance_exchange.py", line 1010, in _api_request
    return await web_utils.api_request(
  File "/opt/hummingbot/hummingbot/connector/exchange/binance/binance_web_utils.py", line 113, in api_request
    raise IOError(f"The request to Binance failed. Error: {error_response}. Request: {request}")
OSError: The request to Binance failed. Error: {"code":-1100,"msg":"Illegal characters found in parameter 'price'; legal range is '^([0-9]{1,20})(\\.[0-9]{1,20})?$'."}. Request: RESTRequest(method=POST, url='https://api.binance.com/api/v3/order', params=None, data={'symbol': 'BTCAUD', 'side': 'SELL', 'quantity': '0.00020000', 'type': 'MARKET', 'newClientOrderId': 'x-XEKWYICXSBCAD1805984ff5179a47b1306', 'price': 'NaN'}, headers={'Content-Type': 'application/json'}, is_auth_required=True, throttler_limit_id='/order')
[...]

PR #5252, opened 2022-04-24
fix/binance_market_order_no_price

lt7:

Fixes Binance market order without price fails with illegal characters #5251

Added new unit test for market orders

**Tips for QA testing**:

Create custom strategy with market orders for binance to test.

### DIFF
--- a/hummingbot/connector/exchange/binance/binance_exchange.py
+++ b/hummingbot/connector/exchange/binance/binance_exchange.py
@@ -463,7 +463,8 @@ class BinanceExchange(ExchangeBase):
         trading_rule: TradingRule = self._trading_rules[trading_pair]
         price = self.quantize_order_price(trading_pair, price)
         quantize_amount_price = Decimal("0") if price.is_nan() else price
-        amount = self.quantize_order_amount(trading_pair=trading_pair, amount=amount, price=quantize_amount_price)
+        if order_type in (OrderType.LIMIT, OrderType.LIMIT_MAKER):
+            amount = self.quantize_order_amount(trading_pair=trading_pair, amount=amount, price=quantize_amount_price)
 
         self.start_tracking_order(
             order_id=order_id,
@@ -501,10 +502,10 @@ class BinanceExchange(ExchangeBase):
                       "side": side_str,
                       "quantity": amount_str,
                       "type": type_str,
-                      "newClientOrderId": order_id,
-                      "price": price_str}
+                      "newClientOrderId": order_id}
         if order_type == OrderType.LIMIT:
             api_params["timeInForce"] = CONSTANTS.TIME_IN_FORCE_GTC
+            api_params["price"] = price_str
 
         try:
             order_result = await self._api_request(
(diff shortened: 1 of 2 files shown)


lt7:
https://snapshot.org/#/hbot-prp.eth/proposal/0xfc400d06f751f735c4f622def8fb73dee21365001e31b7099d8e31b6edefb969

aarmoa:
Hello @lt7. Please take a look at the method `supported_order_types`. Binance connector was developed to not use orders without prices (ie: market order types). That means that orders should never be created without price.
Besides that, the fix implementation is incorrect because in the case a price is not provided the order amount is not quantized, what is an error because we always need to ensure the client creates orders with an amount considered valid by the exchange trading rules.

-- Hummingbot Foundation Technical DAO review

lt7:
> Hello @lt7. Please take a look at the method `supported_order_types`. Binance connector was developed to not use orders without prices (ie: market order types). That means that orders should never be created without price. Besides that, the fix implementation is incorrect because in the case a price is not provided the order amount is not quantized, what is an error because we always need to ensure the client creates orders with an amount considered valid by the exchange trading rules.
> 
> -- Hummingbot Foundation Technical DAO review

Thanks @aarmoa, I missed `supported_order_types` and https://github.com/hummingbot/hummingbot/blob/master/hummingbot/connector/exchange/binance/binance_exchange.py#L466 comment seemed to indicate it was supported, I'll update this method or are there some other concerns?

Good point about the order amount not quantized, that was a miss as https://github.com/hummingbot/hummingbot/blob/master/hummingbot/connector/exchange/binance/binance_exchange.py#L346 self.get_price was timing out in the unit test, any pointers/suggestions how to mock that ?

aarmoa:
@lt7 I would suggest to not use market orders. They introduce the risk of being executed at prices far away from the intended trading price in high volatility scenario. Please remember that you can simulate a market order by sending a limit order with a price equal (or even lower/higher) to the best bid price or best ask price. But since the order includes a price you will be sure that the market cant execute your order at unexpected prices.

lt7:
thanks for the feedback @aarmoa. I'm building a new strategy that had a use-case for market, stop_loss, take_profit etc so thought I'd PR a fix, but no worries if there is no interest. 
 
> @lt7 I would suggest to not use market orders. They introduce the risk of being executed at prices far away from the intended trading price in high volatility scenario. Please remember that you can simulate a market order by sending a limit order with a price equal (or even lower/higher) to the best bid price or best ask price. But since the order includes a price you will be sure that the market cant execute your order at unexpected prices.

JeremyKono:
Will be closing this PR as it is not ideal to be in the codebase.

mifeng:
Personally, I disagree with the approach that exchange connectors shouldn't support market orders. CoinAlpha removed market orders from connectors back in 2020, but in hindsight I think it was a hasty decision that didn't benefit the open source community. I don't think connectors should be opinionated and dictate what strategy builders can and can't do. Instead, they should just map the API of the exchange and give strategy builders full power. I would support and vote for an improvement proposal that adds market order support to the main connectors.

Issue #5276, opened 2022-05-01
Dydx perpetual - Unexpected error occured when listening to order book streams


mancap314:
**Describe the bug**
When running a `perpetual_market_making` strategy on dydx, I get the error:
```
Unexpected error occured when listening to order book streams
```
This error is recurring, but apparently does not prevent the bot to work.

**Steps To Reproduce**
1. Configure the dydx connector (with API and stark key)
2. create a `perpetual_making_strategy`
3. start this strategy

**Release version**
Docker on Ubuntu, version 1.3.0

**Attachments**
```
2022-05-01 21:11:32,846 - 1 - hummingbot.connector.derivative.dydx_perpetual.dydx_perpetual_api_order_book_data_source - ERROR - Unexpected error occurred when listening to order book streams. Retrying in 5 seconds...
Traceback (most recent call last):
  File "/home/hummingbot/hummingbot/connector/derivative/dydx_perpetual/dydx_perpetual_api_order_book_data_source.py", line 180, in listen_for_subscriptions
    async for ws_response in ws.iter_messages():
  File "/home/hummingbot/hummingbot/core/web_assistant/ws_assistant.py", line 69, in iter_messages
    response = await self._connection.receive()
  File "/home/hummingbot/hummingbot/core/web_assistant/connections/ws_connection.py", line 62, in receive
    msg = await self._read_message()
  File "/home/hummingbot/hummingbot/core/web_assistant/connections/ws_connection.py", line 79, in _read_message
    msg = await self._connection.receive(self._message_timeout)
  File "/home/hummingbot/miniconda3/envs/hummingbot/lib/python3.8/site-packages/aiohttp/client_ws.py", line 216, in receive
    raise RuntimeError("Concurrent call to receive() is not allowed")
RuntimeError: Concurrent call to receive() is not allowed
```

*WARNING: Do NOT publish any exchange API keys or your wallet's private key. Whoever has access to them may steal your assets!*

**Optional: your discord username:**
manuelc#5151

sirEven:
Can confirm the issue on macOS running 1.3.0 from source.

phpmac:
Have you found a solution?

sirEven:
> Have you found a solution?

Error still persists on my end (just tested on version 1.5.0)

afkj0hn:
Hi, having the same exact issue as @mancap314 and @sirEven. 

Running latest release - version 1.6.0 on macOS from source.

phpmac:
[...]

PR #5295, opened 2022-05-07
ignore runtime error in case of concurrent call to receive()

mancap314:

- [ x] Your code builds clean without any errors or warnings
- [ x] You are using approved title ("feat/", "fix/", "docs/", "refactor/")

Related to [issue 5276](https://github.com/hummingbot/hummingbot/issues/5276) and after quite extensive investigation (I am just starting on this project...) this error message comes from concurrent call to `receive()` in the `aiohttp.ClientWebSocketResponse` instance (see [line raising error](https://github.com/aio-libs/aiohttp/blob/e485c4d5f8e02ab0c68059236608fc13ba194daa/aiohttp/client_ws.py#L238)). If its class variable `_waiting` is not None, means if it's busy, it can't take call.
Since:
- there are many calls to this `receive()` from independent data sources within the dydx connector
- refactoring those calls to prevent those concurrent calls is quite inextricable and that every (of the many...) attempts lead to quite massive side effect
- this error is not blocking at all and completely harmless
I thought, just passing it is enough.

(There are also minor linting changes introduced by [black](https://pypi.org/project/black/))

Re-running the Docker container with the `create.sh` script, use the same `dydx_perpetual_connector` and the same `perpetual_making_strategy`, observe that this message does not pop up anymore

**Tips for QA testing**:

### DIFF
--- a/hummingbot/core/web_assistant/connections/ws_connection.py
+++ b/hummingbot/core/web_assistant/connections/ws_connection.py
@@ -1,11 +1,9 @@
 import asyncio
 import time
-from typing import (
-    Dict,
-    Optional,
-)
+from typing import Dict, Optional
 
 import aiohttp
+
 from hummingbot.core.web_assistant.connections.data_types import WSRequest, WSResponse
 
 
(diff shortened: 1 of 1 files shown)


aarmoa:
Hello @mancap314. If there are concurrent calls to `receive()` that is an error and the connector (or logic) doing those calls should be fixed. The exception should not be trapped because that will hide the real issue.

-- Review provided in representation of the Hummingbot Foundation Technical DAO

JeremyKono:

PR #5296, opened 2022-05-09
Add a flag for start_command, if true do not let start_command run

james-hummingbot:

- Add a flag to `start_command` to monitor when it can be run, if `start_command` is called while the flag is true, it will do nothing and let the user know that `start` is already running.

Manually tested different cases when the start command should work, and when it should inform the user that the start command is already in progress.

**Tips for QA testing**:


rapcmia (approved):
Test performed:
- Clone feature branch and started client ok
- Generate certs and gateway create
- Setup infura and simple amm-arb using kovan testnet
- Enter start multiple times, the client will prompt that the `bot has already running..`
- Tested on source/docker and AWS instance, all ok

PR #5327, opened 2022-05-20
(fix) Script (Lite) Strategy incorrect subclass import

MementoRC:

The creation of the class for a new subclass of ScriptStrategyBase incorrectly select the base class ScriptStrategyBase when the getmembers() ordering puts the new class lower than ScriptStrategyBase.
This is because in the next() lookup function because issubclass(cls, cls) = True (a class is a subclass of itself in Python)

Passed the failed import of the user local class ScriptStrategyReload, which fails when hummingbot attempts to initialize the markets, which in ScriptStrategyBase is None

**Tips for QA testing**:
Updated the existing Unittests case with a loading of the test_dca_example.py

### DIFF
--- a/hummingbot/strategy/script_strategy_base.py
+++ b/hummingbot/strategy/script_strategy_base.py
@@ -57,7 +57,7 @@ class ScriptStrategyBase(StrategyPyBase):
         script_module = importlib.import_module(f".{script_name}", package=SCRIPT_STRATEGIES_MODULE)
         try:
             script_class = next((member for member_name, member in inspect.getmembers(script_module)
-                                 if inspect.isclass(member) and issubclass(member, cls)))
+                                 if inspect.isclass(member) and issubclass(member, cls) and member_name != cls.__name__))
         except StopIteration:
             raise InvalidScriptModule(f"The module {script_name} does not contain any subclass of ScriptStrategyBase")
         return script_class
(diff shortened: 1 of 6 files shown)


aarmoa:
Hello @MementoRC. It seems like your configuration is not the correct one. You should not be defining new script classes in the same module where ScriptStrategyBase class is defined. When you create new script classes in their own modules then getmembers() will never return the superclass, because it is not part of that module.
Also, please don't commit new script strategies to the project. We are not expecting the community to submit new script strategies to the baseline.

MementoRC:
I am not defining it in the same module, I did debug and see to my surprise that ScriptStrategyBase was picked up
It seems to me that the ScriptStrategyBase will be defined in the module due to:
from hummingbot.strategy.script_strategy_base import ScriptStrategyBase
My Lite Strategy is very basic. simply importing the base class and running the test. I added the test_dca... because I needed to put a testcase in place and hummingbot looks into the scripts directory to find the Lite scripts

I will look into this more carefully
Thank you

MementoRC:
Please look at the screenshot below of the debugger. The module does include ScriptStrategyBase, and a printout of inspect.getmembers(script_module) has ScriptStrategyBase as a member
Screenshot from 2022-05-21 16-07-26

MementoRC:
As you can see, when the PR is removed, the test fails:
image

PR #5360, opened 2022-06-01
gateway create generates necessary traderjoe yml file

james-hummingbot:

`gateway create` on `development` will fail because the bash script does not copy traderjoe.yml. This fixes it.

This also fixes the level db issue. It changes the default directory for the database files to somewhere writable.

This is an urgent bug fix for development!

**Tips for QA testing**:

build the gateway docker, run `gateway create`, make sure it works normally


rapcmia:
Hi @james-hummingbot confirmed that gateway now generates the traderjoe.yml file on `gateway/conf` folder for both macOS and linux source build. However we would still need to check it on docker build and waiting for the fix on the issue on dev-branch so we can make sure it records as well the file when recreating the certs from the client. Test still in progress for the docker build

Screenshot 2022-06-01 170748

dennisocana:
[sc-28354]

dennisocana:
[ch-28354]

Issue #5129, opened 2022-02-22
Gate_io: Unable to pull fee if native token is use as payment for fee


RC-13:
**Describe the bug**
The client is unable to fetch the trade fee when the native token (GT) is used to pay for fees. The CSV file or SQLite file stores 0 amount of fees instead.

**Steps To Reproduce**
Here are the steps to reproduce the issue (see attachments in the section below):
1. Enable GT as payment for a fee in your account, purchase some if none
2. Create a pure market making strategy using Gate_io
3. Wait for a fill order
4. Run history and check the fees paid
5. Open the CSV file or SQLite file and check the fee table
...

**Screenshots**
The first fill in the image below happens before GT payment is enabled, after GT payment is enabled, the second fill took place. 
![image](https://user-images.githubusercontent.com/57189990/155085704-7279835e-7cfe-4096-9694-f7568803611a.png)
![image](https://user-images.githubusercontent.com/57189990/155086084-a3617ebf-74ba-42f0-bfa8-a9f54551bdb6.png)
![image](https://user-images.githubusercontent.com/57189990/155086558-d4615b8e-5cef-4eb7-b9a5-4ea20b46885c.png)

**Release version**
Hummingbot 1.0

**Attachments**
// Please attach your config file and log file to help us debug. It would be difficult for us to help you without those! 

*WARNING: Do NOT publish any exchange API keys or your wallet's private key. Whoever has access to them may steal your assets!*

**Optional: your discord username:**

nikspz:
The issue still exists on dev 1.6.0.
![image](https://user-images.githubusercontent.com/83953535/174130023-8b2d5f78-ae25-4fe9-a552-01ae23ec950a.png)
![image](https://user-images.githubusercontent.com/83953535/174130130-1fb70d84-e4ea-4a73-91d7-b2c5577d6a89.png)
![image](https://user-images.githubusercontent.com/83953535/174130371-3402d4a5-88cb-4ba2-a59e-8c45e27deb04.png)

aarmoa:
This is not an issue, it is a configuration mistake. The client should be configured to use a source for the RateOracle that is able to provide the FX for the tokens involved in the strategy execution. In this case the RateOracle should be configured to use CoinGecko, that provides price information for GT: https://coingecko.com/es/monedas/gatetoken

nikspz:
Reproduced on dev-2.2.0

[...]

PR #5414, opened 2022-06-17
(fix) gate_io native token as fee

MementoRC:

https://snapshot.org/#/hbot-prp.eth/proposal/QmR21ZyXJ7a4VqUC1SX3ensZ747UaymMn57mV59zQw4YEr
Providing a workaround to the issue:
https://github.com/hummingbot/hummingbot/issues/5129
Since GT is not on Binance, the fee GT-USDT price is not found and that generates a warning. The fee is thus not converted into USDT (not sure it would be though)

Ran the updated bot and waited for a trade, then ran history ( no change to the conf_global.yml or conf_fee_override.yml)
image

**Tips for QA testing**:


MementoRC:
PRP: https://snapshot.org/#/hbot-prp.eth/proposal/QmR21ZyXJ7a4VqUC1SX3ensZ747UaymMn57mV59zQw4YEr

MementoRC:
I will try to update the test suite

hummingbot/connector/exchange/gate_io/gate_io_exchange.py
```diff
@@ -420,14 +420,17 @@ def _process_trade_message(self, trade: Dict[str, Any], client_order_id: Optiona
             self.logger().debug(f"Ignoring trade message with id {client_order_id}: not in in_flight_orders.")
             return
 
+        # When GT token is used to pay fee 'gt_fee' is provided by the API
+        if 'gt_fee' in trade and Decimal(trade["gt_fee"]) > 0:
+            flat_fees = [TokenAmount(amount=Decimal(trade["gt_fee"]), token='GT')]
+        else:
+            flat_fees = [TokenAmount(amount=Decimal(trade["fee"]), token=trade["fee_currency"])]
+
```
aarmoa, line 428: The API documentation does not discard the case where fee is paid both in GT and in another token. We should prepare the logic for that scenario, by adding a flat fee for GT fee (if there is any GT fee) and a flat fee for other token fee (if there is and it is not 0).

-- Reviewed in representation of Hummingbot Foundation Technical DAO

aarmoa (changes requested):

hummingbot/connector/exchange/gate_io/gate_io_exchange.py
MementoRC: Gate.io support had replied to my ticket related to this, that if I set the 'Pay with GT' flag, I cannot use my points, thus I did not think about that. Thanks for pointing that out, in Gate.io case, this also requires to check and add the 'point_fee'
Now the display will be tricky as fee is only displayed in 1 currency or should this feature be also supported by the 'history' output?

cardosofede: Hi @aarmoa @MementoRC,

Today we discuss this issue in the Hummingbot Foundation Meeting and we have a proposal for this situation, but we would like to let the decision open to the community.
We think that the fees has to admit payments in the native token of the exchanges (like GT in this case) or base/quote, but POINTS are only used in Gate.io so we think we don't need to include it.

MementoRC:
Should this remain open open?

MementoRC:
Actually, this seems obsolete

PR #5429, opened 2022-06-23
Fix/huobi update event bug

petioptrv:

This PR addresses a bug in Huobi where, on update event of an untracked order, the code will try to reference variables that are not yet defined.

- Unit test added to fix this issue.

**Tips for QA testing**:

- To test the fix, start a strategy using Huobi and then place an order on the exchange that will fill. Observe if the strategy fails because of this untracked order. Before doing that, please test this approach on `development`, where the strategy should currently fail.


carlitogetaladajr:
https://snapshot.org/#/hbot-prp.eth/proposal/0x180728155776bfe4c96282f816f934c330a2f979ae5b60e1e411beab70a2a5e9

PR #5431, opened 2022-06-24
(fix) Rate command fails with pair within quotes

MementoRC:

The rate command from CLI returns 'Rate is not available' if the pair is enclosed in quotes (single, or double). This is because the pair is captured literally as a string and passed to the show_rate.
The fix proposed is to add an eval in the oracle_rate_msg since that is where the 'upper' is applied, i.e. formatting of the pair to meet the required format, but another solution can be investigated

Tested in the CLI with 'BTC-USDT', "BTC-USDT" and BTC-USDT

**Tips for QA testing**:

### DIFF
--- a/hummingbot/client/command/rate_command.py
+++ b/hummingbot/client/command/rate_command.py
@@ -1,10 +1,9 @@
-from decimal import Decimal
 import threading
-from typing import (
-    TYPE_CHECKING,
-)
-from hummingbot.core.utils.async_utils import safe_ensure_future
+from decimal import Decimal
+from typing import TYPE_CHECKING
+
 from hummingbot.core.rate_oracle.rate_oracle import RateOracle
+from hummingbot.core.utils.async_utils import safe_ensure_future
 from hummingbot.exceptions import OracleRateUnavailable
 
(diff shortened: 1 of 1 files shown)


hummingbot/client/command/rate_command.py
```diff
@@ -39,7 +38,7 @@ async def show_rate(self,  # type: HummingbotApplication
     @staticmethod
     async def oracle_rate_msg(pair: str,
                               ):
-        pair = pair.upper()
+        pair = eval(pair.upper())
```
aarmoa, line 41: It does not make sense at all to evaluate a trading pair as if it were a function.

aarmoa (changes requested):
-- Reviewed in representation of the Hummingbot Foundation Technical DAO

hummingbot/client/command/rate_command.py
MementoRC: This seemed like a common technique to remove extra single and double quotes from string. Would you suggest to use json instead. This may not be something you guys want to 'fix' to beginning with

PR #5451, opened 2022-07-01
(feat) Implements the Eve exchange connector

petioptrv:

This PR introduces the Eve exchange according to the latest standards. The PR also prepares the code base to easily implement additional `AlphaPoint` exchanges.

- Implemented all tests, and passing.

**Tips for QA testing**:

- AlphaPoint does not provide trade data for the order book, so the last trade price is not available.
- The Eve exchange is new and the volume on the order books is low. Be mindful that this can sometime produce edge cases that can still be considered bugs, but may not be related to the changes proposed in this PR. One such example is using XEMM on an order book that does not have sufficient volume to cover the order. This causes XEMM to fail. CoinAlpha already has issues created to eventually addressed that bug.

PRP: https://snapshot.org/#/hbot-prp.eth/proposal/0x8e4e2a9e4830c493be065435f9bcbc6b5054c9e6f4f6ea67e63b01702fc3522a

### DIFF
--- a/gateway/jest.config.js
+++ b/gateway/jest.config.js
@@ -15,6 +15,7 @@ module.exports = {
     'conf/migration/migrations.js',
     'src/chains/solana/solana.ts',
     'src/connectors/uniswap/uniswap.config.ts',
+    'src/network/network.controllers.ts',
   ],
   modulePathIgnorePatterns: ['<rootDir>/dist/'],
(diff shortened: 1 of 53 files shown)


aarmoa:
@petioptrv please remember to always add the link to the PRP. Also would be helpful in this case to add the link the the PR in CoinAlpha repository.

PR #5486, opened 2022-07-11
Fix/paper trade fails on empty order book

MementoRC:

The function c_get_prices() returns s_decimal_NaN when the order book is empty. This causes the c_process_crossed_limit_orders_for_trading_pair() to fail with an uncaught error:
decimal.InvalidOperation: [<class 'decimal.InvalidOperation'>]
The proposal circumvent this by exiting the function early

The test test-cross_exchange_market_making.py test_empty_order_book actually reports this error (though the test passes?). Once implemented the error no longer occurs

**Tips for QA testing**:


MementoRC:
The mysteries of what is deemed valuable

PR #5500, opened 2022-07-14
(feat) Updated the Parrot

MementoRC:
**Before submitting this PR, please make sure**:

- [x] Your code builds clean without any errors or warnings
- [x] You are using approved title ("feat/", "fix/", "docs/", "refactor/")

**A description of the changes proposed in the pull request**:
The Parrot was outdated and on LM strategies return a 404 http error

**Tests performed by the developer**:
Added several tests to exercise several aspects of the module

**Tips for QA testing**:

Issue #5240, opened 2022-04-20
Liquidity Mining - Unexpected error while requesting data from Hummingbot API 


rapcmia:
**Describe the bug**
While testing liquidity mining strategy, running `status` command will return an error of `Unexpected error while requesting data from Hummingbot API`. The bot seems to create and cancel order accordingly, no stuck orders on exchange.

![image](https://user-images.githubusercontent.com/73840223/164202181-81f4c159-5c75-4c1f-87e1-3585a0ed8e04.png)

```
2022-04-20 09:53:13,114 - 17099 - hummingbot.connector.parrot - ERROR - Unexpected error while requesting data from Hummingbot API.
Traceback (most recent call last):
  File "/home/ubuntu/github/staging/hummingbot/connector/parrot.py", line 35, in get_campaign_summary
    campaigns = await get_active_campaigns(exchange, trading_pairs)
  File "/home/ubuntu/github/staging/hummingbot/connector/parrot.py", line 71, in get_active_campaigns
    resp_json = await resp.json()
  File "/home/ubuntu/miniconda3/envs/hummingbot/lib/python3.8/site-packages/aiohttp/client_reqrep.py", line 1097, in json
    raise ContentTypeError(
aiohttp.client_exceptions.ContentTypeError: 0, message='Attempt to decode JSON with unexpected mimetype: text/html', url=URL('https://papi.hummingbot.io/v1/mining_data/campaigns')
```

**Steps To Reproduce**
1. Setup simple liquidity mining strategy 
2. Start the bot
3. After creating the orders, run status command 

**Release version**
Version 1.2.0, Staging dev-1.3 and development branch

**Attachments**
[lm-test.zip](https://github.com/hummingbot/hummingbot/files/8520470/lm-test.zip)

drlcai:
Same problem here. I was running the bot in docker with version 1.2 and also from source for the staging version (1.3). Then I  tried status --live command. I got a flashing status report. It seems like it first showed me something then disappeared because of the bug. Then it repeated this error. 

The log file gives: 
2022-04-29 16:40:56,773 - 1 - hummingbot.connector.parrot - ERROR - Unexpected error while requesting data from Hummingbot API.
Traceback (most recent call last):
  File "/home/hummingbot/hummingbot/connector/parrot.py", line 35, in get_campaign_summary
    campaigns = await get_active_campaigns(exchange, trading_pairs)
[...]

PR #5501, opened 2022-07-15
(feat) Updating the parrot URL and some mechanics. Adding a few tests

MementoRC:

The Parrot was outdated and on LM strategies return a 404 http error

Added a few tests to exercise aspects of the module

**Tips for QA testing**:


MementoRC:
PRP referred to previous failed PR #5500. THis PR is the one correcting the testcase

cardosofede (approved):
LGTM

MementoRC:
Related to issue:https://github.com/hummingbot/hummingbot/issues/5240

rapcmia:
PR update:
- Setup simple Liquidity-mining strategy 
- Setup using pairs AVAX-USDT and FEAT-USDT from liquidity mining portal, ok 
- Setup using pair XRP-USDT and ETH-USDT for non liquidity mining
- Run status command and does not return the error of `Unexpected error while requesting data from Hummingbot API.`

However there are instances for ETH-USDT that its returning an error of `Failed to retrieve snapshot info`. 
image
```
2022-09-01 11:45:40,891 - 23503 - hummingbot.connector.parrot - WARNING - Failed to retrieve snapshot info. API response: {'status': 'error', 'message': 'Data not available for timestamp 1657747860000.'}
NoneType: None
```
Tried to crosscheck with other markets and version 1.7.0 but not getting the error above. 

Steps to reproduce:
1. Setup Liquidity_mining strategy
2. Setup using ETH-USDT market then start the strategy
3. Run status command and notice the error 

[lm-gateio.zip](https://github.com/hummingbot/hummingbot/files/9466803/lm-gateio.zip)

rapcmia:
@MementoRC good day to you, just checking if the issues reported can be fixed? It seems this PR was able to solved the issues with `Unexpected error while requesting data from Hummingbot API.` however there was a new issue found.

MementoRC:
@rapcmia Good day to you too. Is this not due to ETH-USDT not being a campaign, thus the PARROT not finding it on the miner.hummingbot.io? Sorry, I saw the LGTM and assumed it had been internally approved and did not look into it. I will do so today and provide feedback

MementoRC:
@rapcmia This is not an error, but a report in the parrot when a particular snapshot does not exist. I did not know how to best handle this case, it does not break the code so I did not think that emitting an exception would be good. See the snippet of code:
```
        campaigns = await get_active_campaigns(exchange, trading_pairs)
        tasks = [get_market_snapshots(m_id) for m_id in campaigns]
        snapshots = await safe_gather(*tasks, return_exceptions=True)
        for snapshot in snapshots:
            if isinstance(snapshot, Exception):
                raise snapshot
            if snapshot['status'] != "success":
                logger().warning(f"Failed to retrieve snapshot info. API response: {snapshot}", exc_info=True)
                continue

```

MementoRC:
Tests are failing after changing a string in the code. I do not know how to make them pass

rapcmia:
hi @MementoRC thanks for the update! Ill have a look with the team about it. For the failing GitHub CI, correct its not part of your PR so we'll have to rerun it again to fix it.

rapcmia (dismissed):
Test performed:
- Setup simple LM strategy for kucoin, binance, ascendex and gateio
- Tested with using miner market and non-miner pairs, ok 
- Run status command after the bot running for atleast 15-20mins, no errors and all ok 
- There are noticeable logs from parrot regarding non-miner pairs specially for Gateio
- Tested on source and docker build

JeremyKono:
@MementoRC can we please have your ETH address for payment? Thanks.

hummingbot/connector/parrot.py
```diff
@@ -38,8 +40,11 @@ async def get_campaign_summary(exchange: str, trading_pairs: List[str] = []) ->
         for snapshot in snapshots:
             if isinstance(snapshot, Exception):
                 raise snapshot
-            if snapshot["items"]:
-                snapshot = snapshot["items"][0]
+            if snapshot['status'] != "success":
+                logger().warning(f"Snapshot info for {trading_pairs} is not available verify that this is a valid campaign pair for this exchange")
+                continue
+            if snapshot["market_snapshot"]:
```
cardosofede, line 46: I will suggest to do this with get, because if not you can have an error here

hummingbot/connector/parrot.py
```diff
@@ -55,44 +60,65 @@ async def get_campaign_summary(exchange: str, trading_pairs: List[str] = []) ->
     return results
 
 
-async def get_market_snapshots(market_id: int):
+async def get_market_snapshots(market_id: int, timestamp: int = 1657747860000):
```
cardosofede, line 63: This is very bad, I saw the method the timestamp is hardcoded and you are just calling the method without any timestamp when this is used

hummingbot/connector/parrot.py
```diff
@@ -55,44 +60,65 @@ async def get_campaign_summary(exchange: str, trading_pairs: List[str] = []) ->
...
     async with aiohttp.ClientSession() as client:
-        url = f"{PARROT_MINER_BASE_URL}market_snapshots/{market_id}?aggregate=1m"
+        url = f"{PARROT_MINER_BASE_URL}user/single_snapshot?client_id=00000000000000000000000000000000000000000&market_id={market_id}&timestamp={timestamp}&aggregate_period=60"
```
cardosofede, line 65: why the client id is hardcoded an 00000000000000000000000000000000000000000
and also, why we are aggregating the period in 60 instead of 1m?

hummingbot/connector/parrot.py
```diff
@@ -55,44 +60,65 @@ async def get_campaign_summary(exchange: str, trading_pairs: List[str] = []) ->
...
         resp = await client.get(url)
         resp_json = await resp.json()
     return resp_json
 
 
-async def get_active_campaigns(exchange: str, trading_pairs: List[str] = []) -> Dict[int, CampaignSummary]:
+async def get_active_campaigns(exchange: str, trading_pairs=None) -> Dict[int, CampaignSummary]:
```
cardosofede, line 71: this change is the same, if you don't pass any value it will return an empty list, we should restore it

cardosofede (changes requested):

hummingbot/connector/parrot.py
MementoRC: This seemed to be the way to call a 'user' API method in a public way, otherwise you need to have an authenticated client_id
This seemed to be the way to pass 1m
Note that I do not have (or don't know where to find) the documentation for this API, so all these were gathered from looking at the commands with Brave -> 'Developer tool' when using the miner.hummingbot.io site with my account

hummingbot/connector/parrot.py
MementoRC: Agreed, I must have wanted to update this with time() and forgot after testing it

hummingbot/connector/parrot.py
MementoRC: This is possibly because PyCharm recommends to use the construct implemented rather than the 'None' in argument

hummingbot/connector/parrot.py
MementoRC: The reason for a default value:
```
{'reason': "Missing params in payload, all params {'market_id', 'timestamp'} "
           'must be present and valid.'}
```
when using '-1':
`{'message': 'Data not available for timestamp -1.', 'status': 'error'}`

MementoRC: Ok, finally figured it out, the valid timestamps can be gathered from the charts, then it can be used for the snapshots. Refactoring and adding the appropriate tests should be completed tomorrow

rapcmia (approved):
LGTM

hummingbot/connector/parrot.py
cardosofede: This is still harcoded, I think that you have to use the timestamp that it's provided in the first query that you make to retrieve the campaigns. There is one value that is last_snapshot_ts

cardosofede (changes requested):

MementoRC:
This is no longer in the PR

cardosofede (approved):

PR #5504, opened 2022-07-15
Feat/strategy tick

minuk302:

Before, Strategy's tick was fixed to 1 second
This pull request enables users to set the strategy's tick.

I set the strategy_update_tick config value to 0.1, 0.5, 3, and printed log for every time strategy's update is called.
I confirmed that the update is called every 0.1, 0.5, 3 second for each setting.

**Tips for QA testing**:


hummingbot/core/time_iterator.pxd
```diff
@@ -8,6 +8,7 @@ cdef class TimeIterator(PubSub):
     cdef:
         double _current_timestamp
         Clock _clock
+        double _overriding_tick
```
MementoRC, line 11: Would it not be possible to override the TimeIterator _clock tick_size instead?

minuk302: you meant **_clock.tick_size = _overriding_tick**?
**time_iterator**'s **_clock** is a clock that has the time_iterator as a **_child_iterators**. The _clock is not exclusive for the time_iterator.
So, **_clock.tick_size = _overriding_tick** will affect all other iterators's tick that are in clock's iterators.

MementoRC: Right, the whole point of it!, lol, silly me

JeremyKono:
Hi, can you please fix failing unit tests in this PR? Thanks.

minuk302:
@JeremyKono I am trying to fix it. But, my proposal did not get enough vote in snapshot. Can it still be merged??
https://snapshot.org/#/hbot-prp.eth/proposal/0x5703dd80e5d483a5a631e830eb6bdf116eaf8a8d65c2e7556aeb64ac49d76fe2

cardosofede:
@minuk302 I think that the implementation could be much more simpler and we shouldn't change the tick size once the clock is created. I create a video in Loom of how to make it configurable via an argument in the start command and also with a conf variable.

cardosofede:
I created this new PR that implements the changes, and also I added the test to have the coverage needed.
https://github.com/hummingbot/hummingbot/pull/5649

PR #5580, opened 2022-08-08
Suggest last docker instance as docker container to run.

meirm:
Save in hidden file the name of the last container launched with ./create.sh or ./start.sh
Next time that you run the ./start command, it will offer to start the previous container.


rapcmia:
Hi @meirm good day to you
We will close this pull request for now so you can resubmit using the template and add details for the for pull request.

PR #5598, opened 2022-08-16
(feat) Investigating Cython refactoring

MementoRC:
This is PR is not intended to be included into HB, but as a demonstration of alternative to cython .pyx

"Augmented Pure Python" mode of coding to migrate HB to Pure Python without loss of performance
   - Introducing Augmented Pure Python mode in PMM/inventory_skew with rudimentary performance tests
   - Augmented Pure Python is .py+.pxd. It allows the maintain uncompiled/debug Python code that is compiled with the updated setup.py
   - A few cases/examples for in-file cython framework with decorators rather than non-python (.pyx) syntax
   - One caveats is that cython-3.0.0a8 is needed for its bugfix of .pyd changes detection
   - The introduction of cython-3.0.0a8 highlights unicode -> str conversion issues in several exchange code that were worked-around in this commit

Ran several tests related to the Augmented Pure Python. Added a few profile tests in the test_pmm.py for that purpose
   - Tested that non-compiled .py+.pyd files are functional and in Pure Python, step-able line by line with debugger
   - Tested that the compiled/non-compiled with cython.py flag `is_compiled`
   - Tested that once compiled the performance is on-par with cython (.pyx) version - in fact faster! but the code is of course no-longer debuggable

**Tips for QA testing**:

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,5 @@
 [run]
+plugins = Cython.Coverage
 source_pkgs = hummingbot
 omit =
         hummingbot/core/gateway/*
--- a/hummingbot/connector/exchange/beaxy/beaxy_exchange.pyx
+++ b/hummingbot/connector/exchange/beaxy/beaxy_exchange.pyx
@@ -1012,7 +1012,7 @@ cdef class BeaxyExchange(ExchangeBase):
         try:
             assert path_url is not None or url is not None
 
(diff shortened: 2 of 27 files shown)


MementoRC:
This PR introduces too many changes in order to use cython 3.0.0a8, which only fix the non-detection of changes in the .pxd as trigger for compilation - Closing and resubmitting a cleaner PR

PR #5607, opened 2022-08-19
(feat) Evaluating Augmented Pure Python w/ PMM

MementoRC:
This is PR is not intended to be included into HB, but as a demonstration of alternative to cython .pyx

"Augmented Pure Python" mode of coding to migrate HB to Pure Python without loss of performance
   - Augmented Pure Python is .py+.pxd. It allows the maintain uncompiled/debug Python code that is compiled with the updated setup.py
   - Provides a new test class to illustrate how to cythonize in Pure Python
   - The current version of cython does not detect changes in .pxd as a trigger for compilation
     version 3.0.0a8 is required, however it also requires some modification for the implicit
     conversion of Unicode to str which is no longer accepted in a8+
   - setup.py is augmented with a detection of the .pxd files identifying the .py modules that
     need to be compiled

Ran several tests related to the Augmented Pure Python. Added a few profile tests in the test_pmm.py for that purpose
   - Tested that non-compiled .py+.pyd files are functional and in Pure Python, step-able line by line with debugger
   - Tested that the compiled/non-compiled with cython.py flag `is_compiled`
   - Tested that once compiled the performance is on-par with cython (.pyx) version - in fact faster! but the code is of course no-longer debuggable

**Tips for QA testing**:

### DIFF
--- a/hummingbot/strategy/data_types.py
+++ b/hummingbot/strategy/data_types.py
@@ -2,6 +2,8 @@ from dataclasses import dataclass
 from decimal import Decimal
 from typing import List, NamedTuple
 
+import cython
+
 from hummingbot.core.data_type.common import OrderType
 
 ORDER_PROPOSAL_ACTION_CREATE_ORDERS = 1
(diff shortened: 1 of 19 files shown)


MementoRC:
The error seems to be caused by cython-3.0a7, it cythonize/compiles correctly at my end with cython-3.0.0a10 (it also worked on the previous version where I forced the change in environment-linux.yml)

MementoRC:
The issue with coverage is that once compiled, the 'Augmented Pure Python' modules (.py + .pxd) are seen as .py with no test since in the current settings these modules are compiled. A solution is to use a different setting in setup.py for coverage to register the cythonize code.

MementoRC:
This version seems no longer mergeable, a new PR replaces it

PR #5626, opened 2022-08-26
(fix) Solve issue in Bybit Perpetual order fills update using HTTP

petioptrv:

This PR addresses an issue with the latest refactor of bybit perpetual. The code did not consider that response structure for order fills between linear and inverse products is different. The fix was originally introduced by @aarmoa in the CoinAlpha internal repository.

**Tips for QA testing**:

Please test that both inverse and linear connector versions process order fills correctly.


arnelhbot (approved):
Test performed:

- Cloned and compiled feature branch
- Connected Bybit_perp and Spot API
- Checked balance
- Created Pereptual_MM and Spot_Perp_Arbitrage strategy
- Confirmed Data integrity matches exchange
- Check status working accordingly
- linear & Inverse order fills in CSV Data file and history match exchange
- Linear & Inverse positions created and canceled on exchange


Note: _When Spot_Perp_Arbitrage is started, it checks if it can place an order and the perp connector for balance in the quote token. It should check against the collateral token instead. Will create a ticket for strategy_

hummingbot/connector/derivative/bybit_perpetual/bybit_perpetual_derivative.py
```diff
@@ -451,10 +451,13 @@ async def _all_trade_updates_for_order(self, order: InFlightOrder) -> List[Trade
...
             try:
                 all_fills_response = await self._request_order_fills(order=order)
-                fills_data = all_fills_response["result"]["trade_list"]
-                for fill_data in fills_data:
-                    trade_update = self._parse_trade_update(trade_msg=fill_data, tracked_order=order)
-                    trade_updates.append(trade_update)
+                trades_list_key = "data" if bybit_utils.is_linear_perpetual(order.trading_pair) else "trade_list"
+                fills_data = all_fills_response["result"].get(trades_list_key, [])
+
+                if fills_data is not None:
```
cardosofede, line 457: I think that this condition won't be reached because if you don't find the key, will return an empty list not a None value.

cardosofede (changes requested):

hummingbot/connector/derivative/bybit_perpetual/bybit_perpetual_derivative.py
petioptrv: I think this is to safe guard against the eventuality that `all_fills_response = {"result": None}`, in which case `fills_data` will be `None`

cardosofede (approved):

arnelhbot (approved):

PR #5683, opened 2022-09-09
Feat/binance perps cancel order

cardosofede:

The method _execute_cancel, was requesting the in_flight_orders + cached orders. I changed it to just check for the active orders and also the logic to stop tracking orders was changed based on the docs. (the code -2011 can have more than one error message so the two conditions needs to be present to stop tracking the order)

image

run a perpetual market making and check that the behavior is the expected

**Tips for QA testing**:
Test Perpetual Market Making for 2-3 days and check if the issue is solved

### DIFF
--- a/hummingbot/connector/derivative/binance_perpetual/binance_perpetual_derivative.py
+++ b/hummingbot/connector/derivative/binance_perpetual/binance_perpetual_derivative.py
@@ -1320,7 +1320,7 @@ class BinancePerpetualDerivative(ExchangeBase, PerpetualTrading):
         try:
             # Checks if order is not being tracked or order is waiting for created confirmation.
             # If so, ignores cancel request.
-            tracked_order: Optional[InFlightOrder] = self._client_order_tracker.fetch_order(client_order_id)
+            tracked_order: Optional[InFlightOrder] = self._client_order_tracker.fetch_tracked_order(client_order_id)
             if not tracked_order or tracked_order.is_pending_create:
                 return
 
--- a/hummingbot/connector/derivative/binance_perpetual/binance_perpetual_derivative.py
+++ b/hummingbot/connector/derivative/binance_perpetual/binance_perpetual_derivative.py
@@ -166,7 +166,7 @@ class BinancePerpetualDerivative(ExchangeBase, PerpetualTrading):
 
     @property
     def limit_orders(self) -> List[LimitOrder]:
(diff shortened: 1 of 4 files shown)


raymondcarl:
Only in my opinion:

The issue on my side seems to be fixed by making a 2 line change in how orders not found on the exchange are handled and processes them the same ways as the Bitmex connector. The PR above handles it in a more professional way by implementing the process_order_not_found method that is already available in the order tracker class. However, there are a lot of other changes in this PR that cleanup the whole process. In my opinion, we should make the smallest change possible for this specific issue.

Then, the improvements in the PR should actually be implemented in an improvement proposal that moves all of the common processes into the common derivative connector class so that it works the same for all of the exchanges except for the portions that are unique to the exchange. For example, there is a lot of commonality between the Binance and Bitmex in terms of the cancellation function, but they are very different in how they handle the cancellation processing (beyond the unique parts of each exchange in terms of message structure, etc.). It would be much easier to work with in the future to move as much functionality as possible to a common derivative connector. The existing derivative base class doesn't do much today.

I have not analyzed the commonality between each derivative connector, but there does seem to be some from looking at the order cancellation process. The issue this PR fixes could have been caught in unit testing, but it's not easy to build consistent unit tests across all 5 connectors. So, I think the improvements above would be better suited towards a refactoring of the whole derivative component.

cardosofede:
@raymondcarl hi Raymond and thank you for your comments.
The solution is to migrate this connector to be a subclass of ExchangePybase that has all this logic implemented instead of being a subclass of Exchange base.
Secondly when you review the failure that you are facing, in this case is that the connector is not notifying the strategy that there is a change in the code, is normal to check in other places where do you have this problem.
- In this case I found issues when creating order (it was raising exceptions that no one will catch), adding the logic to process the order not found (this will ask 3 times until finally delete the order), process the trade cancellation updates when you receive the response (this is better because based on the logs that I have the problem was that the ws message of the cancel wasn't arriving, so if you use the rest message is better and the logic of the ws message will be omitted) and improve the logic of the order updates just to make it only to the tracked orders

arnelhbot (approved):
Test performed:

- Cloned & compiled feature branch
- Connected Binance_perp and checked balance
- Created simple perpetual_mm
- Integrity confirmed synced with exchange
- Order files from history, CSV Data match exchange
- Confirmed cancel order was not observed
- Initial and long term test successful with no issues

Faouzijedidi1 (approved):

Issue #5667, opened 2022-09-06
HitBTC - Unexpected error occurred when listening to order streams 


rapcmia:
**Describe the bug**
When using HitBTC as connector for pure market-making and avellaneda, we noticed that on the log panel it kept on reporting `Unexpected error occurred when listening to order book streams. Retrying in 5 seconds...` 

![image](https://user-images.githubusercontent.com/73840223/188573042-d9050644-22b7-413e-bf55-46636b80de32.png)

```
2022-09-05 08:48:43,492 - 713594 - hummingbot.connector.exchange.hitbtc.hitbtc_api_order_book_data_source - ERROR - Unexpected error occurred when listening to order book streams. Retrying in 5 seconds...
Traceback (most recent call last):
  File "/home/ubuntu/hitbtc/hummingbot/core/data_type/order_book_tracker_data_source.py", line 80, in listen_for_subscriptions
    ws: WSAssistant = await self._connected_websocket_assistant()
  File "/home/ubuntu/hitbtc/hummingbot/core/data_type/order_book_tracker_data_source.py", line 200, in _connected_websocket_assistant
    raise NotImplementedError
NotImplementedError
```
Tried to compare the orderbooks from client and exchange and seems to ok

**Steps To Reproduce**
1. Setup a hummingbot strategy using HitBTC
2. Start the strategy and notice the log panel for the error

**Release version**
Development 1.8

**Attachments**
[logs_ave-hitbtc.zip](https://github.com/hummingbot/hummingbot/files/9494337/logs_ave-hitbtc.zip)

rapcmia:
Severity: P2
Bounty: 50000 (https://hummingbot.org/maintenance/bugs/)
Rationale: Bug related to certified exchange

JeremyKono:
Assigning issue to @MementoRC. Thank you for your contribution!

JeremyKono:
Hi @MementoRC can we have your ETH wallet address to proceed for payment? Thanks!

arnelhbot:
Closing Issue:
Tested on latest Dev1.8.0 and has been fixed. The fix will soon be merged to the next version of Master

PR #5687, opened 2022-09-09
(fix) HitBTC order book failure

MementoRC:

Issue: https://github.com/hummingbot/hummingbot/issues/5667
This is a very basic fix, the issue is cause by the modification of the base class that did not consider inheriting classes not yet implementing the latest methods. The solution is to simply pass the failing method leaving the specific trade and order-book listening is done by specific methods already existing. Similar issue as for coinbase_pro, Kraken, etc

This being a certified exchange, the real fix should be a refactoring, which is likely to be scheduled soon

HitBTC is not accessible to the op

**Tips for QA testing**:

### DIFF
--- a/hummingbot/connector/exchange/hitbtc/hitbtc_api_order_book_data_source.py
+++ b/hummingbot/connector/exchange/hitbtc/hitbtc_api_order_book_data_source.py
@@ -1,25 +1,21 @@
 import asyncio
 import logging
 import time
-
 from decimal import Decimal
 from typing import Any, Dict, List, Optional
 
 import aiohttp
 import pandas as pd
+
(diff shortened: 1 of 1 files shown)


arnelhbot (approved):
Test performed:

- Clone & compile feature branch
- Start bot and create password
- Connect HitBTC API keys and check balance
- Created Pure_MM and Avellaneda_MM strategy
- Order filled and cancel confirmed
- Checked status/-live
- Ran both strategies for 2hours
- Confirmed no order stream error occurred

cardosofede (approved):
approved to let it work, but we need to refactor it!

MementoRC:
Possibly related to:https://github.com/hummingbot/hummingbot/issues/5705

Issue #4962, opened 2021-12-23
Exception thrown in liquidity mining strategy when no orders are available on either side of the order book


klpanagi:
**Describe the bug**
An exception is thrown in liquidity mining strategy while trying to create budget allocation calculate volatility (`update_volatility`), when either side of the order book is empty.
The error is raised within the `update_volatility()` and `create_budget_allocation()` methods of the `LiquidityMiningStrategy` class. You can cross-check this from the attached log file.

Furthermore, I did try with the `pure_market_making` strategy, where the error is also reported in logs, though an exception is not raised. 

**Steps To Reproduce**
Here are the steps to reproduce the issue (see attachments in the section below):
This can be reproduced on any market. The only dependency is that at least one side of the order book (bid/ask) has to be empty
...

**Screenshots**
![Screenshot from 2021-12-23 17-37-06](https://user-images.githubusercontent.com/4770702/147280257-325c9613-4a3b-48c4-a912-6ea05137819d.png)
![Screenshot from 2021-12-23 19-34-56](https://user-images.githubusercontent.com/4770702/147280261-bdc926b7-8347-44ed-b05d-055d6e14a797.png)

**Release version**
0.4.6

**Attachments**
Log file attached.
[hummingbot.log](https://github.com/hummingbot/hummingbot/files/7770956/hummingbot.log)

**Discord: klpanagi#2483**

rapcmia:
Severity: P2
Bounty: 50000 (https://hummingbot.org/maintenance/bugs/)
Rationale: Bug related to a component in Liquidity mining strategy

MementoRC:
This is reproducible with adding a case to the testsuite. This should not be too difficult to correct

JeremyKono:
Assigning this issue to @MementoRC. Thank you for your contribution!

rapcmia:
The fix has been added to the upcoming version 1.9.0 release.

PR #5711, opened 2022-09-16
(fix) LM raises exceptions on empty order books

MementoRC:

Resolution proposal for:[Exception thrown in liquidity mining strategy when no orders are available on either side of the order book#4962](https://github.com/hummingbot/hummingbot/issues/4962)
The proposal fixes this issue by detecting the markets with empty order books and pulling them in a holding dictionary. This is validated at every tick, in case the empty order book is due to a glitch

The deprecated `estimate_fee` was also replaced with `build_trade_fee()`

More test could be added, since this proposal could not be the preferred way to fix this issue, the PR is proposed in this state

Added a test that initializes a market with an empty order book and verified the failure.
Updated the strategy to pass the test
Added a test to verify that only the empty order book markets are filtered out, non-empty order book markets are used and budgets are created for them

**Tips for QA testing**:


hummingbot/strategy/liquidity_mining/liquidity_mining.py
```diff
@@ -275,6 +280,24 @@ def start(self, clock: Clock, timestamp: float):
...
+        Verify that the active markets have non-empty order books. Put the trading on hold
+        for pairs with empty order books, in case it is a glitch?
+        """
+        markets = self._market_infos.copy()
+        markets.update(self._empty_ob_market_infos)
+        for market, market_info in markets.items():
+            if markets[market].get_price(False).is_nan():
+                self._empty_ob_market_infos[market] = market_info
+                if market in self._market_infos:
+                    self._market_infos.pop(market)
```
cardosofede, line 294: Just one simple change, we should log that this market is going to be removed or which markets are unable

cardosofede (changes requested):

hummingbot/strategy/liquidity_mining/liquidity_mining.py
MementoRC: @cardosofede Good point, I log it as a warning (rather than info) since the user may want to stop and redefine the confi rather than rely on the bot mechanic

arnelhbot (approved):
Test performed:

- Clone and compiled feature branch
- Connected API and checked balance
- Created simple Liquidity_minning Strategy
- Selected market with an empty order book
- Confirmed trading has been paused
- Observed Consider redefining your strategy

cardosofede (approved):
LGTM

Issue #5700, opened 2022-09-13
FTX - Websocket disconnection caused unexpected error while listening to user and orderbook stream


arnelhbot:
**Describe the bug**
When reviewing our testbot for `FTX` with a simple `Pure_MM` strategy, we have noticed instances of the `Unexpected error while listening to user and order book stream ` occurring while the bot was running. 

Note: Issue doesn't seem to impact the bot in creating orders when websocket gets disconnected, the bot will cancel created orders when disconnected as expected. No stuck orders on exchange.

User Streem
```
2022-09-12 23:20:11,412 - 968357 - hummingbot.connector.exchange.ftx.ftx_api_user_stream_data_source.FtxAPIUserStreamDataSource - ERROR - Unexpected error while listening to user stream. Retrying after 5 seconds...
Traceback (most recent call last):
  File "/home/ubuntu/ftx/hummingbot/core/data_type/user_stream_tracker_data_source.py", line 48, in listen_for_user_stream
    await self._process_websocket_messages(websocket_assistant=self._ws_assistant, queue=output)
  File "/home/ubuntu/ftx/hummingbot/connector/exchange/ftx/ftx_api_user_stream_data_source.py", line 84, in _process_websocket_messages
    await asyncio.wait_for(super()._process_websocket_messages(
  File "/home/ubuntu/miniconda3/envs/hummingbot/lib/python3.8/asyncio/tasks.py", line 483, in wait_for
    return fut.result()
  File "/home/ubuntu/ftx/hummingbot/core/data_type/user_stream_tracker_data_source.py", line 75, in _process_websocket_messages
    async for ws_response in websocket_assistant.iter_messages():
  File "/home/ubuntu/ftx/hummingbot/core/web_assistant/ws_assistant.py", line 69, in iter_messages
    response = await self._connection.receive()
  File "/home/ubuntu/ftx/hummingbot/core/web_assistant/connections/ws_connection.py", line 62, in receive
    msg = await self._process_message(msg)
  File "/home/ubuntu/ftx/hummingbot/core/web_assistant/connections/ws_connection.py", line 84, in _process_message
    msg = await self._check_msg_types(msg)
  File "/home/ubuntu/ftx/hummingbot/core/web_assistant/connections/ws_connection.py", line 89, in _check_msg_types
    msg = await self._check_msg_closed_type(msg)
[...]

PR #5723, opened 2022-09-20
(fix) Websocket disconnects with 1006 and 1001 errors

MementoRC:

Proposal to address websocket failures with 1000, 1001 and 1006 error codes. Related to issues: 
[FTX - Websocket disconnection caused unexpected error while listening to user and orderbook stream#5700](https://github.com/hummingbot/hummingbot/issues/5700)
Ascendex - Getting "unexpected error occurred when listening to orderbook streams" when the bot is stopped or killswitched for a long time#5602

The issue seems to be caused by several different root cause, but ending up closing the websocket connection either when HB expects it (1000 would be normal from the websocket with msg reporting the close status) or unexpectedly (1006, HB is expecting the connection to be open and its internal _connected flag to be True when the websocket is actually closed)

The proposal here is to simply add a auto-reconnect attempt when the websocket tells us that it is closed when we did not initiate the close, or when unexpectedly the websocket is closed. This is achieved either by catching a non-responsive request and attempting a ping, or by not raising an exception, but logging the disconnect, closing the websocket, updating the status of the internal flag and attempting the reconnect

Added a test to exercise the disconnect-reconnect.
Ran the WSconnector testsuite
* Did not test a live bot since no access to neither these exchanges

**Tips for QA testing**:


cardosofede (changes requested):
This is not the way to solve this "issue".
First, this class shouldn't be trying to reconnect by itself. Why we want this?
Secondly, you are using methods like, await self.disconnect and then assigning self._connected to False when it's already doing that.
I think that the current design is good, and we shouldn't modify it, unless we have an error in the client that it's causing this error.
In this case, if we have an issue with the web socket, the connector will try to reconnect automatically.
image

MementoRC:
@cardosofede Are we saying that this is not an issue? The exception occurs when the WSconnector thinks it is connected, but the msg receive is 'CLOSED' or 'CLOSING', is the solution to remove the error message then?
This is a fairly common method to establish a websocket, send the subscription and conituously receive the msgs with an async generator inner_loop, in fact it is in the loopring websocket

PR #5738, opened 2022-09-26
(feat) Initial VWAP Market Making script for HBot course.

ajjcoppola:
VWAP Market Maker - Script Developer I Couse - Cohort 1

Enhancement of VWAP strategy that sells or buys a fixed amount of base or quote, with minimal slippage, to a market making strategy that sets buy/sell limit orders on both sides of the book, using the VWAP calculations.
One buy, one sell limit order is placed each time step, and if all goes well, after a while, both sides should enentually get filled.
When that doesn't happen, some type adverse selection happened, and you are left with unresolved inventory excess.

This strategy can be used as a testing ground for developing a library of rewards factors, to be used in reinforcement learning, by having the VWAP size correspond to an episode in the reinforcement learning sense.  Because it is simple, and coded as a script, debugging the various problems in a debugger is easy.

Tests were done by running various parameters over multiple hours and days.

See the Notion Stregy Design document for more details, and screenshots of results.

### DIFF
--- a/scripts/vwap_rewards.py
+++ b/scripts/vwap_rewards.py
@@ -1,0 +1,230 @@
+import logging
+import math
+from decimal import Decimal
+from typing import Dict
+
+from hummingbot.connector.utils import split_hb_trading_pair
+from hummingbot.core.data_type.order_candidate import OrderCandidate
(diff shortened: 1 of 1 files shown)


cardosofede:
Added branch to Notion Design Template

Issue #5657, opened 2022-09-02
GatewayV2 - Failed to connect uniswapLP


arnelhbot:
**Describe the bug**
Upon testing `GatewayV2` through Development 1.7.0, we have noticed when connecting `UniswapLP` the client will give an `Unhandled error` in the log pane. `TypeError: string indices must be integers`
```
Traceback (most recent call last):
  File "/home/ubuntu/Hummingbot/pull_request/development/hummingbot/core/utils/async_utils.py", line 9, in safe_wrapper
    return await c
  File "/home/ubuntu/Hummingbot/pull_request/development/hummingbot/client/command/gateway_command.py", line 361, in _gateway_connect
    chains: List[str] = [d['chain'] for d in available_networks]
  File "/home/ubuntu/Hummingbot/pull_request/development/hummingbot/client/command/gateway_command.py", line 361, in <listcomp>
    chains: List[str] = [d['chain'] for d in available_networks]
TypeError: string indices must be integers

```
**Steps To Reproduce**
Here are the steps to reproduce the issue (see attachments in the section below):
1. Gateway_Generate-cert
2. Gateway_Create
3. Gateway_connect uniswapLP
4. Observe log pane for error
...

**Screenshots**
![image](https://user-images.githubusercontent.com/99257257/188099734-1b09b788-b5ec-419b-b594-6b8dc7f3927d.png)

**Release version**
Dev1.7.0

**Attachments**
[UniswapLP_conf_logs.zip](https://github.com/hummingbot/hummingbot/files/9476402/UniswapLP_conf_logs.zip)

vic-en:
@arnelhbot 
Delete gateway_connections.yml file in the humminbot/conf dir. Let me know if the issue persists.

nikspz:
@vic-en The issue persists on the fresh installation using development branch also.

tried to remove gateway_connections off the humminbot_client/conf dir, issue still persists
![Failed to connect uniswapLP](https://user-images.githubusercontent.com/83953535/189121092-be05426b-0928-4cc4-a7a9-5e4154e798ae.gif)

arnelhbot:
@vic-en apologies for my delayed response. But I am also getting the same results as Nikspz with the suggested solution.

rapcmia:
Severity: P2
Bounty: 50000 (https://hummingbot.org/maintenance/bugs/)
Rationale: Bug related to a component in gateway

MementoRC:
[...]

PR #5741, opened 2022-09-26
(fix) Gateway fails to connect to Uniswap LP

MementoRC:

Resolving:[GatewayV2 - Failed to connect uniswapLP#5657](https://github.com/hummingbot/hummingbot/issues/5657)
The issue seems to be that the connector re-uses the Uniswap.config object that is already referenced by Uniswap. When requested the response is a string documenting the reference rather than the object content. The fix proposed is to make a copy, which could be done with the structuredClone() available in Node_17.x (we use Node_14.x ... why?)

Compiled and restarted the gateway
Used the culr manual-test to verify that the `GET /connectors` is as expected
In HB, executed `gateway connect UniswapLP` and received the normal prompt asking for the network

**Tips for QA testing**:


arnelhbot (dismissed):
Test performed:

- Closed and compiled feature branch
- Manually created docker gateway
- Manually created docker+docker gateway
- Gateway generate-cert and create successfully
- Connected uniswapLP_ethereum and polygon with no issues
- Tested on Source AWS and Docker Build

MementoRC:
This is kind of a new bug, EVM_LP was not recognized as a valid connector type in settings.py

vic-en (approved):
Changes LGTM

rapcmia:
PR update: Checked today and the issue we observed yesterday does not occur anymore. 
- Clone and install successfully
- Create docker image for both hummingbot client and gateway-v2 locally for tests
- Gateway create, pulled docker image successfully and observed "online"
- Gateway connect uniswapLP (for ethereum and polygon), no errors on logpane
- Gateway conenct uniswap (for ethereum and polygon), all ok 
- Run simple AMM-arb and UniswapLP using arbitrum, all ok
- Tested on source and docker build

rapcmia (approved):

Faouzijedidi1 (approved):
Approving this as a temporary fix which should be updated as the list of available networks grows. JSON.parse(JSON.stringify()) is a bad practice that can cause delays and loss of data in some cases. In this case, it solves the problem and the issues do not apply, hence, the approval.

MementoRC:
@Faouzijedidi1 The structuredClone() method is only available in a more recent version of Node, should we rather update the Node_14.x to Node_18.x?
I am not versed in typeScript and did not find a better way of fixing this issue. I would prefer implementing something more acceptable, in fact I was very tempted to update Node, but thought it would not be approved

PR #5744, opened 2022-09-27
(feat) add autoPMM script (botcamp)

mrhouzlane:

This script calculates spread and order amount and execute trades. 

**Tips for QA testing**:


scripts/autoPMM.py
```diff
@@ -0,0 +1,109 @@
...
+        return order_amount 
+
+            
+    def calculate_spread(self):
+        
+        mid_price = self.connectors[self.exchange].get_mid_price(self.trading_pair)
+        bid_price = self.connectors[self.exchange].get_price(self.trading_pair, False)
+        ask_price = self.connectors[self.exchange].get_price(self.trading_pair, True)
+
+        bid_ask_spread = (bid_price - ask_price )* mid_price / 100
```
MHHukiewitz, line 48: You take the absolute spread (bid - ask) and multiply it with 1% of mid-price? This will yield very different results, depending on mid-price alone. Higher mid-price = higher spread, which is incorrect.
Also, by subtracting the higher ask from the lower bid, will always result in a negative value for `bid_ask_spread`.

MHHukiewitz: Example with two markets, each having 1% bid and ask spread:
b(id) = 0.99, a(sk) = 1.01, (mid-)p(rice) = 1.00 => `-0.02 * 1/100 = -0.0002` => _0.02 % calculated spread_
b = 990, a = 1010, p = 1000 => `-20 * 1000/100 = -200` => _20 % calculated spread_

MHHukiewitz (changes requested):
You should review the `calculate_spread()` method; depending on the markets you trade, you'll receive (unintentionally) wildly differing spreads.

scripts/autoPMM.py
mrhouzlane:
```
 def calculate_spread(self):
        
        mid_price = self.connectors[self.exchange].get_mid_price(self.trading_pair)
        bid_price = self.connectors[self.exchange].get_price(self.trading_pair, is_buy = False)
        ask_price = self.connectors[self.exchange].get_price(self.trading_pair, is_buy = True)

        bid_ask_spread = (ask_price - bid_price)
        self.logger().info(f"best bid: {bid_price}")
        self.logger().info(f"best ask: {ask_price}")
        self.logger().info(f"bid_ask_spread in % : {bid_ask_spread * Decimal(100)} %")

        return bid_ask_spread

```

This should be fine.

PR #5781, opened 2022-10-06
(fix) fix incorrect quote volume quantize method while quiering the order book

supervik:

The current quote volume quantization uses `c_quantize_order_amount method`. But it can be used only to quantize base asset volume not quote.

The bug is visible and gives incorrect amounts when prices of assets differes like in BTC pairs. 
For example if we take XRP-BTC pair and try to get BTC volume for XRP amount with `c_get_quote_volume_for_base_amount` the wrong quantization method will quantize BTC volume using XRP trading rules like min/max order size that differs a lot from BTC trading rules. 

 Next 3 methods are incorrect and should be fixed:

- `c_get_price_for_quote_volume`
- `c_get_quote_volume_for_base_amount`
- `c_get_quote_volume_for_price`

To see the mistake you can run the script below. The quote volume that you got in `get_quote_volume_for_base_amount `will be 0 even for 10 000 XRP on binance market. 10 000 XRP gives approxiamtely 0.244 BTC and then this amount is quantized on XRP-BTC market and because trading rules min order size is greater than 0.244 XRP the result volime is 0:

```py
from hummingbot.strategy.script_strategy_base import ScriptStrategyBase

class LogPricesExample(ScriptStrategyBase):
    markets = {"binance": {"XRP-BTC"}}

    def on_tick(self):
        volume_1 = self.connectors["binance"].get_quote_volume_for_base_amount('XRP-BTC', True, 1000).result_volume
        volume_2 = self.connectors["binance"].get_price_for_quote_volume('XRP-BTC', True, 0.1).result_volume
        volume_3 = self.connectors["binance"].get_quote_volume_for_price('XRP-BTC', True, 0.00025).result_volume
        
        self.logger().info(f"{volume_1}, {volume_2}, {volume_3}")
```

Run several scripts testing those methods with different values and different pairs and connectors.

**Tips for QA testing**:
`c_get_quote_volume_for_price` has another bug. It uses the wrong order_book method `c_get_volume_for_price` instead of `c_get_quote_volume_for_price`
 I think I need another PR for that or should I add 1 more commit here?


cardosofede:
I think that the bug is real, but I don't know if we should quantize the results... probably we just need to return the values as they are, what do you think?
this can be applied for all the queries

supervik:
> I think that the bug is real, but I don't know if we should quantize the results... probably we just need to return the values as they are, what do you think? this can be applied for all the queries

Yes, I also thought about that first. But then test it and saw the result :grimacing: :
0.02552000000000000101696429055664339102804660797119140625

nikspz (approved):
Test performed:

 - Cloned and installed feature branch
 - Started bot with no errors
 - Manually created docker image successfully
 - Started the scipt provided and reviewed issue with 0E8 was fixed
 - Tested Certified connectors're not affected/work as expected
 - Tested on AWS Source and Docker

PR #5836, opened 2022-10-27
fix/zero asset balance

ethicrypt:

**A description of the changes proposed in the pull request**: When modifying the inventory asset ratio configuration, Hummingbot was always showing a base asset balance of 0 even when the balance command showed otherwise. This is because only the quote asset was having its balance checked, and this fix includes the base asset in the balance check which resolves the issue.

**Tests performed by the developer**: I created several different strategies and tried modifying the inventory asset ratio with inventory skew enabled as well as disabled. In every scenario, the correct results were returned with the fix applied. I verified this by comparing the values returned with the return from the "balance" command.

**Tips for QA testing**: It would appear this bug only appears in pure market making and Avellaneda strategies. It may also require inventory skew being enabled.


ethicrypt:
Closing this pull request in order to redo it properly.

advaitrane98:
@ethicrypt is there any ETA for the redo? We would love to try out this functionality and hence would be happy to support in form of token incentives.

ethicrypt:
> @ethicrypt is there any ETA for the redo? We would love to try out this functionality and hence would be happy to support in form of token incentives.

I'm aiming to get it submitted today.

PR #5840, opened 2022-10-27
Fix/gateway testing methods

MementoRC:

There is a few very subtle (at least for me) sub-optimal constructs in the MockWebSocketServer module:
   - The aiohttp.ClientSession is created outside an async - still works but will cause issues with closing the session
   - In fact there seems to be no need for a session in the server factory (It seems that is it just an artifact to get the correct ws_connect address)
   - Creating the ws_connect address at this stage, effectively requires to have a global EventLoop, but this does not seem necessary
   - The server stop is creating also attempting to benefit from the global ev_loop, even though it has its own new_event_loop
 
In this proposal:
   - Simplify the factory to re-use its even loop (it runs in its own thread, with its own event loop called with new_event_loop)
   - Remove the session, which allows to not have the need for a global session, and thus global ev_loop
   - Correct the testcase to create the session in async with each test (sessions are opened and closed for each tests)
   - Record the ws_connect before starting the patcher

Pass the tests - Note that this will likely break other tests, which ideally should not be the case, agreed?

**Tips for QA testing**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@
 .PHONY: development-diff-cover
 
 test:
-	coverage run -m nose --exclude-dir="test/connector" --exclude-dir="test/debug" --exclude-dir="test/mock" --exclude-dir="test/hummingbot/connector/gateway"
+	coverage run -m nose --exclude-dir="test/connector" --exclude-dir="test/debug" --exclude-dir="test/mock"
 
 run_coverage: test
 	coverage report
--- a/hummingbot/core/mock_api/mock_web_socket_server.py
+++ b/hummingbot/core/mock_api/mock_web_socket_server.py
@@ -1,11 +1,10 @@
 import asyncio
+import errno
+import socket
 from threading import Event, Thread
 from typing import Optional
-import socket
-import errno
 from urllib.parse import urlparse
 
-import aiohttp
 from aiohttp import web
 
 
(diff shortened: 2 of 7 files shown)


MementoRC:
@JeremyKono @rapcmia How do we create a new branch to which I could submit several complementing PRs to follow this one?

JeremyKono:
Can you elaborate how to create a branch? A new branch that you could push to create a new PR? @MementoRC

MementoRC:
@JeremyKono I would like to create step-by-step changes to a branch on which I would create PRs instead of creating one PR against `development` that would have tons of changes and may be declined
Like, I could create PRs against `staging` or `master` for example
Note that I don't know if that makes sense, I am just wondering how to make proposals that have large modification requests

mifeng:
> @JeremyKono I would like to create step-by-step changes to a branch on which I would create PRs instead of creating one PR against `development` that would have tons of changes and may be declined Like, I could create PRs against `staging` or `master` for example Note that I don't know if that makes sense, I am just wondering how to make proposals that have large modification requests

I think a feature branch is the right approach here, i.e. create a fork and keeping adding commits to it, and keep it synced with `development`. Since `staging` and `master` are based on `development`, devs should not base branches on them

rapcmia:
Gateway has been refactored and moved to a new repository, https://github.com/hummingbot/gateway

PR #5957, opened 2022-12-17
Botcamp/flexible savings

discosultan:
**NB! This is a botcamp submission and is not intended to be merged.**

Links:
- [Strategy design](https://hummingbot-foundation.notion.site/Flexible-Savings-fac085b9db694fa7bd1329f649cd1553)
- [Investopedia article on four-week rule](https://www.investopedia.com/articles/technical/02/052102.asp)

Add a flexible savings script. It implements a directional strategy. However, the purpose of this script is not the strategy itself but the usage of a *custodian*. A custodian is a component that is responsible for holding your assets while they are not in active use, such as required for an open order. This is useful mainly when dealing with low-frequency strategies that operate on daily candles, for example.

An example of a custodian can be Binance's savings custodian. When trading BTC-USDT, instead of holding your assets in a spot account, it may decide to hold them in a flexible savings account instead. A flexible savings account does not lock up the assets so that they can be redeemed for trading immediately at any point in time. This allows, in addition to the profits from the strategy itself, to also earn daily interest on the assets.

Another example could be a Cardano custodian. The staking on the Cardano blockchain is unique in the sense that to earn staking rewards, you only need to have ADA in your wallet during the beginning of an epoch for the staking snapshot. This means that every 5 days when the snapshot is about to happen, the custodian could withdraw your ADA, lock in the staking rewards and deposit the ADA back to the exchange for trading.

The strategy implemented here is called the four-week rule strategy. In its original form, it looks at the closing prices of daily candles. If the current price is higher than the maximum of the past 4 weeks, it opens a long position. If the current price is lower than the minimum of the past 4 weeks, it opens a short position. This makes it a "breakout" strategy that tries to catch big price moves. Works well in a trending market. Not so well in a side-ways market. Additionally, a simple moving average is used to exit positions early and lock in profits and minimizelosses. The moving average is typically half the duration, which is 2 weeks long. See the link above for more info.


cardosofede (commented):
Congratulations @discosultan, is an awesome example!
I only reviewed the code of the script, but these days I will review the implementation of juno.
There are only 2 things that you can review here:
- Can you create the Juno library and make it compatible to install in the current conda environment?
- In the on_tick method you are allowing the execution of the setup, the idea of that to use the connectors when they are ready, but then all the logic will be handled with the new candles, right?

discosultan:
@cardosofede Thank you for the review!

> Can you create the Juno library and make it compatible to install in the current conda environment?
I can get that done in the coming week. That would clean up the PR quite a bit and make it mergeable if it's going to be just the example script.

> In the on_tick method you are allowing the execution of the setup, the idea of that to use the connectors when they are ready, but then all the logic will be handled with the new candles, right?
Correct.

discosultan:
@cardosofede I've extracted the juno library. Note that I renamed it to [barbthroat](https://github.com/discosultan/barbthroat) to better fit into the 🐦 family.

PR #5958, opened 2022-12-18
Fix/polygon amm

FlorentG74:

- [X ] Your code builds clean without any errors or warnings
- [ X] You are using approved title ("feat/", "fix/", "docs/", "refactor/")

This change fixed the 'Waiting for allowances" error message that block the launching of strategies on DEXs on the MATIC network when the base token (MATIC) is part of the trading pair. The reason is that the contracts don't provide explicit allowances on MATIC ()

1/Creation of a cross-exchange arbitrage strategy between Binance and Uniswap/Quickswap and launching of the strategy. Once fix is applied the "waiting for allowances" message only gets displayed until allowances are confirmed on the other token of the pair and then trading starts.
2/Addition of gateway manual tests to verify that behaviour with regards to allowances is consistent post fix.
3/Addition of gateway manual tests to verify that prices can be retrieved from uniswap and quickswap

**Tips for QA testing**:


rapcmia:
Hi @FlorentG74 good day to you
Our team was able to check this PR earlier and noticed that there was a change on the `gateway_evm_amm.py` which would affect most of our gateway connectors and not ideal change for a fix for Polygon connector. We would suggest to try the updated development branch's gateway where users can now modify the json file to add/delete to the token list.

rapcmia:
Hi @FlorentG74 please be advised that the foundation has decided to cancel the bounty for the issue ticket. The latest gateway update enables users to update the tokenlist url and this way to mitigate the issue for waiting allowances and token approval. We appreciate your time and close this PR for now, thank you

PR #5967, opened 2022-12-21
Add steamroll script

TestDataGuy:

Notion Design Template: https://hummingbot-foundation.notion.site/Steamrolling-e0e59f5bb243401188eb61ae04318e8f
This script attempts to add liquidity to exchanges with lower volumes using an exchange with more volume as a proxy for price.

**Tips for QA testing**:


supervik (commented):
The code does what is intended, but there are some problems:

1. I see that the logic with flag variables `self.buy_order_placed`, `self.sell_order_placed`, `self.long_position`, `self.short_position` has issues - they are not updated properly. That's why orders will be placed when they shouldn't as well as canceled. 
For example when an unfilled order is canceled the variable `self.buy_order_placed`, `self.sell_order_placed` are set to **True** but there is no check whether the "buy/sell" order or the "long/short" position was canceled. The same is for `did_fill_order` method that have `self.long_position`/`self.short_position` updated to **True** but there is no check what kind of order was filled. Whether it was a "position" or an "order".
**How to fix:** Test the logic with flags variables and fix the conditions where neccessary.

2. When the strategy is running it is difficult to say what is going on - it waits the perfect spreads moment to trigger limit orders and there is no output to show that. I would suggest to add the relevant information to `format_status`. Like current spreads between liquid and less_liquid exchange and current active orders. 
**How to fix:** Add more output to display with the status command

I see that this strategy could be fixed easily to make it work as intended. Other than that I think that it is best suited for perpetuals where it is easier to define the logic with opening and closing positions.

cardosofede:
Added branch to Notion design template

PR #6102, opened 2023-02-20
add start docker in hummingbot

ejambulo:
Added this command in README.md.

## Start Docker in Hummingbot
The `sudo service start docker` script will enable to start docker of Hummingbot.

```
sudo service start docker


fengtality:
Closing this PR since https://github.com/hummingbot/deploy-examples is the updated method that users should use to install Hummingbot with Docker

PR #6131, opened 2023-03-05
Botcamp/ Cross Dex Arbitrage Script

riven314:


fengtality:
## Botcamp cohort #3 submission review

Great job @riven314 in navigating the current state of the Gateway codebase to successfully submit a solid script! 

### Review comments

- Adding AMM price and balance fetchers as a data feed is a really elegant way of handling those common tasks. I'll talk to @cardosofede about adding those 2 data feeds into the official codebase.
- Your custom `status` function with the arbitrage monitor looks great. I would suggest tracking the balances as well in your status:

```
        lines.extend(["", f"Gas token bal: {self.balance_data_feed.balances[self.chain].balances[self.gas_token]}"])
        lines.extend(["", f"Base bal: {self.balance_data_feed.balances[self.chain].balances[self.base]}"])
        lines.extend(["", f"Quote bal: {self.balance_data_feed.balances[self.chain].balances[self.quote]}"])
```

- There might be a bug with your gas token level check. I ran this on BSC `mainnet` and was seeing messages about topping up my BNB level, but I have 0.12 BNB and verified that the datafeed was fetching that value correctly.

### Future extensions

If you want to keep working on this strategy, I suggest the following:

- Generally, single chain DEX arb will be difficult with Hummingbot, since winners are competing at the MEV level. You should find more success with cross chain DEX ab.
- If you want to stay within BSC, I think when Hummingbot supports Uniswap V3 DEXs like Izumi or Pancakeswap's new perp DEX should open up more interesting script opporunities that don't rely on speed in the same way that arbitrage does

Issue #6021, opened 2023-01-20
Exchange base order_book_ask_entries / bid method returns incorrect orderbook prices


leocape:
**Describe the bug**
// A clear and concise description of what the bug is.

The helper methods `order_book_bid_entries` and `order_book_ask_entries` in `exchange_base.pyx `return incorrect prices for the orders in the order book, for example price 17.77 with order_quantum 0.01 returns 17.76.

Context:
These methods return an iterator of the local orderbook, with the `ClientOrderBookRow` class being the iterator type of which it's price and amount type is `Decimal`.

For performance, the local orderbook is a class of `OrderBookRows` which keep their prices in `float` (faster than `Decimal`), so these prices need to be converted into `Decimal` to suit the return type of `ClientOrderBookRow` in the methods above.

The methods above however when converting from `float` to `Decimal` do not convert the float to `String` first, but are converting directly to `Decimal`, which is causing rounding issues. Further to that, as these prices and amounts are already quantized (they come directly from the connector / exchange), they do not need to be quantized again at this point.

**Steps To Reproduce**
// A concise description of the **steps to reproduce the buggy behavior**:

Difficult to reproduce without providing an entire strategy and as the rounding error only happens on certain numbers , but can be seen inside a strategy while using a non liquid pair (so that the orderbook is not moving quickly) whose price quantum is 0.01, when calling the method:

```
for order in self._exchange.order_book_ask_entries(self._trading_pair):
            top_ask_price = order.price
```

This top ask price is incorrect, in my testing the exchange order was 17.77, and the top_price here was 17.76.

**Screenshots**
// If applicable, add screenshots to help explain your problem.
na
**Release version**
// Include your bot's version number (Can be found at the upper left corner of your CLI).
1.11.0 (was present in previous versions too since the diff mentioned above)
**Attachments**
// Please attach your config file and log file to help us debug. It would be difficult for us to help you without those! 
na

*WARNING: Do NOT publish any exchange API keys or your wallet's private key. Whoever has access to them may steal your assets!*

**Optional: your discord username:**
leocape

discord chat:
Last message in conversation:
https://discord.com/channels/530578568154054663/557343500908363786/1067876648915050627

leocape:
![image](https://user-images.githubusercontent.com/67806134/214594172-1b31150c-4b9e-4417-97fa-80ee1a62fc6f.png)

MementoRC:
[...]

PR #6149, opened 2023-03-15
(fix) remove quantize and add str transformation before decimal conve…

cardosofede:

This PR address issue #6021, by doing the following:
- Removing the quantize operation after receiving the value of processed by the order book (for me this doesn't make sense since the information is already in the order book, which means that it's quantized beforehand)
- Convert floats to str before creating the decimal.

Run the format status example that is querying the order book and check that is working properly.

**Tips for QA testing**:
- Test strategies that are using the OrderBook. You can use the format status example to check if the values are reasonable.
- Ask the user that submitted the bug @leocape to check if it's solved.


nikspz:
Tests performed:
- Cloned and installed feature branch
- connected gold and silver connectors
- Ran the format status example scrypt and check that is working properly querying the order book
- checked pureMM and XEMM strategies status and status --live prices matched with the exchange
- checked order_book --live matched with the exchange for gold and silver connectors (binance, ascendex, gate.io, kucoin)
Pending tests:
pending for user provide script to reproduce issue #6021

hummingbot/connector/exchange_base.pyx
```diff
@@ -168,9 +168,9 @@ cdef class ExchangeBase(ConnectorBase):
         cdef:
             OrderBook order_book = self.c_get_order_book(trading_pair)
             OrderBookQueryResult result = order_book.c_get_vwap_for_volume(is_buy, float(volume))
-            object query_volume = self.c_quantize_order_amount(trading_pair, Decimal(result.query_volume))
-            object result_price = self.c_quantize_order_price(trading_pair, Decimal(result.result_price))
-            object result_volume = self.c_quantize_order_amount(trading_pair, Decimal(result.result_volume))
+            object query_volume = Decimal(str(result.query_volume))
```
MementoRC, line 171: I don't follow why we do this "blind" `str` conversion. I understand that the representation of an arbitrary human-defined float (i.e. a precise rational with a given precision) is perceived as wrong with say Decimal(0.03), thus requiring Decimal("0.03") - note that this is NOT str(0.03)  - but Decimal(str(float)) seems to be a blind and implicit formatting of the float. I think it would be much clearer to specify the precision that we expect with the conversion

MementoRC: We could be specific, `str(float)` will vary depending on the machine you use (I think, it would, back in the days of Sun spark station double 16digits vs pentium 18digits  when I was writing a finite-element code). For instance, we could write `Decimal(str(round(float, 9)))` or `Decimal(f"{float:.9}")` - Anyway, likely knit-picking, I think the str() gives the illusion of infinite precision like in Decimal("0.03") == Decimal(str(0.03)), which I think is rarely true

nikspz:
User has not provided script to test the issue to reproduce, will close the PR for now, will re-open if user will respond

nikspz (approved):
Test performed:
Ran script provided by @riven314 in #6244 using kucoin and binance spot exchange
Ran same script on development and review issues for the incorrect amount, which is not saw testing this branch fix.
Binance on development:
image


This PR6149 fix:
image

PR #6162, opened 2023-03-23
Feat/persistent client session

MementoRC:

Proposing a Persistent aiohttp.ClientSession() wrapper using a WeakSingletonMetaclass to implement the singularity and persistence. This class uses EventLoop in Thread in order to workaround the necessity to start the ClientSession within a running EventLoop, while still conserving the capability to be called synchronously.
Similarly, the `close()` method of the ClientSession is called when the session goes out of hard-reference (using a finalizer) or when the PersitentClientSession instance is garbage collected (using the `__del__`) method to close the ClientSession (even if the loop on which it started is closed) - This is likely still clunky, i.e., the GC is likely to happen late in the life of these objects

The key driver for this proposal is the non-conforming use of ClientSession.

The primary use will be to go through a full refactor of the async test suite in order to use the IsolatedAsyncioTestCase, which kills the main event loop, on which myriad of tests are erroneously depending and cannot allow spot usage

**Tips for QA testing**:


MementoRC:
Ok ... messed it up again with no clues as to why these completely unrelated changes got added to my branch

MementoRC:
Not sure why these commit show-up, the files changes looks right

PR #6164, opened 2023-03-23
(fix) upgrade KuCoin's symbols API version.

pgold:

Upgrade KuCoin's symbols endpoint from v1 to v2, as v1 has been deprecated and might return currencies that do not match those returned by other endpoints (for instance, the accounts endpoint) (see deprecation notice: https://docs.kucoin.com/#get-symbols-list-deprecated).

Ran PMM strategy on KuCoin and verified that the changes worked (whereas the strategy didn't work before the fix, due to a mismatch of names returned by the different API endpoints, leading to a currency showing up in the balance with correct amounts, but this balance wasn't taken into account by the strategy). 

**Tips for QA testing**:

The pairs I've seen affected on KuCoin are those with a lower case character in the name of one currency (e.g. CUSD-USDT).


cardosofede (approved):
Based on this pop-up: 
image
This change should be important.
If the PMM strategy works we should approve it! @nikspz @rapcmia

nikspz (approved):
Test performed:

- Clone and install fix branch
- Created/start pureMM strategy using kucoin, tried different pairs
- review balance for cUSD showed now for rate oracle source kucoin and global token USDT (crosschecked showed 0 on current dev, so PR fix it)

nikspz:
Merged the PR in behalf of foundation team to development and will be available on the next version 1.15.0

Ibrahim7388:
Receive

PR #6199, opened 2023-04-03
(feat) Tweaks to Hummingbot environment

MementoRC:

Proposing a first-pass clean-up for the conda environment for hummingbot
   - Updated install/uninstall to prompt for the conda environment
       . Make the environment name consistent with the environment[xxx].yml
   - Update the setup.py to remove redundant `install_packages`
   - Removed obsolete packages:
      . zero_ex was no longer required
The environment-linux.yml was the only one changed since other environments are not available

Fresh install and make tests

**Tips for QA testing**:


MementoRC:
@rapcmia @cardosofede I di not understand why is this failing. It is completely clean at my end, I did not touch bittrex at all, just some minor changes in the environment

hummingbot/connector/derivative/kucoin_perpetual/kucoin_perpetual_api_order_book_data_source.py
MementoRC, line 1: Ok, this seems to be due to a wrong setup of the test, in which a 'Z" is added to a timestamp that already has a timezone info in the '+00' style

MementoRC: Also, I did not notice the copilot adding a fix of his own:
# There is a badly formed timestamp in the response, so we use the current time instead.
which is not what we should do

MementoRC:
This is now superseeded with: https://github.com/hummingbot/hummingbot/pull/6378, which focus on the environment.yml and install scripts - the workflow proposal will have to be revisited after @cardosofede recent refactoring

PR #6219, opened 2023-04-12
(fix) add Offline to list of market states

vdmerweandre:

When there is an **_Offline_** market status returned from the Btc Markets api, the exchange connector is trying to retrieve the fees for the market, but can not find the mapping since the Offline markets are filtered out at the mapping step.  

Run unit & integration tests

**Tips for QA testing**:
Only occurs when there is an Offline market available in the api like there currently is for ADA: https://api.btcmarkets.net/v3/markets


nikspz (approved):
https://docs.btcmarkets.net/#tag/market_status

PR #6229, opened 2023-04-13
Feat/simplify dockerfile

cardosofede:

Improve the installation process by removing all environment.yml files that are platform specific. Also, the Docker workflow is simplified to one file for all architectures.
The final structure has only 1 environment.yml file under the folder setup/ and 1 Dockerfile.
install.sh was modified too, and now is using the same file.

If you want to build the image:
`docker build -t hummingbot/hummingbot:multi_arch_test .`

In case that you want to build and push the results you can use buildx. To create a new instance

`docker buildx create --platform linux/arm64,linux/amd64`

`docker build use $name_of_buildx_instance`

`docker buildx build --platform linux/arm64,linux/amd64 -t hummingbot/hummingbot:multi_arch_test --push .`

That will upload a version for all architectures directly. Probably this is a good task for @david-hummingbot and @rapcmia to see how it works. Based on ChatGPT this are all the available platforms:

- linux/amd64
- linux/arm64
- linux/arm/v7
- linux/ppc64le
- linux/s390x
- windows/amd64
- windows/arm64

Please check if that's true, to create the corresponding buildx that works for all platforms.

Here you have a refactored docker-compose.yml that works with the new dockerfile: https://gist.github.com/cardosofede/8cd795b26d2989cad376ba231b506955

Check that now we need to specify the full path: /home/hummingbot/data....

- uninstall and install conda environment
- build docker image and run it
- push image, clone it again and run it

**Tips for QA testing**:
- Run ./uninstall and ./install in all OS (windows, ubuntu, Mac) in all arch(amd64, arm64,..)
- Build docker image and run it from all combinations mentioned before
- Build an image with buildx that works in all architectures and upload it to hummingbot registry
- Run the compose file with the new image created by QA in all combinations of OS and Arch mentioned before


klpanagi:
Awesome work @cardosofede !!
Beyond multi-arch support, the miniconda base image is optimized for both size and performance!

Dockerfile.conda
```diff
@@ -0,0 +1,76 @@
...
+
+ENV INSTALLATION_TYPE=docker
+
+# Install system dependencies
+RUN apt-get update && \
+    apt-get install -y sudo libusb-1.0 && \
+    rm -rf /var/lib/apt/lists/*
+
+# Create mount points
+RUN mkdir -p /home/hummingbot/conf /home/hummingbot/conf/connectors /home/hummingbot/logs /home/hummingbot/data /home/hummingbot/certs /home/hummingbot/scripts
```
fengtality, line 62: Is a mount point needed for `/home/hummingbot/conf/strategies`?

fengtality (commented):
See comment

Dockerfile.conda
cardosofede: I will need to do some tests but probably can be there. I think that probably is not necessary since you don't have it at the start but I can check

cardosofede:
@klpanagi I removed the volumes that where created after the mount. What do you think about this, and which in use cases do you think can be better to have it?

MementoRC:
I would suggest to rather use this environment: https://github.com/hummingbot/hummingbot/pull/6199/files#diff-13e9c25e4720c461f255e78eb529ddf7a99cc58f524bf70a7e65e637d2d760ff

MementoRC:
It seems still valuable to have environment tuned to architectures, in some rare cases some conda packages cannot be found for an architecture and needs to be replaced by a pip

If you look at the PR I proposed you will see that most of the pip packages have conda available and should rather use conda to create a consistent environment. The PR I propose is a first step toward migrating to python 3.9 to python 3.11 and tuning the environment to have non-conflicting dependencies for each architecture

Ideally, there would be an additional script that would evaluate whether the pip packages have equivalent conda replacement and clean-up the environment

It is a bit unfortunate that we end-up having different solutions for intertwined problems :/

rapcmia:
PR update:
- Run `uninstall, clean` and `install`, all ok 💯 
    - WSL ubuntu 20.04 ✅
       image
       - Create local docker image using `docker build -t hummingbot:hummingbot:$tag .`
- Ran `buildx` to test dockerhub repository
    - `docker buildx create --name test-6229-01 --platform linux/arm64,linux/amd64` ✅
    - `docker buildx use test-6229-01` ✅
    - `docker buildx build --platform linux/arm64,linux/amd64 -t hummingbot/hummingbot:multi_arch_test --push .` ✅ 
    - Successfully uploaded → [https://hub.docker.com/repository/docker/rapcmia/test01/tags?page=1&ordering=last_updated](https://hub.docker.com/repository/docker/rapcmia/test01/tags?page=1&ordering=last_updated)
       image
    - MBP(m1) running on anaconda3/osx-64 ✅
      image
       - Successfully build local docker image 
- Ran tests on using helper scripts
    - Helper scripts working found on d`eploy-examples/bash-scripts/` , all ok
    - Create docker using local and uploaded image ✅
       image
        - MBP (m1)
       image
        - WSL ubuntu20.04

Most of the tests are for Client installation and still looking for other possible cases. We should also keep in mind that the approach on uploading a image to dockerhub using `buildx` took around `40mins`
image


Pending:
- Updating version
- Gateway
- Docker compose
- Observe client + gateway when setting up with strategies and longterm

rapcmia:
PR update:
- Test updating version using uploaded docker image built using `buildx`, all ok 
- Docker compose simple and multiple bots ✅

cardosofede:
@MementoRC I just saw your comments. I discussed it with @fengtality and this solution will simplify the workflow and the extra dependencies are not so heavy for the user.
The proposal for you is to create a new environment.yml with the upgrade for python 3.9-3.11 and have those 2 environment files and the user can select with python version he wants to use. What do you think?

MementoRC:
Sure, I already have a 3.9 conda focused, but it requires some clean-up of the code (I thought there was already a PR for that)

I am supportive of the improvement. I do not understand what is wrong with the environment.yml that I proposed. It only makes the conda environment more consistent.

cardosofede:
@MementoRC you mean the dependencies with >= ...?
I like that but at some point how we can make sure that the new version of those dependencies doesn't have a new api?
Also, I think that you removed some dependencies that are not needed and that is great too.
Once this PR is merged (that is more focussed on simplifying Docker) submit your PR. We should assign a bounty @carlitogetaladajr to memento for his work

rapcmia:
PR update:
- Test latest commit `fb86656a97de0f`
- Check behavior of multiarch approach on test docker repo when uploading a new image
   image
   - Successfully uploaded and append the old image 💯
- Setup gateway docker
   image
    - Run `gateway generate-certs`, ok on client
    - Check the folder, no gateway files created
    - Tried to change permission to `sudo chmod -R a+rw`  and generate certs again, still no go
    - Build a local image and upload to test dockerhub repo
        - Build docker client and generate gateway, still no go ❌ 
    - Crosscheck the development branch
        - Pull latest docker `development image`
        - Generate gateway certs and confirm the files are available ✅ 
          image
    - Tried to create image using docker build -t hummingbot/hummingbot: $TAG -f Dockerfile . to check if this works using the Dockerfile 
      image
        - On the screenshot above seems successful yet when checking `docker images` it is not available. Image name was `pr6229b`
        image
- Test with strategies and connectors
  image
    - Connect binance API key successfuly on client
    - Check folder, no files under `conf/connectors`
    - This is the same scenario for strategies too
       image
    - It is also observed that .password_verification is not available. On the screenshot below is sample from development branch which is not the same on this PR
      image
- Test hummingbot-update.sh
  - Confirmed able to select the correct architecture if its `AMD64 `or `ARM64` ✅ 
  - However it does not carry over the `password, connectors and strategies`. When starting the updated container, it starts from scratch ❌ 


Steps to reproduce:
- Build local image ( optional: `then upload to test docker repository` )
- Use the [bash_scripts](https://github.com/hummingbot/deploy-examples) to create hummingbot docker client
- Observed behavior after successfully connected a exchange and created strategy
  -  Check folder directory where the docker container files are stored e.g `hummingbot_files/conf/..`
- Observed behavior when setting up docker gateway
- Observe behavior when successfully run `hummingbot-update.sh`

For testing uploaded multiarch you may use our test repository: https://hub.docker.com/repository/docker/rapcmia/test01/tags?page=1&ordering=last_updated

rapcmia:
> * Tried to change permission to `sudo chmod -R a+rw`  and generate certs again, still no go

Issue related to missing files on `hummingbot_files/conf and /certs` folder has been fixed on this PR commit https://github.com/hummingbot/deploy-examples/pull/14/commits/85ebeda280e6c03b97a8e19d5ede8634e22f27a3 ✅ 

Test in progress

fengtality (approved):

PR #6273, opened 2023-04-29
(fix) Proposing update to AsyncCallScheduler + testsuite

MementoRC:

Proposing an update to the AsyncCallScheduler class. It seems that in some situations, this class mixes-up the Event Loop it is supposed to work with. This seemed to be happening during the effort to re-factor the tests that incorrectly (IMHO) uses the main event loop. Some tests using a local new_event_loop for clean start/stop would initialize the AsyncCallScheduler with the main loop, the sync test would block the event loop and the asyncio.run() would attempt to use the AsyncScheduler, which worked with the main loop (asyncio.run() creates its own local loop)

It also appeared that there were no test suite for this class, the primary contribution is to add the tests, of course the tests maybe skewed towards working with the new AsyncCallScheduler

Ran the new testsuite successfully

**Tips for QA testing**:


MementoRC:
I have not figured out how to have my own development branch in my repo so that I can add the feature that I want, yet being able to merge feature that I think may be merge to the official dev branch :'(
Why is this separate PR: `(feat) Proposing a multi-arch install/Build/Test workflow` showing here?

MementoRC:
Re-submitted

PR #6381, opened 2023-06-17
Fix/test class for isolated asyncio

MementoRC:

Proposing a fix to the impossibility to use IsolatedAsyncio TestCase due to it closing the main event loop and causing all tests afterward to fail with:
`RuntimeError: There is no current event loop in thread 'MainThread'.`

Using asyncio.run() causes the same issue (used in Coinbase Advanced Trade tests to cleanly isolated async tests)

This proposal corrects the behavior of the previously proposed decorator (async_to_sync) and adds wrappers to 3 event loop use cases:
   - IsolatedAsyncioTesCase: Simply wraps the class with preservation and re-assignment of the main loop
   - Class-level usage of an event loop: Creates a local loop while preserving the main loop
   - Test-level usage of an event loop: One event loop per test

Tried to provide an exhaustive test suite to verify the expected behavior

**Tips for QA testing**:


MementoRC:
Man that is a major annoyance, or at least I am a major moron, still thinking both are not mutually exclusive ... the pain though!

PR #6382, opened 2023-06-17
(fix) Run IsolatedAsyncio in HB

MementoRC:

Proposing a fix to the impossibility to use IsolatedAsyncio TestCase due to it closing the main event loop and causing all tests afterward to fail with:
`RuntimeError: There is no current event loop in thread 'MainThread'.`

Using asyncio.run() causes the same issue (used in Coinbase Advanced Trade tests to cleanly isolated async tests)

This proposal corrects the behavior of the previously proposed decorator (async_to_sync) and adds wrappers to 3 event loop use cases:
   - IsolatedAsyncioTesCase: Simply wraps the class with preservation and re-assignment of the main loop
   - Class-level usage of an event loop: Creates a local loop while preserving the main loop
   - Test-level usage of an event loop: One event loop per test

Tried to provide an exhaustive test suite to verify the expected behavior

**Tips for QA testing**:


cardosofede:
@fengtality I think this PR is good and will be very helpful to refactor the current tests we already have. 
I don't think that we should be using the Isolated approach because as I discussed with Memento we should rewrite all the tests at once, and if we are doing that I prefer to use Pytest, that already has support for async test with the decorator @pytest.mark.asyncio.
Good job @MementoRC

MementoRC:
Thanks @cardosofede I realize that I should add some clean-up when tasks are started within the test. Any other cases that need to be investigated?
Given the interest, I created a HIP, I am flexible on the HBOT, I just put a number close to the incentive rewards
https://snapshot.org/#/hbot-ip.eth/proposal/0x94fec9fe725d2bcecdeda86af9cdab6592a414275baeaafb7ce3f18a42d8b910

cardosofede:
@MementoRC I think that the requested amount of HBOT's can be higher since it's going to be necessary. What do you think about 50k?

MementoRC:
@cardosofede Sounds cool. I used to write the tests in Pytest, but after a year of HB I have forgotten, but it should not be to difficult, actually I often use pytest in Pycharm. I could investigate

cardosofede (approved):
LGTM! Good job

Issue #6209, opened 2023-04-11
Integrate fixed_rate_conversion into amm_arb strategy


hui121315:
## Problem Definition

Currently, the amm_arb strategy calculates arbitrage opportunities using real-time exchange rates. In some cases, it may be desirable to use a fixed exchange rate instead of a real-time one to provide a more stable price difference when calculating arbitrage opportunities.

## Feature Request

I would like to request the integration of the fixed_rate_conversion feature from the cross_exchange_market_making strategy into the amm_arb strategy. This feature would allow users to set a constant exchange rate between the two assets they are trading, instead of using the real-time exchange rate.

## Implementation Steps

1. Add a new variable `fixed_conversion_rate` to the amm_arb strategy with a default value.
2. Modify the `AmmArbStrategy` class's `__init__` method to include the new 
3. Update the strategy configuration file to include the new `fixed_conversion_rate` parameter.
4. Update the documentation to describe the new functionality.

Please let me know if you need any additional information or clarification on this feature request. Thank you!

fengtality:
## Bounty

**Sponsor**: Hummingbot Foundation
**Rationale**: AMM ARB is a Core Strategy for Epoch 2, which means that the Foundation will be the official maintainer for these strategies, prioritize updates and fixes to them, and maintain their documentation pages.
**Assessment**: P2: General / Fixable in 1-2 days
**Bounty Amount**: 50,000 HBOT or 250 USDT
**Developer Share**: 100% - Paid upon merge to `development`
**Status**: Open

fengtality:
I think this is a well-defined enhancement request to the AMM-ARB strategy: https://docs.hummingbot.org/strategies/amm-arbitrage/. If devs interested in taking this bounty have questions about implementation, please ask below.

srirambandi:
Hi @fengtality, I'd like to take up this bounty.

srirambandi:
Thanks Mike. Will start my work on this from tomorrow. Will update everything in this thread and a draft PR I'll submit for this bounty.

fengtality:
@srirambandi any update on this?

srirambandi:
Hey Mike! Will do a draft PR with the commits tonight 🙌 Need yours or someone’s review to check the logic of that later. I was working in crypto_com connector and logic, and it has come to a good stage, pending just 2 more unit test files.

Anyway will update here with the draft PR tonight, thanks.

nikspz:
[...]

PR #6422, opened 2023-06-29
feat/ Integrate fixed_rate_conversion into amm_arb strategy

srirambandi:

This is a PR(Draft) to solve #6209. Integrate fixed_rate_conversion into `amm_arb` strategy

NA

**Tips for QA testing**:

NA


fengtality (changes requested):
Looks good overall!

I think you can simplify this by having 2 params:
- `rate_oracle_enabled`: bool, default False
- `fixed_conversion_rate`: float, default 1

You can make the first question "Do you want to use the rate oracle?" and the 2nd question works well.

I believe these changes would result in a better user experience.

srirambandi:
Got it! Will push with new changes. Thanks for the quick review!

fengtality (changes requested):
These changes looks good to me.

I also think we should handle the case where the user need to convert the base tokens. If the user is converting base tokens, 

Therefore, please revise the logic to:

1. If either base-base or quote-quote symbols are different
2. For each one, if there is a difference, ask user if they want to use rate oracle to handle each conversion (default = No)
3. If user confirms No, then ask user to define fixed rate

I believe this allow the strategy to handle both base and quote conversions.

In addition, please add unit tests to check that this works correctly under different conditions, such as:
1. base tokens are different, conversion rate is fixed
2. quote tokens are different, conversion rate is fixed
etc

rapcmia:
> These changes looks good to me.
> 
> I also think we should handle the case where the user need to convert the base tokens. If the user is converting base tokens,
> 
> Therefore, please revise the logic to:
> 
> 1. If either base-base or quote-quote symbols are different
> 2. For each one, if there is a difference, ask user if they want to use rate oracle to handle each conversion (default = No)
> 3. If user confirms No, then ask user to define fixed rate
> 
> I believe this allow the strategy to handle both base and quote conversions.
> 
> In addition, please add unit tests to check that this works correctly under different conditions, such as:
> 
> 1. base tokens are different, conversion rate is fixed
> 2. quote tokens are different, conversion rate is fixed
>    etc

Hi @srirambandi good day to you
Can you please update base on requested changes so we can have a look? thank you

nikspz:
Closing this PR since there're no response from the developer after a week

PR #6428, opened 2023-07-03
Fix/dexalot failed transaction status

Beastlorion:

I've added a condition that accounts for failed transactions. Previously, failed transactions were treated as pending and got stuck in a loop to find their status, which would make constant http calls and drain cpu processing power.

FYI this is my first public pull request in a very long time. Let me know if I need to change anything.

I'm using this in live instances and it's working with no errors.

**Tips for QA testing**:
Force a failed transaction, perhaps by lowering the gas limit below the needed amount, or by populating the sqlite file for a certain market with a pending order that links to a failed transaction on-chain.


fengtality:
please rebase this PR against the latest version of `development` as it looks like there are unrelated changes here

PR #6501, opened 2023-07-23
fix/Binance rate limit weight for ticker price change

supervik:

Request weight should be equal to **1** in binance_constants for 24h ticker price change (/api/v3/ticker/24hr)
It is equal to 40 when symbol parameter is omitted, but here we request one trading pair statistics so the correct weight is 1
https://binance-docs.github.io/apidocs/spot/en/#24hr-ticker-price-change-statistics

I checked the output of the **_get_last_traded_price** function in **binance_exchange.py** where TICKER_PRICE_CHANGE_PATH_URL is used and confirmed that it actually returns the data for one trading pair

**Tips for QA testing**:


cardosofede (approved):
Nice catch! I reviewed and the method that you mentioned is the only one that is using that endpoint and the request only contains 1 trading pair so you are right!

PR #6512, opened 2023-08-04
Feat/injective spot offline vaults

aarmoa:

This PR expands the capabilities of the Injective V2 spot connector, adding logic to support trading through off-line vaults.

The PR also includes a refactoring to the connectors's configuration in order to allow connectors to use full Pydantic configurations (meaning that the connectors now can have configuration models with N submodels, each of them supporting the use of secret parameters)
The new version is backwards compatible and the connectors will still be using the old configuration style model, unless they explicitly declare the use the new module (with this PR only Injective V2 spot connector uses the new configuration model).

There is also an improvement in the logic that checks the orders are included in a transaction. Now the connector updates the order hashes in case they were incorrectly calculated locally in the bot, to reduce the chances of producing lost orders.

Added unit tests for the new logic.
All tests passing in green.

**Tips for QA testing**:
Validate that old connectors can still be configured as usual.
Validate that the private key for Injective V2 connectors is stored in the config file encrypted

This PR is a continuation of the work done in https://github.com/hummingbot/hummingbot/pull/6493. No need for a PRP.


rapcmia:
PR update:
- Test latest commit `904786af41`
- Check old connectors {GOLD, SILVER} due to changes on configuration
   - Connect exchanges API and tested invalid keys ok
   - Check connector `.yml` files and api/secret are encrypted
- Setup injective_v2:
   - Connector now selects between mainnet, testnet and custom
   - Connector status now only shows `injective_v2` only, 
      - _How would I know if which network connected?_ 
         - We think the only way to know is by checking the `balance`
            image
             - On the screenshot we setup testnet and from the `balance` it is noticeable huge amounts 
      - _How would I change from different network ( testnet -> mainnet/custom )?_
         - User must run the command again like normal CEX setup and select the network
      - _Validate that the private key for Injective V2 connectors is stored in the config file encrypted?_ 
          - Yes, checking `injective_v2.yml` pkey is encrypted when setting up `testnet`
   - Connector now support two types of account configuration `{delegate_account, vault_account}`
      - `vault_account` not yet tested `pending`
   - `custom_network` not yet tested `pending` 

---

Ran tests on `mainnet`

image
   - When selecting network, there is a added value of `node=’lb’`
   - When selecting `mainnet_network` getting error of `Can't instantiate abstract class InjectiveNetworkMode with abstract methods network, use_secure_connection` (see screenshot attached above)

---

Ran tests on `testnet`

image
   - On the prompt of type of account config, there are added characters also
   - Only on testnet was able to move from the setup and there are two types of account type `{delegate_account, vault_account}`
   image
     - Able to create and cancel orders ✅
     - Trade data between client history and CSV matched on testnet trade history

rapcmia:
- Test latest commit `a2a33874afb1bb`
- Mainnet is now fixed and able to setup `delegate_account`
- However there are still UI issuewhen selecting network and account configuration 
   image
- For `vault_account` config, unfortunately this is not available on our resource (seems referring to smart contract_base vault for managing assets) so QA can't further look into 🙇🏼

aarmoa:
Hello @rapcmia 

> When selecting mainnet_network getting error of Can't instantiate abstract class InjectiveNetworkMode with abstract methods network, use_secure_connection (see screenshot attached above)

Good catch, I have pushed already a fix for this.

> When selecting network, there is a added value of node=’lb’
> On the prompt of type of account config, there are added characters also

The connector has a default configuration that is used when querying for the available trading pairs. And those characters you see are the way Hummingbot displays the Pydantic configuration instances (the default way for Pydantic). If you want to change that it will be a task for the foundation. Changing the way Pydantic is used in Hummingbot is outside the scope of this PR (@cardosofede)

rapcmia (approved):
LGTM
- Connector does not require gateway connection 
- Check `{binance, kucoin, ascendex and gateio}` spot and perpetual conenctors due to changes on connector configration usign full Pydantic config 
- Setup Injective_v2
  - Connector status now only display `injective_v2`
  - On setup, user can select between:
      - Network to connect `{mainnet_network, testnet_network, custom_nework}`
          - `custom_network` no tested ❌ 
      - Supports two account configuration `{delegate_account, vault_account}`
          - `vault_account` not tested ❌ 
  - Successfully connected wallets for mainnet and testnet
  - Ran balance command, ok 
  - Ran tests on PMM for both networks
- Fixed issues related to mainnet 
- Build docker image and did the same tests

---

_There are minor UI issues related to Pydantic config as adviced by developer. The foundation team will have a look on this concern. Other than that PR is good to be approve under bronze standards. Thank you_

cardosofede (approved):
LGMT!

Issue #6517, opened 2023-08-04
Add Coincap.io as the new rate oracle source


nikspz:
Description: 
Rate oracle: https://coincap.io/
We're looking to integrate the coincap.io rate oracle into Hummingbot. 
A lot of Rate oracles do not work very well, not working with some exotic trading pairs or have rate limits too low.
We need rate oracle source that should be more `reliable`, with the good implementation in the Client.
The adding new rate oracle should expand the range of supported cryptocurrencies. 
The rate oracle should provide accurate real-time price data. The successful integration should be documented and compatible with our existing trading bot architecture.

Sponsor: Hummingbot Foundation
Developer share: 100% - paid upon merge to development
Bounty Amount: 100,000 HBOT
Status: Assigned

petioptrv:
@nikspz, here is the PR for this issue: https://github.com/hummingbot/hummingbot/pull/6524

petioptrv:
@nikspz, the code is undergoing internal QA review. Should be submitting it soon.

rapcmia:
`Coincap` has been merged to development and will be available on `release version 1.120.0`

nikspz:
Closing this bounty as paid, feature available on latest development  and release version 1.20.0

PR #6534, opened 2023-08-22
Feat/coin cap rate source

petioptrv:

This PR is associated to bounty issue: https://github.com/hummingbot/hummingbot/issues/6517.

This PR introduces the [CoinCap](https://coincap.io/) rate source for the rate oracle. It is possible to obtain an API key [here](https://coincap.io/api-key).

CoinCap adds an alternative to CoinGecko, and its API allows for price streaming, which is included in this PR.

Note that CoinCap's API uses asset names rather than symbols, which means users must configure an asset map like `'BTC:bitcoin,DOGE:dogecoin'` to the `assets_map` of the rate source's config.

Note also that the only consequence of a faulty API key (or no API key) is that the user will have lower rate limits. If a faulty API key is detected, the client will issue a warning log.

The PR had to make significant changes to the configuration modules in order to allow configuring a secret API key in the client config map.

- Unit tests added.
- Manually tested the functionality.

**Tips for QA testing**:

Obtain an API key [here](https://coincap.io/api-key) and run a full test on the functionality of the rate provider. Please not that the global token must be configured to `USD` in order for `rate --token {TOKEN}` to work. That is because CoinCap only supports USD as a quote, and that rate functionality cannot perform conversions. However, the `rate --pair {BASE}-{QUOTE}` will work with pairs that don't include `USD` as quote.


rapcmia:
PR update:
- Test latest commit `d73b100fe1f9`
- Check [Coincap](https://coincap.io/) and [created API key](https://coincap.io/api-key)
- Ran `config rate_oracle_source`
  image
  - How to add multiple symbols? 
     image
  - What happens if invalid API key?
     image
  - What happens if incorrect global token?
     image
       - Global must be set to USD when using coincap using `config global_token.global_token_name`
- Setup AMM-arb on pancakeswap
    - WBNB not supported on coincap, still works ok
    - Added multiple symbols and displays the balance ok
    - Start with not adding USDT on symbols getting:
      image
       - Added USDT on `config rate_oracle_sourc.asset_map`
       image
- Build docker image and check rate_oracle coincap ✅

rapcmia (approved):

cardosofede (approved):
LGTM! Thanks for the contribution @petioptrv 
Do you think that are other important things that we can improve in the codebase in terms of configuration?
@rapcmia after finishing testing we can merge it once the test are passing

cardosofede:
@petioptrv yeah totally agree, I think that we should refactor a little bit the settings and the app, I'm working in the distribution of the library and currently the codebase is using the HummingbotApplication for a lot of things and running a lot of process that probably we should omit unless we are looking for them

cardosofede:
@petioptrv I fixed in another PR the issues with the environment. Now the workflow is failing due formatting conventions.

rapcmia:
Merged to development and will be part of release version 1.20.0

PR #6558, opened 2023-08-30
Feat/polkadex v2

petioptrv:

This PR introduces an update to the Polkadex connector to use the [V2 API](https://documenter.getpostman.com/view/16239785/2s93shy9EY#8094c956-8ffe-4522-b442-8fa38f057329).

- Unit tests.

**Tips for QA testing**:

Note that every now and then, the periodic order status update may fail because it happened immediately after submitting an order. Since Polkadex is a decentralized exchange, it takes a couple of seconds for the order to make it onto the chain and to become available for order status updates.

[ch-43708]

### DIFF
--- a/hummingbot/core/event/events.py
+++ b/hummingbot/core/event/events.py
@@ -37,6 +37,8 @@ class MarketEvent(Enum):
 
 class OrderBookEvent(int, Enum):
     TradeEvent = 901
+    OrderBookDataSourceUpdateEvent = 904
+    PublicTradeEvent = 905
 
 
 class OrderBookDataSourceEvent(int, Enum):
--- a/hummingbot/connector/derivative/phemex_perpetual/phemex_perpetual_constants.py
+++ b/hummingbot/connector/derivative/phemex_perpetual/phemex_perpetual_constants.py
@@ -6,8 +6,6 @@ from hummingbot.core.data_type.in_flight_order import OrderState
 EXCHANGE_NAME = "phemex_perpetual"
 MAX_ORDER_ID_LEN = 40
 
-HB_PARTNER_ID = "HBOT"
-
 DEFAULT_DOMAIN = ""
 TESTNET_DOMAIN = "phemex_perpetual_testnet"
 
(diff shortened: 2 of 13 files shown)


rapcmia:
PR update:
- Test latest commit `061d77b6ec42a7`
- Request CA QA team to share account for [mainnet](https://orderbook-internal.polkadex.trade/) and successfully import ✅
- Polkadex available on `connect` command 
- Successfully connected using mnemonic passphrase with balance displayed correct ✅ 
- Setup simple PMM on `DOT/USDT`
   - Observe order fetch status failed which is expected for the connector 
   - Orders are created and cancelled successfully, no stuck order observed
   - Set tight spreads with longer refresh time but unable to have a opportunity for trade with running for 4hrs
- Build docker image and ran tests, all ok

rapcmia (approved):

hummingbot/core/event/events.py
```diff
@@ -37,6 +37,8 @@ class MarketEvent(Enum):
 
 class OrderBookEvent(int, Enum):
     TradeEvent = 901
+    OrderBookDataSourceUpdateEvent = 904
+    PublicTradeEvent = 905
```
cardosofede, line 41: @petioptrv hey pet, the TradeEvent is actually the public trade event check that the order book is triggering it when a new trade arrives

cardosofede (approved):
the changes are okey but just check that and see if you want to implement it in another PR or do the changes on this one.
Btw, I think that we should also trigger the event of the order book updates what do you think?

hummingbot/core/event/events.py
petioptrv: The idea here was to differentiate the "data-source" event from the event that the clients of the exchange connector want to consume, but I agree that, in this case, it's redundant. I'll make the change.

petioptrv: Done, @rapcmia, the only part of the functionality that needs to be re-tested is that public trade events are being received by the bot

cardosofede (approved):

PR #6597, opened 2023-09-30
fix/ position_executor close_price() returns None if take profit type is LIMIT

andreyserdjuk:

If take_profit_order_type is `OrderType.LIMIT`, `self.close_order.average_executed_price` returns `None`.
it causes `TypeError` in all dependent methods (take_profit_condition, stop_loss_condition etc.).

In PR `self.take_profit_order` is used instead of `self.close_order` if take profit order type is LIMIT.

**Tips for QA testing**:

### DIFF
--- a/hummingbot/smart_components/position_executor/position_executor.py
+++ b/hummingbot/smart_components/position_executor/position_executor.py
@@ -112,6 +112,8 @@ class PositionExecutor(SmartComponentBase):
         elif self.executor_status == PositionExecutorStatus.ACTIVE_POSITION:
             price_type = PriceType.BestBid if self.side == TradeType.BUY else PriceType.BestAsk
             return self.get_price(self.exchange, self.trading_pair, price_type=price_type)
+        elif self.take_profit_order_type.is_limit_type():
+            return self.take_profit_order.average_executed_price
         else:
             return self.close_order.average_executed_price
 


cardosofede:
@andreyserdjuk hey Andrey! thanks for contributing. Your PR is pointing to master, and I think that I solved this issue in the current release. Can you check?
If not provide the steps to reproduce

andreyserdjuk:
@cardosofede , I've already checked development and master branches: `PositionExecutor::close_price()` takes only `self.close_order.average_executed_price` (when executor in completed status). But the problem is `self.close_order` used only when `PositionExecutor::take_profit_order_type` is not limit type (see `PositionExecutor::control_take_profit()`).
To reproduce the bug just take any strategy, which extends DirectionalStrategyBase and override `take_profit_order_type: OrderType = OrderType.LIMIT`.

cardosofede:
hey @andreyserdjuk, the PR that you checked before is already solving the issue. You were right with the problem but the solution is to assign as a close order the take profit order in case that this was the last scenario:
https://github.com/cardosofede/hummingbot/blob/ed7a671884dfba1576f1ca5347ea4204d3178ba2/hummingbot/smart_components/executors/position_executor/position_executor.py#L360

PR #6669, opened 2023-11-21
(fix) account for single order level in linear distribution

fengtality:

* accounts for case where `n_levels` equals 1 in linear distribution


cardosofede:
@fengtality the solution is good and we should test if has to be replicated in other distributions, but you need to add a test to cover that line that you are adding

cardosofede (approved):
LGTM!

PR #6674, opened 2023-11-22
added pre-commit install check

TeoVibe:
Added this for convenience since I ran into the issue when installing on a fresh device.

Will also add some dependancy checks before the compile if this is deemed worthy :)


cardosofede:
@rapcmia @nikspz @david-hummingbot This might be interesting, I would also recommend adding the check for build-essentials since it's needed to compile the code.
@TeoVibe also is needed to point the pr to development

rapcmia (approved):

fengtality:
This PR just add the following check to `./install`, so the dev should re-do this PR on a clean branch if they want to get this into the codebase. If this is really helpful, we can re-create the PR.

```
# Check if pre-commit is installed, if not, pip install pre-commit
if ! command -v pre-commit &> /dev/null; then
    echo "pre-commit not found, installing..."
    pip install pre-commit
fi
```

PR #6680, opened 2023-11-25
(fix) fix default button color

fengtality:

- This fixes the default button color, which was previously had a black background and led to the log pane buttons being invisible on certain machines.
- Now they should be visible in all environments:

Screen Shot 2023-11-25 at 12 04 42 PM


cardosofede (approved):
LGTM!

PR #6683, opened 2023-11-26
fix/ cancel and order status for MEXC connector

BuddhaSource:
Something broke MEXC cancel logic, added a fail safe logic to catch cancel.
HB tries to get the status but, in return gets the error "-2001" which means an unknown error and is not handled. 

- [Yes ] Your code builds clean without any errors or warnings
- [ Yes] You are using approved title ("feat/", "fix/", "docs/", "refactor/")

**Tips for QA testing**:


nikspz:
hi @BuddhaSource Could you please fix failing tests?
https://github.com/hummingbot/hummingbot/actions/runs/6997190186/job/19047458574?pr=6683

PR #6698, opened 2023-12-04
Feat/add close price reference controller

cardosofede:

This PR addresses the issue of trading in some pairs in which the naming is different from the source of candles used.
For example, in VEGA the pairs are called ETHUSDPERP-USDT and the binance candles are called ETH-USDT so to make that match I'm providing a new argument in the controllers to specify which is the source of the close price from the candles.
If this parameter is None, the default trading pair will be used.

**Tips for QA testing**:


rapcmia:
Ran tests on binance_perpetual and testnet
- with candles_trading_pairs ✅
- with candles_trading_pairs set to “” and "None"
    image
- with candles_trading_pairs set to None
    image

When `candles_trading_pairs` are set on both "", "None" and `None`, the script does not place orders nor using the set trading_pair market. No error found when running tests

cardosofede:
@rapcmia candles trading pair can't be None, the new parameter is in the controller and it's called close_price_trading_pair that is different from the trading pair and the candles trading pair.
If close_price_trading_pair == None, then the trading pair will be used

rapcmia (approved):
LGTM
- Run tests on `v2_market-making_dman_v3_complex_config.py`
- Observed that the status now shows close_price_trading_pair where it displays the set value 
- with `close_price_trading_pair=""`
   image
    - Script still worked ok and uses the default ETH-USDT
- with `close_price_trading_pair=None`
   image
- Tried using two different trading pair, yes still displays the set `close_price_trading_pair` and expected to have errors since two different markets ✅

Issue #6646, opened 2023-11-08
Gateway - amm_trade function do not work correctly with small prices.


jellebuth:
### Describe the bug

Whenever the order price on Gateway's "amm_trade" function in "hummingbot/core/gateway/gateway_http_client.py" is too small. e.g. 
![image](https://github.com/hummingbot/hummingbot/assets/83583507/7759e8cf-c1c7-42a8-bf94-f3f68135b7f6)

It was unable to create an order due to the following issue:

Error: {'message': 'The limitPrice param may be null or a string of a float or integer number.', 'httpErrorCode': 404, 'errorCode': -1, 'stack': 'Error: The limitPrice param may be null or a string of a float or integer number.\n    at throwIfErrorsExist (/Users/jellebuth/Downloads/gateway/dist/src/services/validators.js:38:15)\n    at /Users/jellebuth/Downloads/gateway/dist/src/services/validators.js:103:40\n    at /Users/jellebuth/Downloads/gateway/dist/src/amm/amm.routes.js:25:51\n    at Generator.next (<anonymous>)\n    at /Users/jellebuth/Downloads/gateway/dist/src/amm/amm.routes.js:8:71\n    at new Promise (<anonymous>)\n    at __awaiter (/Users/jellebuth/Downloads/gateway/dist/src/amm/amm.routes.js:4:12)\n    at /Users/jellebuth/Downloads/gateway/dist/src/amm/amm.routes.js:24:85\n    at /Users/jellebuth/Downloads/gateway/dist/src/services/error-handler.js:37:28\n    at Layer.handle [as handle_request] (/Users/jellebuth/Downloads/gateway/node_modules/express/lib/router/layer.js:95:5)'} 

I believe it is because of the formatting as when I changed the code to the following:
![image](https://github.com/hummingbot/hummingbot/assets/83583507/92aeccd6-805c-4430-b8dd-8c5ea569122b)

The issue was not present anymore.

### Steps to reproduce

1.  Run AMM Arb on PCS to MEXC on pair ARV-WBNB (PCT) to ARV_USDT (MEXC)
2. Set Min_profitability to -1 to allow for a trade
3. The trade fails as the price input in the "place order" function is formatted as 1.713321E-7 rather than 0.000000017713321

This can be reproduced by:
1.  Run AMM Arb on PCS to MEXC on pair ARV-WBNB (PCT) to ARV_USDT (MEXC)
2. Set Min_profitability to -1 to allow for a trade
[...]

PR #6699, opened 2023-12-05
Fix/ #6646 Gateway Small Order Price

HarjobandeepSingh:
**A description of the changes proposed in the pull request**:
This pull request proposes changes to the amm_trade function in the gateway_http_client.py file of the Gateway module in Hummingbot. The changes are aimed at addressing an issue where the function was unable to create an order when the order price was too small.

The changes include additional checks and error handling in the amm_trade function:

Check if the amount and price are valid numbers: The function now checks if the amount and price are instances of int, float, or Decimal. If not, it raises a ValueError.

Check if the amount and price are not null: The function checks if the amount and price are not null. If either is null, it raises a ValueError.

Check if the price is too small: The function checks if the price is less than 1e-7. If it is, it raises a ValueError.

These changes help to prevent the function from proceeding with invalid input values, which could lead to unexpected behavior or errors. By raising descriptive error messages when these checks fail, it becomes easier to identify and address issues related to small order prices. This should help to prevent the specific error that was previously encountered when the order price was too small.

**Tests performed by the developer**:
The developer performed several tests to ensure the robustness of the changes made to the amm_trade function:

Valid Number Test: Tested the function with amount and price as valid numbers to ensure that the function works as expected in normal conditions.

Null Value Test: Tested the function with amount and price as null to verify that the function raises a ValueError as expected.

Small Price Test: Tested the function with a price value less than 1e-7 to verify that the function raises a ValueError as expected.

Large Price Tested the function with a price value greater than 1e-7 to ensure that the function works as expected when the price is not too small.

These tests were designed to cover a range of scenarios and edge cases to ensure that the function behaves as expected in various conditions. The developer also performed additional manual tests to verify the overall functionality of the amm_trade function.

PR #6706, opened 2023-12-09
fix/ Gateway Small Order Price #6646

HarjobandeepSingh:
*A description of the changes proposed in the pull request*:
This pull request proposes changes to the amm_trade function in the gateway_http_client.py file of the Gateway module in Hummingbot. The changes are aimed at addressing an issue where the function was unable to create an order when the order price was too small.

The changes include additional checks and error handling in the amm_trade function:

Check if the amount and price are valid numbers: The function now checks if the amount and price are instances of int, float, or Decimal. If not, it raises a ValueError.

Check if the amount and price are not null: The function checks if the amount and price are not null. If either is null, it raises a ValueError.

Check if the price is too small: The function checks if the price is less than 1e-7. If it is, it raises a ValueError.

These changes help to prevent the function from proceeding with invalid input values, which could lead to unexpected behavior or errors. By raising descriptive error messages when these checks fail, it becomes easier to identify and address issues related to small order prices. This should help to prevent the specific error that was previously encountered when the order price was too small.

*Tests performed by the developer*:
The developer performed several tests to ensure the robustness of the changes made to the amm_trade function:

Valid Number Test: Tested the function with amount and price as valid numbers to ensure that the function works as expected in normal conditions.

Null Value Test: Tested the function with amount and price as null to verify that the function raises a ValueError as expected.

Small Price Test: Tested the function with a price value less than 1e-7 to verify that the function raises a ValueError as expected.

Large Price Tested the function with a price value greater than 1e-7 to ensure that the function works as expected when the price is not too small.

These tests were designed to cover a range of scenarios and edge cases to ensure that the function behaves as expected in various conditions. The developer also performed additional manual tests to verify the overall functionality of the amm_trade function.


HarjobandeepSingh:
@nikspz other changes that i made in my previous request is already added in main repo of project by your team 

```
    async def amm_trade(
            self,
            chain: str,
            network: str,
            connector: str,
            address: str,
            base_asset: str,
            quote_asset: str,
            side: TradeType,
            amount: Decimal,
            price: Decimal,
            nonce: Optional[int] = None,
            max_fee_per_gas: Optional[int] = None,
            max_priority_fee_per_gas: Optional[int] = None
    ) -> Dict[str, Any]:
        # XXX(martin_kou): The amount is always output with 18 decimal places.
        request_payload: Dict[str, Any] = {
            "chain": chain,
            "network": network,
            "connector": connector,
            "address": address,
            "base": base_asset,
            "quote": quote_asset,
            "side": side.name,
            "amount": f"{amount:.18f}",
            "limitPrice": str(price),
            "allowedSlippage": "0/1",  # hummingbot applies slippage itself
        }
```
        
        
        **thses changes given by me in my previous pull request
       you can cahcek #6699**

HarjobandeepSingh:
@nikspz this mean my code worked ?

nikspz:
> @nikspz this mean my code worked ?

What is the commit you mentioned above?

HarjobandeepSingh:
> > @nikspz this mean my code worked ?
> 
> What is the commit you mentioned above?
#6699  @nikspz

HarjobandeepSingh:
@nikspz

hummingbot/core/gateway/gateway_http_client.py
```diff
@@ -492,7 +492,7 @@ async def amm_trade(
             "quote": quote_asset,
             "side": side.name,
             "amount": f"{amount:.18f}",
-            "limitPrice": str(price),
+            "limitPrice": f"{price:.20f}",
```
mlguys, line 495: This change is actually right in the bounty description, maybe we should add more value to this PR by adding test case to check for this issue?

mlguys:
> Small Price Test: Tested the function with a price value less than 1e-7 to verify that the function raises a ValueError as expected.

I think you should change this since the goal of the bounty is that the function should work correctly with small prices. @HarjobandeepSingh

PR #6749, opened 2023-12-29
fix/simplify funding loop

MementoRC:

This is to try to understand what is wrong with this PR from fede

**Tips for QA testing**:


cardosofede:
@MementoRC thanks for this work memento!
I didn't test asyncio.condition, it's like a flag? We can keep using this mechanism but we can remove the success flag since we can collect it later and if not we are going to block the bot with this process.
Btw, @carlitogetaladajr @fengtality I asked Memento to do the investigation around this issue and I think that might be considered a P2 bounty.

rapcmia (approved):
LGTM
- Funding payment for the following connectors has been monitored and observed ok
   - Binance perpetual
   - Gateio perpetual
   - Kucoin perpetual
   - dYdX perpetual

MementoRC:
@cardosofede The Condition adds a lock to the Event. It could be used to prevent the apparent race condition that occurred.
The other PR on using a timed-looping is also updated and seem to work, but an odd error is preventing the checks to pass.
It is not clear which method is more robust to me, so I simply updated both

cardosofede (approved):

PR #6869, opened 2024-02-24
Fixed ImportError due to circular dependency

losingle:
- Removed the problematic import statement from `hummingbot/core/utils/estimate_fee.py` that was importing `TradeFeeSchemaLoader` from `hummingbot.client.config.trade_fee_schema_loader`.
- This change resolves the `ImportError` that occurred due to a circular import when initializing Hummingbot.
- Code in `estimate_fee.py` was refactored to avoid relying on the direct import, ensuring proper module initialization without import loops.

With this fix, the codebase should no longer encounter the ImportError during runtime, thus maintaining the smooth operation of Hummingbot's initialization process.

### DIFF
--- a/hummingbot/core/utils/estimate_fee.py
+++ b/hummingbot/core/utils/estimate_fee.py
@@ -1,14 +1,9 @@
+import warnings
 from decimal import Decimal
 from typing import List, Optional
-import warnings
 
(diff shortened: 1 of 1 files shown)


losingle (commented):
Resolved ImportError due to circular import in estimate_fee.py

PR #6871, opened 2024-02-25
(fix) The MEXC Content-type in the header of API calls must be set to  application/json.

icp123:
The MEXC API Content-type in the header of API calls must be set to application/json.
MEXC will intercept illegal requests (Content-type: the network file type transmitted between the client and the server).
The above optimization measures will be implemented on February 22, 2024, at 08:00 (UTC).

see: https://www.mexc.com/support/articles/17827791513387


philipp1992 (approved):
looks good and tested on development image

PR #6976, opened 2024-04-20
fix/strategy base on directional_strategy_base uses status command will occured error "position_executer has no attritube 'position_config' "

cCurious:


cardosofede (approved):
LGTM

PR #6992, opened 2024-04-26
(fix) naming  get_connector_spec fix

isreallee82:

while working on https://github.com/hummingbot/gateway/pull/316 noticed
curl tests work fine with balance and allowances
but  I get error from the client 

ValueError: not enough values to unpack (expected 2, got 1)

Screenshot 2024-04-26 at 1 51 22 PM

Due to naming convention client uses avalanche i.e Network as chain, making the split process not happen leading to a bug

Screenshot 2024-04-26 at 1 58 05 PM

**Tips for QA testing**:


fengtality:
From the code in this PR, you are just changing the name of the locally scoped `chain` variable to `chain_name`. That should have any impact outside of naming and I don't see how it can fix an error like `ValueError: not enough values to unpack (expected 2, got 1)` where it's related to arguments not being there or something like that.

isreallee82:
@fengtality @nikspz  after alitlle digging the iteration must have stopped on avalanche from SUPPORTED_CHAINS set, so its final result must have been Uniswap_avalanche_avalanche,

so I figured the reason why chain_name worked instead of chain was there was changes to arrangements in the set SUPPORTED_CHAINS which placed ethereum first on the iteration list

For chain
Screenshot 2024-05-04 at 1 34 28 PM

For chain_name
Screenshot 2024-05-04 at 1 36 31 PM

Either ways both solutions were buggy 


I  figured another solution on my latest commit
since connectors and supported chains have no underscore I came up with this solution.
I am opened to corrections @fengtality

isreallee82:
I think this approach is more simple and solves the problem @nikspz

fengtality (approved):

fengtality:
> I think this approach is more simple and solves the problem @nikspz

Agreed!

nikspz (approved):
Test performed:
- Cloned and installed PR6992 and gateway latest development
- connected gateway uniswap_ethereum_arbitrum, pangolin_avalanche_avalanche, uniswapLP_ethereum_arbitrum, sushiswap_ethereum_mainnet, sushiswap_avalanche_avalanche, quickswap_polygon_mainnet
- checked balance successfully

PR #7040, opened 2024-05-21
Feat/ add feature to specify quote asset as order amount in pmm strategy

zeeshanakram3:


cardosofede:
@zeeshanakram3 hey thanks for the contribution. I approved the workflow but as it's a change in a core strategy you need to go through the governance process. Check the docs under the contributions section to understand it.
I think that the test might fail since you are adding new functionality without coverage, and also we are trying to move away from the v1 strategies and the v2 strategies are all using quote asset, so you might want to check that if that's what you are using.

PR #7060, opened 2024-05-29
Adding DEX connector to Sui's DeepBook 

gordonkoehn:
Adding a connector to Sui's on-chain order book DeepBook using [PySui ](https://github.com/FrankC01/pysui).


gordonkoehn:
@cardosofede 
Thanks for the notice. 

This is a discontinued PR, a remnant from a Hackathon project we undertook at Consensus 2024 this summer.

Just starting out with Hummingbot then, we didn't finish it in time. Thus, I mark this as a Draft PR and have changed the base as you suggested.

Hope to continue this another time.

PR #7084, opened 2024-06-20
Added trailing stop for Position Executor

peterwilli:

This allows trailing stop to work with the dashboard. You need this PR combined with the Dashboard PR (https://github.com/hummingbot/dashboard/pull/149)!

Both PRs need to be accepted in order for this to work, and hummingbot-deploy's Backend API needs to be compiled with the hummingbot project with this PR. 

See screenshot:
image

Tested the backtest with trailing stop, and ran `make test`, all is fine over here


peterwilli:
Fixed a bug with a negative PNL and also added an epsilon to compare the actual stoploss v.s the trailing stop accurately!

image

peterwilli:
image

Stoploss now stays at a max of it's previous value, reflecting irl scenarios. I use vectorized operations and avoid loops for this so performance isn't affected. The code is also made more simple and concise.

cardosofede:
@peterwilli will review this tomorrow and we need to coordinate the inclusion of it with the new release including the refactored candles feed

cardosofede:
hey @peterwilli can you also add this to the DCA Simulator?

peterwilli:
Hey lovely @cardosofede , yeah I can, but I never used DCA simulator before. I cannot assert it's correctness as well as with PositionExecutor, because I can actually see what it's doing there in live trades

cardosofede:
@peterwilli okey don't worry for now, the main difference between the PositionExecutor and the DCAExecutor is that the DCA has multiple open orders and just one close order, and all the things like Take profit, stop loss and trailing stop are calculated from the average price executed at that moment.

hummingbot/strategy_v2/backtesting/executors_simulator/position_executor_simulator.py
```diff
@@ -30,6 +38,7 @@ def simulate(self, df: pd.DataFrame, config: PositionExecutorConfig, trade_cost:
         df_filtered['cum_fees_quote'] = 0.0
         df_filtered['filled_amount_quote'] = 0.0
         df_filtered["current_position_average_price"] = float(config.entry_price)
+        df_filtered['sl'] = float(sl)
```
cardosofede, line 41: to avoid confusion with stop loss that usually is sl, I recommend calling it ts (trailing stop)

hummingbot/strategy_v2/backtesting/executors_simulator/position_executor_simulator.py
```diff
@@ -44,16 +53,25 @@ def simulate(self, df: pd.DataFrame, config: PositionExecutorConfig, trade_cost:
         df_filtered['net_pnl_quote'] = df_filtered['net_pnl_pct'] * df_filtered['filled_amount_quote']
         df_filtered['cum_fees_quote'] = trade_cost * df_filtered['filled_amount_quote']
 
+        # Make sure the stop loss rises with the trailing stop trigger price (if any)
```
cardosofede, line 56: here the same, is better if we keep them separate as different things. The trailing stop is more like a dynamic take profit, and you can just run the bot without stop loss and with trailing stop

hummingbot/strategy_v2/backtesting/executors_simulator/position_executor_simulator.py
```diff
@@ -44,16 +53,25 @@ def simulate(self, df: pd.DataFrame, config: PositionExecutorConfig, trade_cost:
...
+            df_filtered.loc[df_filtered['net_pnl_pct'] > trailing_sl_trigger_pct, 'sl'] = (
+                df_filtered['net_pnl_pct'] - float(trailing_sl_delta_pct)
+            ).cummax()
+
         # Determine the earliest close event
         first_tp_timestamp = df_filtered[df_filtered['net_pnl_pct'] > tp]['timestamp'].min() if tp else None
-        first_sl_timestamp = df_filtered[df_filtered['net_pnl_pct'] < -sl]['timestamp'].min() if sl else None
-        close_timestamp = min([timestamp for timestamp in [first_tp_timestamp, first_sl_timestamp, tl_timestamp] if not pd.isna(timestamp)])
+        first_sl_timestamp = df_filtered[(df_filtered['net_pnl_pct'] < -df_filtered['sl']) & (abs(df_filtered['sl'] - float(sl)) <= epsilon)]['timestamp'].min() if sl else None
+        first_trailing_sl_timestamp = df_filtered[(df_filtered['net_pnl_pct'] < df_filtered['sl']) & (abs(df_filtered['sl'] - float(sl)) > epsilon)]['timestamp'].min() if sl else None
```
cardosofede, line 65: the overall logic seems right

cardosofede (changes requested):
Thanks emerald for the contribution!!
I added some comments, seems like the logic is okey but this simplification that is isolate the ts, sl and tp will make it more clear.
I liked the cummax and the evaluation of the net pnl with that value ;)

peterwilli:
Hey lovely @cardosofede ! Sorry for late reply 

I understand what you're saying. But in live trading isn't it so that the stop loss rises if the price moves in your advantage? That is to say, i dont think that a trailing stop is like a dynamic take profit. A take profit is triggered once the price moves in your advantage past a certain threshold.

A stop loss is when the price moves opposite direction past a threshold. You could close an order now and it will always be above your stop loss which is the case for a take profit (unless I have missed something in the way hummingbot live trading does it)

I think it's possible to separate it, though. I'll check this when back home

cardosofede:
@peterwilli heyyyy, yes in the mechanics, if the trailing stop gets activated the stop loss will not take effect anymore but there are some cases where makes sense to treat it as a separated thing:
- not having stop loss, just trailing stop and time limit
- DCA, the stop loss only works when all the orders are filled and the trailing stop works from the first order only

peterwilli:
@cardosofede thanks for the explanation, I updated the PR and it now contains your feedback + I updated to the latest development branch!

I tested out the code and got identical results with this change v.s the old one, so behaviour remains the same

cardosofede:
The code looks good to me, I will create the environment in the backend api to debug it since we don't have unit test yet for this

nikspz:
failed to trigger trailing stop, 

Steps to reproduce:
1. Cloned PR7084, install and manually built docker image as hummingbot/dashboard:latest
2. Cloned dashboardPR149, manually built docker image as hummingbot/hummingbot:latest
3. pulled latest hummingbot/backend-api:latest
4. cloned fresh deploy repo and removed pull lines
5. bash setup.sh
6. Set up strategy to easily get trailing stop (low spread and high for others)

Actual:
Reviewed trailing stop is not triggered, however take profit could be triggered successfully


take profit 10
image
take profit 10 + time limit 14400
image


bollinger:
bollinger-trailingstop2
pmm_simple:
Pmmsimple-no-trailingstop22

cardosofede:
I found an issue with the code:
The configuration is the following:
image

I debugged the code with a breakpoint here:
image

As you can see the TS is activated correctly when the net pnl pct > 1%, but after the condition is met the ts should appear for all the observations and it has some gaps.
image

The next ts is correct, but only because we dismissed the previous nans
image

The issue can be fixed here, adding either a fill of nans with the previous values or changing the filter to index the dataframe from the first value that is greater than the ts:
<img width="834" alt="image" src="https://github.com/hummingbot/hummingbot/assets/36869960/c6b04c07-23cf-47ea-bba1-c4c97e692ca9"

peterwilli:
@cardosofede Thanks for the feedback, you were right, there was an issue that if pnl drops again, the trailing stop fails to trigger. I took your second advice to fix it, as it seems cleaner to make the mask work right away rather than filling it with NaNs afterwards...

I tested it out, in my results all output is now triggered from the point the first trailing stop triggers.

peterwilli:
> Great job @peterwilli

Awh thaaaaanks!! It was lovely to work with you hehe

PR #7117, opened 2024-07-08
fix: broken link in `install` script

intelliot:
The address led to a 404 Page Not Found. This fixes the URL.


hummingbot/VERSION
```diff
@@ -1 +1 @@
-dev-1.29.0
+2.0.0
```
cardosofede, line 1: @rapcmia we need to update the dev version right?

install
```diff
@@ -12,7 +12,7 @@ fi
 
 if [ "${CONDA_EXE}_" == "_" ]; then
     echo "Please install Anaconda w/ Python 3.7+ first"
-    echo "See: https://www.anaconda.com/distribution/"
+    echo "See: https://www.anaconda.com/download"
```
cardosofede, line 15: the changes are okey to me

setup.py
```diff
@@ -33,7 +33,7 @@ def build_extensions(self):
 
 def main():
     cpu_count = os.cpu_count() or 8
-    version = "20240531"
+    version = "20240703"
```
cardosofede, line 36: @rapcmia same here

cardosofede (changes requested):
The PR is okey to merge but we need to make sure that the versions are updated.
@intelliot here are the contribution guidelines: https://hummingbot.org/developers/contributions/
I changed the base branch to development since all the contributions passes through it

hummingbot/VERSION
nikspz: Yes should be dev-2.0.1 for now
https://github.com/hummingbot/hummingbot/blob/development/hummingbot/VERSION

nikspz (approved):
Checked links, LGTM

PR #7310, opened 2024-11-24
Fix / Binance WSS reconnection fix

komposter:
Binance requires a specific response to each **ping** message, otherwise it stops streaming.

Hummingbot reconnects correctly, but it takes up to **5 seconds** each time. If you want to build a strategy that is sensitive to these disconnects, you may want to fix the **pong** message so that Binance continues streaming.

This fix affects the WSConnection base class, so it should be tested on other exchanges (I don't have other accounts right now, so I can't do that).


cardosofede:
@komposter this is something that I was going to work on this week actually!
My question is, your PR tries to solve the problem of Binance but you are modifying the core ws assistant, this applies to all connectors?

komposter:
> My question is, your PR tries to solve the problem of Binance but you are modifying the core ws assistant, this applies to all connectors?

I don't think so, I only tested it with binance and binance_perpetual.
But I couldn't find any specific implementations of `WSConnection` in the code, only the main class which is used by all exchanges.

Do you have any ideas how to inherit from it and use it in a modified form for some specific exchanges?

cardosofede:
@komposter nope, my point is that this fix should be working for all the connectors not only for Binance, and if Binance requires an specific message, we should implement it in the connector when we are processing the messages

komposter:
> @komposter nope, my point is that this fix should be working for all the connectors not only for Binance, and if Binance requires an specific message, we should implement it in the connector when we are processing the messages

I understand that it is supposed to work with other exchanges. However, as I said, I can't check this because I don't have any other accounts at the moment.

I don't know the hummingbot architecture well enough to implement this behavior only in a specific exchange. However, if you give me some hints, I might give it a try.

komposter:
> btw, check that you need to implement the tests too

I see it fails on `test_receive_sends_pong_on_ping`
Do you mean I need to fix this? Or do you want me to write more tests for the specific case I tried to fix?

komposter:
@nikspz 
Thanks for your tests!
Fixed that error in the last commit.

nikspz:
commit 5c24757 

Test performed:
- Clone and install PR7310
- Run tests using binance_perpetual (macd_bb_v1 controller) for 22 hours
- Run binance (pmm_simple controller) for 10 hours
- Review no WARNING - The websocket connection was closed (The WS connection was closed unexpectedly. Close code = 1008 msg data: 1008)
- Review no (reconnection event) INFO - Subscribed to public order book and trade channels after start / INFO - Successfully obtained listen key

cardosofede (approved):
great contribution @komposter

PR #7311, opened 2024-11-24
Fix / position_executor for spot accounts

komposter:
For spot trading (I trade on Binance), the commission is deducted from the balance of the base asset. For example, if you buy 60864 SHIB-USDT, binance charges a transaction fee of 60.86 SHIB. So the SHIB balance after “opening” the position would be 60864-60.86 = 60803.14.
But when it comes to closing (or setting a limit-TP order), hummingbot tries to send a sell order with the initial amount (60864 SHIB) and fails due to the error “Account has insufficient balance for requested action.”.

With the fixes I made, it subtracts the commission from the initial position size, so the close order (either a market at time_limit or a limit-TP order) is sent and executed correctly.

I also reverted the `renew_take_profit_order` call if the TP limit-order has an incorrect amount. I don't know why it was removed (and renew_take_profit_order is never called after these changes).


cardosofede:
@rapcmia let's review this one for spot!

rapcmia:
PR update:
- Run bots with v2 controller-config on both this PR and development branch to compare results 
- Confirmed changes on amount now reflected on orders used to close executors  (market or limit) ✅ 
- Confirmed different close types worked event with partial fills 
  - stop_loss, early_stop, time_limit, take_profit ✅ 
- Run pmm_simple controller on binance 3/3 bots reviewed

---
Todo:
- Found an instance of closeType early_stop with 0 values not recorded on logs but on executor table (sqlite) and exchange trade history `investigating with dev branch`
- Need to run another connector to test using different controller-config `not started`

komposter:
I noticed that it failed the code check: `E226 missing whitespace around arithmetic operator`.
Could you please run your code formatter so that we don't have to go through the hassle of new fixes and checks?

rapcmia (approved):
- Setup on source ok 
- Run tests using v2 simple pmm controller
- Run tests on both PR and development branch to compare results 
- Run tests using Binance and OKX 
- Review logs, Trade data (CSV and exchange history)
- Review executor open/close orders, also with respective closeTypes ok 
	- Open: Default amount 
	- Close: Default amount - trade_fees

komposter:
@cardosofede 

Do you expect me to fix formatting?
Don't you have a preconfigured code formatter that you can just run over the patched file?

cardosofede:
@komposter when you run the command ./install pre commit is installed in your environment and should not let you commit until the problems are resolved. all the contributions are made in the same way, if you want to add a command in the Makefile to format the code feel free to contribute it too.

komposter:
@cardosofede So how did I commit this code?
I don't understand what I need to do. I followed the contribution guide, but it seems like there's something missing.

cardosofede:
@komposter 
The error says that there is a missing whitespace in the line 105. If you add that whitespace that part of the workflow should pass.
image

komposter:
@cardosofede Is this rejected?

rapcmia:
> @cardosofede Is this rejected?

Hi @komposter the changes from this PR has been added to https://github.com/hummingbot/hummingbot/pull/7316/files#diff-05075a82519ee5d21dc0a05ccfd3b25143214f27b10b0a796bfa7fd97d7c8366

PR #7321, opened 2024-12-01
fix/hyperliquid_spot_backend_error

yancong001:

fix the issue similar to  https://github.com/hummingbot/backend-api/issues/43

**Tips for QA testing**:

### DIFF
--- a/hummingbot/connector/exchange/hyperliquid/hyperliquid_auth.py
+++ b/hummingbot/connector/exchange/hyperliquid/hyperliquid_auth.py
@@ -46,7 +46,8 @@ class HyperliquidAuth(AuthBase):
     def construct_phantom_agent(self, hash, is_mainnet):
         return {"source": "a" if is_mainnet else "b", "connectionId": hash}
 
-    def sign_l1_action(self, wallet, action, active_pool, nonce, is_mainnet):
+    def sign_l1_action(self, action, active_pool, nonce, is_mainnet):
+        wallet = eth_account.Account.from_key(self._api_secret)
         _hash = self.action_hash(action, active_pool, nonce)
         phantom_agent = self.construct_phantom_agent(_hash, is_mainnet)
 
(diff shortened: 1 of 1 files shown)


isreallee82:
This is not a bug. The key difference between the two code snippets lies in how the wallet is handled and the choices regarding reusability, scoping, and dependencies, which are more of a refactor...

cardosofede:
@isreallee82 is a bug since you can't create an instance of the connector without the keys.
Actually the right solution should be if self._trading_required .... -> create wallet

isreallee82:
hmm I see, it should be called only when trying to sign a transaction

cardosofede:
@isreallee82 mmm, that's a solution that was implemented in this PR but I think that the original solution of doing the initialization in the __init__ was the right choice so you don't need to re-create that object every time that you sign. on the other hand, it should be only initialized if trading is required. if you want to take over and do it feel free to contribute, I think that this PR is not related to any bounty it is a contribution and you are the creator of this connector ;)

isreallee82:
> @isreallee82 mmm, that's a solution that was implemented in this PR but I think that the original solution of doing the initialization in the **init** was the right choice so you don't need to re-create that object every time that you sign. on the other hand, it should be only initialized if trading is required. if you want to take over and do it feel free to contribute, I think that this PR is not related to any bounty it is a contribution and you are the creator of this connector ;)

Oh, it's fine. I'm open to any potential improvement. since this change is aimed at solving an error on the backend-api. I was only confused when the commit was described as a bug.

cardosofede:
@isreallee82 

this is the error anytime that I want to initialize the connector without keys to use the public funcitonalities
image

isreallee82:
oh i see @cardosofede

isreallee82:
tried the trading_required method @cardosofede and it works fine

isreallee82:
It’s so fun coding on hummingbot 🤩, just a little tweak i will run test and create pr.

isreallee82:
pr #13936

PR #7357, opened 2024-12-27
fix/get_historical_candles

socialsmoker223:

When using CandlesBase to fetch historical candles over a wide time range, the process could enter an infinite loop if the specified range includes unavailable candles. To address this, I refactored the get_historical_candles method to ensure that current_end_time progresses correctly and prevents such infinite loops.

Validated historical candle fetching over a wide range of timestamps to confirm the loop processes all expected intervals

**Tips for QA testing**:


cardosofede:
this might be interesting and affects the candles module, what do you think @tomasgaudino ?

Issue #7361, opened 2024-12-29
Unexpected error while fetching bybit trading fees


snde:
### Describe the bug

I am executing arbitrage between Bybit and Jupiter with the trading pair PENGU-USDT. The exchange is reporting that it cannot fetch the trading fee, and the trading pair also seems incorrect. The screenshot includes my configuration and logs.
![1735438881116](https://github.com/user-attachments/assets/e48a06c6-c17b-4711-a792-edbea4422cae)
![1735438856390](https://github.com/user-attachments/assets/e00cfe79-1165-4943-a9e6-5a851a70a4a8)

### Steps to reproduce

1. cex bybit
2. dex jupiter
3. pengu-usdt 

### Release version

v2.2.0

### Type of installation

Docker

nikspz:
cc: https://discord.com/channels/530578568154054663/563462222878212098/1326494722939818054
```
2025-01-06 16:20:53,342 - 15 - hummingbot.connector.exchange.bybit.bybit_exchange.BybitExchange - NETWORK - Unexpected error while fetching trading fees.
Traceback (most recent call last):
  File "/home/hummingbot/hummingbot/connector/exchange_py_base.py", line 752, in _trading_fees_polling_loop
    await safe_gather(self._update_trading_fees())
  File "/home/hummingbot/hummingbot/core/utils/async_utils.py", line 22, in safe_gather
    return await asyncio.gather(*args, **kwargs)
  File "/home/hummingbot/hummingbot/connector/exchange/bybit/bybit_exchange.py", line 309, in _update_trading_fees
    trading_pair = await self.trading_pair_associated_to_exchange_symbol(symbol=tpfee["symbol"])
  File "hummingbot/connector/exchange_base.pyx", line 108, in trading_pair_associated_to_exchange_symbol
    return symbol_map[symbol]
  File "/opt/conda/envs/hummingbot/lib/python3.10/site-packages/bidict/_base.py", line 524, in getitem
    return self._fwdm[key]
KeyError: 'CTANUSDT'
2025-01-06 16:20:53,342 - 15 - hummingbot.connector.exchange.bybit.bybit_exchange.BybitExchange - WARNING - Could not fetch new trading fees from Bybit. Check network connection.
[17:16]
Like this
[17:16]
I am trading XRPUSDT but the error shows CTANUSDT
```

rapcmia:
Fix has been merged to development and will be part of release version 2.3.0

PR #7373, opened 2025-01-08
fix/bybit-update-trade-fees

leocape:

Quick fix for the ByBit connector trade fees error:

hummingbot.connector.exchange.bybit.bybit_exchange.BybitExchange - NETWORK - Unexpected error while fetching trading fees. 

Essentially the issue is that the resulting trading pair map only contains the trade-able pairs (which is correct) but the trading fees payload contains all pairs, not just the trade-able pairs. So when the method tries to map the rate pair from the trading pairs, there is no matching pair to map from.

eg CTANUSDT is not trade-able, so it is not in the mapped trading pairs, but it is present in the trade fee payload.

We continue on key error during trading fee to trading pair map

All

**Tips for QA testing**:
Connect to bybit, start a PMM strategy, no funds required, the strategy should at least start without getting blocked by the previous fee rate pair error.

### DIFF
--- a/hummingbot/connector/exchange/bybit/bybit_exchange.py
+++ b/hummingbot/connector/exchange/bybit/bybit_exchange.py
@@ -306,8 +306,12 @@ class BybitExchange(ExchangePyBase):
         # await self._update_exchange_fee_rates()
         fee_rates = await self._get_exchange_fee_rates()
         for tpfee in fee_rates:
-            trading_pair = await self.trading_pair_associated_to_exchange_symbol(symbol=tpfee["symbol"])
-            self._trading_fees[trading_pair] = tpfee
+            try:
+                trading_pair = await self.trading_pair_associated_to_exchange_symbol(symbol=tpfee["symbol"])
+                self._trading_fees[trading_pair] = tpfee
+            except KeyError:
+                # Skip pairs that are not trade enabled ie. they are not present in the trading pair map
+                continue
 
     def _process_trade_event_message(self, trade_msg: Dict[str, Any]):
         """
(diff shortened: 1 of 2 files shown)


nikspz:
Thanks @leocape Please review add unit test coverage here
https://github.com/hummingbot/hummingbot/actions/runs/12670547528/job/35310497590?pr=7373

https://github.com/leocape/hummingbot-base/blob/fix/bybit-update-trade-fees/test/hummingbot/connector/exchange/bybit/test_bybit_exchange.py

nikspz:
Test performed:
- Cloned and installed PR7373
- connected bybit successfully
- set up tight spreads for filled orders events
- review no error while fetching fees
- reviewed fees recorded in history / sqlite matched with exchange


Crosschecked issue https://github.com/hummingbot/hummingbot/issues/7361 on development:

image

Using this PR7373:
image

nikspz (approved):

cardosofede (approved):
Hi @leocape thanks for the contribution. I think that what you are doing here will reduce the amount of information and what you can do to consider this symbol is in the except, generate the base-quote and still store the fees. I don't know if you try it but maybe it's a solution. either way as the token is not tradable we can just accept this solution.
let me know if you try it but for now is approved

cardosofede:
@leocape hey Leo, you can use the async method to get the pair based on the exchange pair, check the connector that is really common that. When the connector gets initialized generates automatically this map.

cardosofede:
@leocape yes, but what you can do is to use the same logic to build the trading pair name, it's strange that you pay fees in a not tradable pair. but well I think that the initial try except considere a lot of cases since I think that we will not trade in those pairs too much.
The coolest solution will be having some way of storing a list to make the mapping

PR #7427, opened 2025-02-08
FEAT: add executor factory

MementoRC:
**Before submitting this PR, please make sure**:

- [x] Your code builds clean without any errors or warnings
- [x] You are using approved title ("feat/", "fix/", "docs/", "refactor/")

**A description of the changes proposed in the pull request**:
The Executor factory allows to decouple the ExecutorOrchestrator from the specific implementation of Executors. Each Executor/Config registers itself into the factory

**Tests performed by the developer**:
Testsuite

**Tips for QA testing**:

PR #7437, opened 2025-02-12
FEAT: Improve Executors abstraction with Factory

MementoRC:

The Executor factory allows to decouple the ExecutorOrchestrator from the specific implementation of Executors. Each Executor/Config registers itself into the factory

Continues/Replaces: https://github.com/hummingbot/hummingbot/pull/7427 - to conserve inline docs and pytest in local branch

WIP

**Tips for QA testing**:


MementoRC:
@cardosofede This is the solution you should have looked at, instead of the mapping

cardosofede:
the solution seems more robust but at the same time much more complex than just adding the mapping that is really simple and easy for anyone else to understand, what benefits this solution provides that we should take into account to implement it? sometimes I think that simple is better than complicated and if the two of them are solving the same problem and one is much more simpler I will pick that one. btw I like the update feature we should do that, and I also liked the decorator ;) but in this case I prefer the simpler one

MementoRC:
Well, this solution provide a complete separation of concerns. It could potentially resolve circular imports. devs can add executors without touch the orchestrator and potentially add mistakes. Users can have their own executors without having to introduce a merge conflict with the development branch.

PR #7470, opened 2025-03-02
(fix) adds ms to trade updates for gate.io spot and perpetual connectors

komposter:
I realized that trade updates from the gate.io exchange were coming without **milliseconds**.
The reason was using the “create_time” field instead of “**create_time_ms**”.

I checked trade streams for both spot and perpetual connectors, all trades now contain ms in their timestamps.

**Tips for Dev team**:
Other updates, such as [trade fills](https://www.gate.io/docs/developers/futures/ws/en/#orders-notification), are also missing milliseconds, even though the API also provides fields with ms.

I didn't fix all the streams and API calls because I don't trade on this exchange yet and can't test everything, but it doesn't seem hard to do if someone needs it.


hummingbot/connector/derivative/gate_io_perpetual/gate_io_perpetual_api_order_book_data_source.py
```diff
@@ -93,7 +93,7 @@ async def _request_order_book_snapshot(self, trading_pair: str) -> Dict[str, Any
 
     async def _parse_trade_message(self, raw_message: Dict[str, Any], message_queue: asyncio.Queue):
         for trade_data in raw_message["result"]:
-            trade_timestamp: int = trade_data["create_time"]
+            trade_timestamp: float = float(trade_data["create_time_ms"]) * 1e-3
```
nikspz, line 96: hi @komposter Could you explain why is it float (not int) here?
image

nikspz:
commit e307886


PR7470:
image
image

Latest development branch:

image
image

nikspz:
> You can see these 3 zeros in each timestamp. They will be replaced with meaningful values with my fix.

Yeah this is same 000 in this PR #7470 (for trades filled)

[gperp.zip](https://github.com/user-attachments/files/19109624/gperp.zip)
[7470gatewayms.zip](https://github.com/user-attachments/files/19109620/7470gatewayms.zip)

cardosofede (approved):
for me this changes are okey, pls @nikspz compare with binance using the same notebook

nikspz:
Same script on binance
image

nikspz (approved):

PR #7512, opened 2025-03-21
Implement file based persistent command history.

WarrenFerrell:

Added persistent command history to the Hummingbot Console with configurable command filtering. The changes include:

1. Added configuration options in `client_config_map.py`:
   - `command_history_file`: Enable/disable file-based history (default: False)
   - `command_history_file_path`: History file location (default: ~/.hummingbot/cmd_history/command_history.txt)
   - `command_history_exclusion_list`: List of command prefixes to exclude (default: ["exit"])

2. Modified `layout.py` to use `FileHistory` from prompt_toolkit for persistent command storage and added command filtering through a custom `FilteredFileHistory` class.

3. Updated `hummingbot_cli.py` to pass client configuration to input field creation.

- Started the application and verified command history is retained between sessions
- Confirmed that the "exit" command is not stored in command history
- Verified that other commands are properly stored and can be recalled
- Verified that setting command_history_file to false restores previous behavior

**Tips for QA testing**:

1. Start Hummingbot and enter some commands
2. Exit and restart Hummingbot
3. Use up/down arrows to verify:
   - Previous commands are available
   - "exit" command is not in history
   - History file exists at ~/.hummingbot/cmd_history/command_history.txt


WarrenFerrell:
don't merge yet. broke a test.

WarrenFerrell:
@cardosofede 
This is what the command_history.txt file looks like. It only stores the input command and only when you use a command was not already the most recent command (I tested by using balance about 10 times in a row. I haven't looked into anything about prompt_toolkit customization except for excluding exit from history (since it was super annoying to accidentally press up after reopening the app and have it immediately close. We could include stop, status and/or balance as one of the default words to exclude as well as that would greatly reduce the number of entries that actually get written but since the config is off by default I don't think that's necessary. Went ahead and changed to a config Map.
```
# 2025-03-29 18:18:36.454425
+stop

# 2025-03-29 18:18:41.079173
+balance

# 2025-04-01 12:33:13.293644
+stop
```

rapcmia:
Test `35b491d897428bf57fc2327c4d62213331f3b7a7`
- Run tests on linux (ubuntu20.04) using source build 
- New commands `command_history`
  image- Set `command_history.use_history_file` to true ✅
    - Changed default path from `~/.hummingbot/cmd_history/command_history.txt` to `~/hummingbot/cmd_history/command_history.txt`
    - When set to false, does not fill any of stored history commands  ✅
        - Then set to true, managed to pull all history commands when using up / down keys ✅
- Observed that it does not create .txt file if we add new ❗ 
  image
    - manually created test.txt and add `a+rw` permission ✅ 
    - loaded and worked ✅ 
- Confirmed all config changes are stored on `command_history.txt`
  image
  - Relaunch hummingbot after changes 
    - Confirmed that all history configs are stored and not `exit` (default) ✅ 
    - Run few more commands like balance paper ✅ 

---

### Test command_history_execlusion_list: does not accept anything we want to add ❌ 
- does not accept if command has space in between like `gateway generate-certs` 
- tried to add different types of variation separated by comma and getting same results as above. Observed that it does not consider exit as well ❌
  image

---

###  Docker build 
- Enabled and confirmed updated on `conf_client.yml`
- Does not register any of the commands on history ❌ 
  - Does not seem to recognize the path we set from `history_file_path`

WarrenFerrell:
I added a min_length constraint to prevent setting of the path to ''  . The makes it easier to set the default as pressing enter creates an invalid entry. Exclusion list needs to be able to accept an empty list so it can't have that constraint. Also specified that if using the top directory you need to use ./ . Overall not actually a bug as it's invalid config values that cause that. And The same lack of documentation occurs with log_file_path config.

WarrenFerrell:
@nikspz Sorry I thought I had commented on this. That is expected behavior. "default conf" is not a valid filepath. That said, I added an error message to clarify what went wrong

PR #7523, opened 2025-04-01
fix issue causing the rate oracle to forcibly default to BinanceRateS…

WuonParticle:

This PR fixes an issue where the rate oracle was forcibly defaulting to BinanceRateSource via rate_oracle_source_on_validated regardless of user configuration. The fix ensures that user-specified rate sources are properly respected, allowing for alternative rate sources to be used as configured. This is a significant problem for us residents as binance does not work as a rate source for them. 

- Verified that coin_gecko can be properly configured to be rate_source and persists when using arbitrage_controller

**Tips for QA testing**:

- Configure different rate sources in the config and verify they're used correctly
- Check the logs to confirm the correct rate source is being initialized
- Test with rate_oracle_source set to different values to ensure the setting is respected

### DIFF
--- a/hummingbot/client/config/config_helpers.py
+++ b/hummingbot/client/config/config_helpers.py
@@ -652,15 +652,37 @@ def load_connector_config_map_from_file(yml_path: Path) -> ClientConfigAdapter:
     return config_map
 
 
-def load_client_config_map_from_file() -> ClientConfigAdapter:
-    yml_path = CLIENT_CONFIG_PATH
+class MostRecentConfigLoadCache:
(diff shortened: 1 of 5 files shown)


WuonParticle:
I had put in draft because the initial change that used the hummingbot application interfered with backtesting causing backtests initialize a gateway client. They still worked but it was annoying. Created a cache for the most recently loaded config in my second implementation.

rapcmia:
Reviewed this against the current development branch, and the related code path appears to have changed. The proposed fix no longer matches the current implementation. This PR can be close for now, otherwise it can be reopened with new updates. Thank you

PR #7527, opened 2025-04-02
fix bug in arbitrage_controller preventing multiple arbitrage configs 

WuonParticle:

This PR fixes a bug in the arbitrage controller that was preventing multiple arbitrage controllers from functioning together correctly. Issue is that the trading pairs for a specified market were being set each time instead of extended. This worked in the case where there was never a repeated exchange but does not work if multiple configs share a connector. 

Key changes:
1. Checked if the connector is already added to markets before updating markets dict

This fix allows users to run multiple arbitrage strategies simultaneously without interference between different controller instances.

- Tested XRP_USD & ADA_USD across kraken and Coinbase simultaneously 

**Tips for QA testing**:

- Run an arbitrage strategy with multiple configurations
- Verify each controller operates independently


WuonParticle:
It could be interesting to inherit from Dict for the dictionary passed into update_markets method like follows. I didn't do this due to scope but worth considering if multi config strategies are expected to be used frequently. As the error was an easy one to make. 

```
class MarketDict(Dict[str, Set[str]]):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)

    def add_or_update(self, connector: str, trading_pairs: Set[str]) -> None:
        if connector in self:
            self[connector].update(trading_pairs)
        else:
            self[connector] = trading_pairs
```
Then the code change could have just been 
```
  for connector_pair in [self.exchange_pair_1, self.exchange_pair_2]:
                markets.add_or_update(connector_pair.connector_name, {connector_pair.trading_pair})
```

cardosofede:
> It could be interesting to inherit from Dict for the dictionary passed into update_markets method like follows. I didn't do this due to scope but worth considering if multi config strategies are expected to be used frequently. As the error was an easy one to make.
> 
> ```
> class MarketDict(Dict[str, Set[str]]):
>     def __init__(self, *args, **kwargs):
>         super().__init__(*args, **kwargs)
> 
>     def add_or_update(self, connector: str, trading_pairs: Set[str]) -> None:
>         if connector in self:
>             self[connector].update(trading_pairs)
>         else:
>             self[connector] = trading_pairs
> ```
> 
> Then the code change could have just been
> 
> ```
>   for connector_pair in [self.exchange_pair_1, self.exchange_pair_2]:
>                 markets.add_or_update(connector_pair.connector_name, {connector_pair.trading_pair})
> ```

I like a lot this approach! will make the addition of markets much more simpler

cardosofede (approved):

rapcmia (approved):
- setup generic.arbitrage_controlle using jupiter and kucoin
- setup 2 controller-configs on script ok 
- run v2 strategy and observe when arbitrage occurs ok 
- review logs and sqlite file ok

rapcmia:
Merged to development and part of release 2.5.0

PR #7549, opened 2025-04-14
add_or_update marketdict implementation

WuonParticle:

key changes:

1. Adds `GroupedSetDict` and `MarketDict` data types to `common.py` to provide a convenient way to add values to the set mapped to by a dictionary. This primary usage for Hummingbot being the addition of trading pairs for a connector via add_or_update. 

2. Adds LambdaDict data type with a default_value_factory which is used when a key is not found and a get_or_add method. Ended up not using get_or_add in this PR but left it because it is needed if the lambda reference is not available during class initialization. 

3. Implements the logical locations to utilize add_or_update. Existing usages that don't use add_or_update still work. 

- Ran affected test suite (had to modify 1 test.)
- Created unit tests for newtypes
- Verified that `GroupedSetDict` correctly handles adding and updating values
- Confirmed type checking works as expected with the new types
- Tested compatibility with Pydantic v2 schema generation
- Validated that existing code using these types and not using these types continues to work

**Tips for QA testing**:

- Test market-related operations that use `MarketDict`
- Verify that existing controllers that don't use MarketDict still work

### DIFF
--- a/controllers/generic/arbitrage_controller.py
+++ b/controllers/generic/arbitrage_controller.py
@@ -25,9 +25,7 @@ class ArbitrageControllerConfig(ControllerConfigBase):
     quote_conversion_asset: str = "USDT"
 
     def update_markets(self, markets: MarketDict) -> MarketDict:
-        for connector_pair in [self.exchange_pair_1, self.exchange_pair_2]:
-            markets.add_or_update(connector_pair.connector_name, connector_pair.trading_pair)
-        return markets
+        return [markets.add_or_update(cp.connector_name, cp.trading_pair) for cp in [self.exchange_pair_1, self.exchange_pair_2]][-1]
 
 
 class ArbitrageController(ControllerBase):
--- a/controllers/generic/arbitrage_controller.py
+++ b/controllers/generic/arbitrage_controller.py
@@ -1,9 +1,10 @@
 from decimal import Decimal
-from typing import Dict, List, Set
+from typing import List
 
 import pandas as pd
 
 from hummingbot.client.ui.interface_utils import format_df_for_printout
+from hummingbot.core.data_type.common import MarketDict
 from hummingbot.data_feed.candles_feed.data_types import CandlesConfig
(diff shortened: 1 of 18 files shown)


WarrenFerrell:
I renamed LambdaDict to LazyDict locally as I think it better represents the reason for the pattern. I should be able to get the PR updated with the change tomorrow.

cardosofede (commented):
I like the new MarketDict but I don't like the lazy dict data much, I think that using dict comprehension is more natural for python code

cardosofede:
@rapcmia let's test in the same bot two directional trading controllers in the same pair, 1 mm in another pair and one arb controller to cover all the potential controller types and the usage of this functionality

WuonParticle:
I could separate out the LazyDict code but I'd argue it's actually the more helpful change of the two as it provides a significant performance benefit especially during startup - by not creating objects that won't be needed. It also has a maintenance benefit since I've seen cases where a particular connector constructor is broken and the having all the connectors be created lazily allows ignoring of ongoing issues completely unrelated to the current usage.

rapcmia (approved):
- Setup on Linux (ubuntu24.04) on source build 
- Setup 1 script with 2 or more configs below
	- 2 directional trading controllers `same pair` (GS and bollinger_v2)
	- 1 mm (pmm_simple, generic.pmm and basic_order_example)
	- 1 arb controller (arb_controller)
- Review logs
	- Observe if controller works as expected 
	- Orders placed, canceled and filled 
- Review sqlite 
	- Trace orders on log file, all ok 
	- Compare history command and tradeFill ok 
	- Review executor table and corresponding closeTypes ok 

Test configs, sqlite and logs here: [05072025a.zip](https://github.com/user-attachments/files/20076321/05072025a.zip)

WuonParticle:
fixed using `git rebase -i -x 'pre-commit run --from-ref HEAD~ --to-ref HEAD' upstream/development`
there was a missing blank line in controller_base.py . The pre-commit rules must have changed slightly on development since it passed CI twice before failing.

PR #7569, opened 2025-04-21
Improve backtesting performance by over 40% by using datetime index f…

WuonParticle:

- [x] Tests all pass

This PR introduces significant performance improvements to the backtesting engine by optimizing timestamp-based lookups. The key changes include:

1. Replaced timestamp-based filtering with pandas datetime index lookups in `ExecutorSimulation` class, using `searchsorted` for efficient timestamp lookups (by far the most impactful change)
2. Modified `BacktestingEngineBase` to use datetime index for all timestamp operations
3. Updated both `DCAExecutorSimulator` and `PositionExecutorSimulator` to use datetime-based slicing instead of timestamp filtering

These changes have resulted in a performance improvement of over 40% in backtesting operations by:
- Eliminating expensive DataFrame filtering operations
- Leveraging pandas' optimized datetime index operations
- Reducing memory usage by avoiding creation of intermediate filtered DataFrames

- [x] Performance benchmarks comparing old and new implementation
- [x] Backtesting accuracy verification on same timeframe and interval
- [x] Testing with Position executors via simple_pmm
- [x] Verification of datetime index conversion and handling
- NOTE: the test suite is failing locally on development so was unable to easily test it without my other local changes. 

**Tips for QA testing**:

- Run backtesting with various timeframes to verify performance improvements do not change results
- Test with different trading strategies to ensure accuracy is maintained
- Verify that all timestamp-based operations work correctly
- Check memory usage during backtesting operations

### DIFF
--- a/hummingbot/strategy_v2/backtesting/backtesting_data_provider.py
+++ b/hummingbot/strategy_v2/backtesting/backtesting_data_provider.py
@@ -94,7 +94,7 @@ class BacktestingDataProvider(MarketDataProvider):
         """
         key = self._generate_candle_feed_key(config)
         existing_feed = self.candles_feeds.get(key, pd.DataFrame())
-        existing_feed = self.ensure_epoch_index(existing_feed)
+        # existing_feed = self.ensure_epoch_index(existing_feed)
 
         if not existing_feed.empty:
             existing_feed_start_time = existing_feed["timestamp"].min()
@@ -111,7 +111,9 @@ class BacktestingDataProvider(MarketDataProvider):
             start_time=self.start_time - candles_buffer,
             end_time=self.end_time,
         ))
-        self.candles_feeds[key] = self.ensure_epoch_index(candles_df)
+        # TODO: fix pandas-ta improper float index slicing to allow us to use float indexes
+        # candles_df = self.ensure_epoch_index(candles_df)
+        self.candles_feeds[key] = candles_df
         return candles_df
 
     def get_candles_df(self, connector_name: str, trading_pair: str, interval: str, max_records: int = 500):
--- a/hummingbot/strategy_v2/backtesting/backtesting_data_provider.py
+++ b/hummingbot/strategy_v2/backtesting/backtesting_data_provider.py
@@ -94,6 +94,7 @@ class BacktestingDataProvider(MarketDataProvider):
         """
         key = self._generate_candle_feed_key(config)
         existing_feed = self.candles_feeds.get(key, pd.DataFrame())
+        existing_feed = self.ensure_epoch_index(existing_feed)
 
         if not existing_feed.empty:
             existing_feed_start_time = existing_feed["timestamp"].min()
@@ -110,7 +111,7 @@ class BacktestingDataProvider(MarketDataProvider):
             start_time=self.start_time - candles_buffer,
             end_time=self.end_time,
         ))
-        self.candles_feeds[key] = candles_df
+        self.candles_feeds[key] = self.ensure_epoch_index(candles_df)
         return candles_df
 
     def get_candles_df(self, connector_name: str, trading_pair: str, interval: str, max_records: int = 500):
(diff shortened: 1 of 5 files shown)


WuonParticle:
I have some other changes that bring the performance increase to almost 60% in total but they rely on code in some of my other PRs

cardosofede (approved):
this is amazing!! thanks for the contribution, from my POV is ready to go, just to validate that the output is the same let's run a bt with this library and with the previous one @david-hummingbot and compare that we have the same results

WuonParticle:
discovered a bug. in prepare_market_data the pd.merge_asof messes up the index (I was only testing with pmm_simple where the code path was not hit). This is how I fixed it locally, I don't have a ton of experience with pandas to be sure if this is the best fix, Sonnet 3.7 seemed to think it was.

```
        # Make sure we have a datetime index for performance (most of the time it's already a datetime index)
        if not isinstance(backtesting_candles.index, pd.DatetimeIndex):
            backtesting_candles.index = pd.to_datetime(backtesting_candles.index, unit='s')

        if "features" not in self.controller.processed_data:
            backtesting_candles["reference_price"] = backtesting_candles["close_bt"]
            backtesting_candles["spread_multiplier"] = 1
            backtesting_candles["signal"] = 0
        else:
            backtesting_candles = backtesting_candles.reset_index(names='datetime_index')
            backtesting_candles = pd.merge_asof(backtesting_candles, self.controller.processed_data["features"],
                                                left_on="timestamp_bt", right_on="timestamp",
                                                direction="backward")    
            # Restore the datetime index using the preserved column
            backtesting_candles = backtesting_candles.set_index('datetime_index')
```

WuonParticle:
Is that the whole log? All I see are actually just warnings but I am on my phone. We can resolve those warnings by using copy on the df though, I've done that in several places sometimes because they're annoying.

WuonParticle:
I've identified and fixed the issue. The SettingWithCopyWarning is unrelated and was there before and remains after the fix. The issue had to do with some candle feeds having a range index. So to fix the problem, I switched over to using epoch seconds as a float everywhere in the backtesting code. I did not do this in candles_base as that could cause issues anywhere that a candles dataframe was accessed using a numeric index. If anyone is using a numeric index (outside of using iloc) on a dataframe then they could see their code not work for backtesting. The solution is to just switch to using iloc.

WuonParticle:
Is that something that doesn't happen on the current development branch?

WuonParticle:
I think that error is unrelated as it comes from backend-api and is unrelated to backtesting https://github.com/hummingbot/backend-api/blob/209d1ce200fc66dc6c28d6e3fbcd5380f9afca3a/services/accounts_service.py#L168

WuonParticle:
@david-hummingbot could you please ping me on discord @WuonParticle . My results when running backend-api locally don't match either of your screenshots. 
Current Development  Maybe

WuonParticle:
Took me a stupidly long time but I finally figured it out based on differences locally. Took me so long because the problem is actually the pandas-ta library and is related to the below helpful warning. I moved the epoch index initialization to after update_processed_data has already been called and that allows the performance increase without interfering with pandas-ta. I did not test the performance after the change so it might be slightly degraded over having the index always set. It's possible that this method is actually faster as well though since rangeIndices are extremely memory efficient (as seen in my debugging below). 

```
bots/controllers/market_making/pmm_dynamic.py:96: FutureWarning: The behavior of obj[i:j] with a float-dtype index is deprecated. In a future version, this will be treated as positional instead of label-based. For label-based slicing, use obj.loc[i:j] instead
```

I threw together a change in pmm_dynamic to show the differences in results between the int and float indices. Basically all values which should have NA have a computed value instead. 
```
macd_output_int_idx.info()
<class 'pandas.core.frame.DataFrame'>
RangeIndex: 1440 entries, 0 to 1439
Data columns (total 3 columns):
 #   Column         Non-Null Count  Dtype  
---  ------         --------------  -----  
 0   MACD_21_42_9   1399 non-null   float64
 1   MACDh_21_42_9  1391 non-null   float64
 2   MACDs_21_42_9  1391 non-null   float64
dtypes: float64(3)
memory usage: 33.9 KB


macd_output_float_idx.info()
<class 'pandas.core.frame.DataFrame'>
Index: 1440 entries, 1746230400.0 to 1746489420.0
Data columns (total 3 columns):
 #   Column         Non-Null Count  Dtype  
---  ------         --------------  -----  
 0   MACD_21_42_9   1440 non-null   float64
 1   MACDh_21_42_9  1440 non-null   float64
 2   MACDs_21_42_9  1440 non-null   float64
dtypes: float64(3)
memory usage: 77.3 KB
```

david-hummingbot (approved):

PR #7572, opened 2025-04-24
Feature / Hydration and Polkadot. Adding Hydration AMM connector and Polkadot chain

danilo-silva-funttastic:
Feature / Hydration and Polkadot. Adding Hydration AMM connector and Polkadot chain

This implementation is linked to the following NCP: [NCP-23: Hydration AMM Connector and Polkadot Chain](https://snapshot.box/#/s:hbot-ncp.eth/proposal/0x701eed6edab58f8c43d800bd1486df3d39cd3d1a89605d1ea3c3f1a0b56166e9)

Hummingbot Gateay PR: https://github.com/hummingbot/gateway/pull/441
Hummingbot Site documentation PR: https://github.com/hummingbot/hummingbot-site/pull/442


cardosofede:
@fengtality for me the changes are ok but they are changing a lot the gateway http client, idk if it's satisfying your new standards

Danilo-Araujo-Silva:
Thanks @cardosofede for your feedback ;)

About the changes on the `gateway_http_client.py` we tried to touch it as less as possible so this wouldn't cause any friction. We have only added the AMM routes (before the file was only targeting the CLMM ones) like `amm_pool_info`, `amm_quote_swap`, `amm_execute_swap`, and so on. The only method that needed a direct update was the `get_tokens` one, since it was not enabling the possibility to fetch just some specific tokens.

By the way guys, I'd like to suggest for us to avoid to add hardcoded support for the chains on the client like we can see [here](https://github.com/hummingbot/hummingbot/blob/249d827fe99c21a51c2753fb8b24c32b073944e2/hummingbot/client/settings.py#L108). And also [these 4 methods](https://github.com/hummingbot/hummingbot/blob/249d827fe99c21a51c2753fb8b24c32b073944e2/hummingbot/connector/gateway/gateway_base.py#L475) as well. This has recently broken the Q&A test (after I rebased the branch), but if this is all standardized on the gateway side, this would not be needed.

/cc @fengtality

fengtality:
hold off until v2.7 is done as there are more chain related changes coming.

PR #7575, opened 2025-04-26
New Strategy: Delta-Neutral Raydium LP Hedging with Centralized Exchange (CEX) Short

papapana:
Goal

Implement a delta-neutral liquidity mining strategy by providing liquidity on a Raydium full-range pool (e.g., SOL-USDC) while simultaneously shorting the volatile asset (e.g., SOL) on a centralized exchange (e.g., OKX).

---

Hypothesis

By matching the LP exposure with a corresponding short position, we can neutralize price risk and isolate the liquidity pool’s yield as pure profit.

---

Description

Liquidity pools often offer attractive yields but expose users to asset volatility.
This strategy dynamically tracks LP exposure and maintains a matching short position to preserve the portfolio’s dollar value while harvesting the pool’s yield.

The logic is split across two modules:
	- lp_hedging.py: strategy controller, rebalancing logic, exposure tracking.
	- lp_management.py: handles interactions with the liquidity pool, e.g., querying balances.

### DIFF
--- a/controllers/directional_trading/lp_hedging.py
+++ b/controllers/directional_trading/lp_hedging.py
@@ -48,7 +48,8 @@ class LpHedgingControllerConfig(DirectionalTradingControllerConfigBase):
     pool_address: str = Field(
         default="",
         json_schema_extra={
-            "prompt": "Enter the pool address (e.g. 58oQChx4yWmvKdwLLZzBi4ChoCc2fqCUWBkwMihLYQo2): ",
+            "prompt": "Enter the pool address "
+            "(e.g. 58oQChx4yWmvKdwLLZzBi4ChoCc2fqCUWBkwMihLYQo2): ",
             "prompt_on_new": True,
         },
     )
@@ -274,6 +275,8 @@ class LpHedgingController(DirectionalTradingControllerBase):
         self.rebalance_type = RebalanceType.NOT_NEEDED
         self.last_opening_attempt = 0
         self.profitability = Decimal(0)
+        self.cumulative_fees = Decimal(0)
+        self.num_short_adjustments = 0
 
     async def get_pool_info(self):
         self.gw_status = await check_gateway_status(
@@ -375,6 +378,7 @@ class LpHedgingController(DirectionalTradingControllerBase):
         create_actions = []
         signal: int = self.processed_data["signal"]
         if signal != 0 and self.can_create_executor(signal):  # type: ignore
+            self.num_short_adjustments += 1
             price = Decimal(
                 self.market_data_provider.get_price_by_type(  # type: ignore
                     self.config.connector_name,
@@ -450,6 +454,9 @@ class LpHedgingController(DirectionalTradingControllerBase):
             self.hedging_amount = Decimal(0)
             self.current_hedging_usd = Decimal(0)
             self.rebalance_type = RebalanceType.NOT_NEEDED
+            self.position_summary = self.get_position_summary()
+            if self.position_summary:
+                self.cumulative_fees = self.position_summary.cum_fees_quote
             if self.config.use_current_short_size:
                 active_orders = self.market_data_provider.get_connector(
                     self.config.connector_name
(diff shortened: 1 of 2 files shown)


controllers/directional_trading/lp_hedging.py
```diff
@@ -0,0 +1,819 @@
...
+    rate_over_period = (current_value / initial_value) - Decimal(1)
+
+    # APR: annualize the simple rate
+    apr = (
+        rate_over_period
+        * 365
+        / (Decimal(time_period_sec) / (Decimal(24) * Decimal(3600)))
+    ) * Decimal(100)
+
+    # APY: continuous compounding: exp(apr_fraction) - 1, then percentage
```
fengtality, line 222: I recommend using a simple, non-compounded approach for APR since it's difficult to predict the compounding frequency. Also, CLMM LP position APRs tend to be high anyway, so continuous compounding might overinflate expectations.

controllers/directional_trading/lp_management.py
```diff
@@ -0,0 +1,562 @@
...
+
+
+@dataclass
+class PoolInfo:
+    base_token: str | None
+    quote_token: str | None
+    last_price: Decimal
+
+
+async def check_gateway_status(
```
fengtality, line 36: I believe this fn is no longer needed if you are using the GatewayLP class since it inherits from GatewayBase which contains a Gateway status check.

controllers/directional_trading/lp_hedging.py
```diff
@@ -0,0 +1,819 @@
...
+            )
+
+            self.calculated_margin = (
+                abs(self.current_hedging_usd) / self.config.leverage
+            )
+            self.total_portfolio_value = (
+                self.total_value_usd
+                + self.hedging_pnl
+                + self.calculated_margin
+                - self.cumulative_fees
```
fengtality, line 526: I suspect fees are already reflected in `total_value_usd` if this calculation already uses the token amounts in the pools

fengtality: Also, I wasn't sure about the computation of `self.hedging_pnl` and `self.calculated_margin` here. Combined, this should be the P&L of the short position.

fengtality (commented):
Left some comments!

sdyanglei:
I’m encountering this issue. Do I need to configure “pool_info” somewhere else as well? Thank you!
image

papapana:
Thank you for looking into it! In my version, there is `pool_info` in the GatewayHttpClient which is 
```
    async def pool_info(
        self,
        connector: str,
        network: str,
        pool_address: Optional[str] = None,
        base_token: Optional[str] = None,
        quote_token: Optional[str] = None,
        fail_silently: bool = False,
    ) -> Dict[str, Any]:
        """
        Gets information about a AMM or CLMM pool
        Either pool_address or both base_token and quote_token must be provided
        """
        query_params = {
            "network": network,
        }
        if pool_address is not None:
            query_params["poolAddress"] = pool_address
        if base_token is not None:
            query_params["baseToken"] = base_token
        if quote_token is not None:
            query_params["quoteToken"] = quote_token

        return await self.api_request(
            "get",
            f"{connector}/pool-info",
            params=query_params,
            fail_silently=fail_silently,
        )```

This PR was for demonstration purposes for the Botcamp, and it's based on branches that are in progress. I will close it and open a new one when the API for the GatewayHttpClient is stable. And probably I will make everything an executor and `lp_management.py` won't be needed

sdyanglei:
Hi, due to frequent errors on my local setup and a very tight work schedule, I have to temporarily pause running this strategy. But I still wanted to sincerely thank you — this strategy is truly amazing.

It perfectly matches the solution I had envisioned for a long time, and you’ve already implemented it so well. I happened to come across it, and as someone with limited programming skills but long experience in LP operations, I was genuinely excited and grateful.

Once I’m done with my current workload, I’ll definitely resume running your strategy. Thank you again, and I’m also looking forward to more of your great PRs!

PR #7586, opened 2025-04-29
(feat) support for MEXC Assessment Zone

llegoelkelo:
…to enable trading pair in assessment zone

- [x] Tests all pass

As of March 23rd 2025, MEXC restricts who can trade on pairs that are in their Assessment Zone only to certain API keys that have been whitelisted (see announcement [here](https://www.mexc.com/support/articles/17827791522788?handleDefaultLocale=keep)).

These changes make it so that some of the requests made by the MEXC connector are signed if auth is available. In addition to this, the rate source will try to use the connector config file if it exists.

**Tips for QA testing**: QA will be particular difficult for this change because in order to test, you will have to be whitelisted to a pair that is in the assessment zone. I think regression tests are the best way to QA this change.


nikspz:
hi @llegoelkelo is it fixed issue for you? Are you able to trade your token (I think it was RODAI)?

llegoelkelo:
@nikspz yes, I am able to trade our token just fine after my changes. Please keep in mind that assessment zone pairs can only be traded by certain whitelisted API keys.

cardosofede:
@llegoelkelo Thanks for the contribution and getting deeper in the framework.
In this case I think that the only connector that will use this feature is MEXC and adding an extra sign if possible is not going to be std across other connectors.
My recommendation is to manage everything inside the connector and in the method that you are going to send the request, if the auth is present pass the request as is_auth_required True so you don't need to modify core classes

cardosofede:
I accidentally merged this one, pls create it again

llegoelkelo:
@cardosofede your suggestion makes complete sense. I'll make the changes and recreate the PR.

PR #7613, opened 2025-05-16
Refactor and Provide a way to disable TLS 1.3 on connectionsFactory a…

WuonParticle:

- [x] Tests all pass  

This PR introduces the ability to disable TLS 1.3 in the `ConnectionsFactory` class to address compatibility issues with VPNs that may block TLS 1.3 connections. Key changes include:  
- Refactored to a proper singleton pattern to avoid class attribute shadowing by instance attributes
- Added a `set_disable_tls_1_3()` method to toggle TLS 1.3 support.  
- Modified `_get_shared_client()` to use TLSv1.2 when TLS 1.3 is disabled.  
- Ensured proper cleanup of client sessions when TLS settings change.  

- Added comprehensive unit tests for the new functionality.  
- Verified backward compatibility with existing connections.  
- Tested manually with openVPN setup that previously had issues with TLS 1.3.  

**Tips for QA testing**:  
- Verify that existing connections (without TLS 1.3 disabled) remain unaffected.  
- Check for any performance impact (or just verify it works) when TLS 1.3 is disabled.

### DIFF
--- a/hummingbot/core/web_assistant/connections/connections_factory.py
+++ b/hummingbot/core/web_assistant/connections/connections_factory.py
@@ -1,4 +1,5 @@
-from typing import TypeVar
+import ssl
+from typing import List, TypeVar
 
 import aiohttp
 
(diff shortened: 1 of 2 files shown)


rapcmia:
I do not think we need this PR because the setting is not connected to any user configuration or current caller, so users cannot enable the workaround. It also adds a way to lower TLS security for the shared client.

We can revisit it if the team decides to support VPN-related connection settings in the future.

PR #7669, opened 2025-06-17
docs/added new visualization of the codebase for new contributors

ivanmilevtues:

The first two are not relevent but I checked them.

- [x] Tests all pass

This PR contains high-level diagrams for the hummingbot's codebase. You can see how they render in Github here:
https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/hummingbot/on_boarding.md

Our goal is to help new team members (in this case I suppose it is anyone on github) get up to speed with the codebase more quickly. We believe that having a high-level overview from the start is crucial, and diagrams are one of the most effective ways to convey this. By visualizing the codebase, new contributors can gain context more easily and focus on the specific components relevant to their work - we believe that especially if you are working with good willed contirbutors online they will want to start from a single component (module) of the codebase.

I’d love to hear about your current process for onboarding people to the codebase—do you think these diagrams could be integrated into it?

Any feedback is more than welcome! We've just released a free GitHub Action that can automatically update the diagrams and this way keep them always up-to-date.

Full disclosure: we're trying to turn this into a startup, but we're still in a very early stage and figuring out what will actually be useful for people.

I would usually start by opening a discussion, but they are not enabled for this repository so I opened a PR instead!

**Tips for QA testing**:

### DIFF
--- a/.codeboarding/Configuration_Persistence.md
+++ b/.codeboarding/Configuration_Persistence.md
@@ -1,0 +1,167 @@
+```mermaid
+
+graph LR
+
+    Configuration_Management_Core["Configuration Management Core"]
+
+    Application_Settings_Connector_Registry["Application Settings & Connector Registry"]
+
+    Database_Connection_Migration["Database Connection & Migration"]
+
+    Operational_Data_Persistence_Data_Models_["Operational Data Persistence (Data Models)"]
+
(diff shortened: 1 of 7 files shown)


fengtality (changes requested):
See comment

Issue #7676, opened 2025-06-23
Hyperliquid - Connector is missing order state "reduceOnlyCanceled"


PanzerKunst:
### Describe the bug

The following error in the logs:
```
2025-06-20 23:22:54,493 - 560407 - hummingbot.connector.derivative.hyperliquid_perpetual.hyperliquid_perpetual_derivative.HyperliquidPerpetualDerivative - ERROR - Unexpected error in user stream listener loop.
Traceback (most recent call last):
  File "/home/panzerkunst/hummingbot/hummingbot/connector/derivative/hyperliquid_perpetual/hyperliquid_perpetual_derivative.py", line 549, in _user_stream_event_listener
    self._process_order_message(order_msg)
  File "/home/panzerkunst/hummingbot/hummingbot/connector/derivative/hyperliquid_perpetual/hyperliquid_perpetual_derivative.py", line 620, in _process_order_message
    new_state=CONSTANTS.ORDER_STATE[current_state],
              ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
KeyError: 'reduceOnlyCanceled'
```

Indicates that `hyperliquid_perpetual_constants.py` is missing "reduceOnlyCanceled".

This bug has existed for a long time, it's surprising that nobody noticed it until now.

### Steps to reproduce

1. Run a bot against Hyperliquid Perpetual
2. Send Reduce Only orders

### Release version

2.6.1

### Type of installation

Source

PR #7673, opened 2025-06-18
Feat/add headless support

cardosofede:
**Before submitting this PR, please make sure**:

- [x] Your code builds clean without any errors or warnings
- [x] Tests all pass
- [x] You are using approved title ("feat/", "fix/", "docs/", "refactor/")

**A description of the changes proposed in the pull request**:

This PR introduces headless mode functionality that allows Hummingbot's trading core to operate independently of the CLI application, enabling deployment as a standalone trading engine or integration into external systems. The new TradingCore class provides modular access to connector management, market data, and strategy execution requiring the full Hummingbot application interface.

Additionally, this PR fixes a critical issue in V2 controllers where switching between controllers using different connectors (e.g., Kucoin to Binance) caused "Connector not found" errors. The executor orchestrator was retaining positions from previous sessions that referenced unavailable connectors. The fix adds market validation during position loading, storage, and performance reporting to ensure only positions matching the current strategy's markets are processed.
The implementation maintains full backward compatibility - existing CLI functionality and strategies work. The headless mode opens new deployment possibilities including serverless functions, microservices, API integrations, and containerized trading bots, while the position isolation fix ensures reliable operation when between different exchange configurations.

**Tests performed by the developer**:

- unit test
- run all strategies

**Tips for QA testing**:
- do regression tests is a very important change

PR #7691, opened 2025-07-01
fix/ Map reduceOnlyCanceled to CONSTANTS.ORDER_STATE in hyperliquid_perpetual

floor-licker:
This is the fix for #7676

**Root Cause**: 
The Hyperliquid perpetual exchange returns a specific order status `reduceOnlyCanceled` when reduce-only orders are canceled. However, the connector's `ORDER_STATE` mapping only included standard statuses like `canceled`, `filled`, `open`, etc. As a result, when the `_process_order_message` method tried to look up this status on line 620 of `hyperliquid_perpetual_derivative.py`, it resulted in a KeyError.

**Solution**:
Map `reduceOnlyCanceled` to `OrderState.CANCELED` since a reduce-only canceled order is fundamentally a canceled order and should be treated the same in terms of logging and order management.


nikspz:
https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/api/info-endpoint

nikspz:
Added here
https://github.com/hummingbot/hummingbot/pull/7673

PR #7841, opened 2025-11-04
feat/use gateway_api_host as CN and DNS in certs

### LINKED ISSUE #7399: Bug Report: gateway generate-certs: use gateway.gateway_api_host as CN and DNS instead of 'localhost'
### Describe the bug

Hello,

gateway generate-certs has to use gateway.gateway_api_host as CN and DNS. 
'localhost' is hardcoded in source code and hummingbot won't connect to gateway configured with another hostname/fqdn.
It is also better to use something like 'hummingbot' as O instead of 'localhost'.

Thanks

### Steps to reproduce

1. config gateway.gateway_api_host mygateway
2. gateway generate-certs
3. hummingbot restart -> GATEWAY remains OFFLINE because of bad CN/DNS in certificate

### Release version

2.2.0

### Type of installation

Docker

### Attach required files

_No response_

lgrawet:

- [X] Tests all pass

This PR fixes #7399
client_cert will use 'hummingbot' as CN and DNS
server_cert will use 'localhost' as CN and DNS by default OR gateway_api_host attribute when configured.
ca_cert will use 'localhost' as O and CN by default OR gateway_api_host attribute when configured.

./compile
./start
config gateway.gateway_api_host: some_host
gateway generate-certs

Check csr and certs with openssl cli.
Also tested with config gateway_api_host default value
Use certs with hummingbot and gateway, connection OK

**Tips for QA testing**:

### DIFF
--- a/test/hummingbot/core/utils/test_ssl_cert.py
+++ b/test/hummingbot/core/utils/test_ssl_cert.py
@@ -49,7 +49,7 @@ class SslCertTest(unittest.TestCase):
 
         # create the public key, assert that the file exists
         public_key_file_path = temp_dir + "/public_key_test"
-        generate_public_key(private_key, public_key_file_path)
+        generate_public_key(private_key, public_key_file_path, "gateway")
         self.assertEqual(os.path.exists(public_key_file_path), True)
 
     def test_generate_csr(self):
@@ -64,7 +64,7 @@ class SslCertTest(unittest.TestCase):
 
         # create a csr and assert that it exists
         csr_file_path = temp_dir + "/csr_test"
-        generate_csr(private_key, csr_file_path)
+        generate_csr(private_key, csr_file_path, "gateway")
         self.assertEqual(os.path.exists(csr_file_path), True)
 
     def test_sign_csr(self):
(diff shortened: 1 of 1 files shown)


fengtality:
The underlying issue is real — `GatewayHttpClient` uses `ssl.create_default_context()` with hostname verification, so the hardcoded `localhost` SAN in `ssl_cert.py` breaks any remote `gateway_api_host` — and the parameter plumbing here looks right. Two things before this can merge:

1. `generate_csr` puts the host into `x509.DNSName(host)` unconditionally, but OpenSSL only matches IP addresses against `iPAddress` SAN entries, so a gateway configured by IP (the most common remote setup) would still fail verification — please branch on `ipaddress.ip_address(host)` and emit `x509.IPAddress` for IPs.
2. The branch now conflicts with development: `SAN_DNS` there is `[DNSName('localhost'), DNSName('gateway')]` (the `gateway` entry supports Docker compose hostnames) and the client key is now passphrase-encrypted (0b8a355cc), so after rebasing please keep `localhost`/`gateway` in the SAN list alongside the configured host rather than replacing them.

It would also be worth asserting the SAN/CN contents in `test_ssl_cert.py`, since the existing file-existence checks don't cover the new behavior.

PR #7872, opened 2025-12-04
fix: allow decimal values in gateway config updates

fengtality:
## Summary
- Combined `int` and `float` type checks into a single numeric check in `_validate_config_value`
- Allows decimal values (e.g., `1.2`, `0.001`) to be accepted even when the current stored value is an integer
- Returns integers for whole numbers when appropriate to preserve type consistency

## Problem
Gateway config update fails for decimal values like `baseFeeMultiplier 1.2` and `priorityFee 0.001` with "Expected integer value" error.

**Root cause:** When a user changes a value like `baseFeeMultiplier` from `1.2` to `2`, Python stores it as an `int`. When trying to revert to `1.2`, the validation sees the current value is `int` and rejects the decimal input.

## Test plan
- [x] Change `baseFeeMultiplier` from `1.2` to `2`, then revert to `1.2`
- [x] Change `priorityFee` from `0.001` to `1`, then revert to `0.001`
- [x] Verify integer values still work (e.g., `port: 15888`)

Fixes https://github.com/hummingbot/gateway/issues/557

### DIFF
--- a/hummingbot/client/command/gateway_config_command.py
+++ b/hummingbot/client/command/gateway_config_command.py
@@ -294,18 +294,15 @@ class GatewayConfigCommand:
                     self.notify(f"Error: Expected boolean value (true/false), got '{value}'")
                     return None
 
-            elif isinstance(current_value, int):
-                # Integer conversion
+            elif isinstance(current_value, (int, float)):
+                # Numeric conversion - accept both int and float values
+                # This allows reverting from integer to decimal values
                 try:
-                    return int(value)
-                except ValueError:
-                    self.notify(f"Error: Expected integer value, got '{value}'")
-                    return None
-
-            elif isinstance(current_value, float):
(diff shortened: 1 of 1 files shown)


rapcmia:
Commit 4bc6a0f1256b993ddabbebb9148813fc5fefe867
- Changed solana mainnet-beta minPriorityFeePerCU ✅
  image
  - Same test did with `baseFeeMultiplier` ✅
  - Same test with `gateway.gateway_api_port` from `15888` → `09.12312i3`
      - Then revert changes back to `15888`, gateway reconnected after hummingbot relaunch ✅
  - Check docker build with same tests when on secure mode, all ok 
    image
  - All changes are confirmed with each of namespaces.yml files

rapcmia:
Merged to development and part of release version 2.11.0

PR #7978, opened 2026-01-25
fix(pmm): remove dead code in inventory cost price delegate

sjhddh:
## Summary
- Remove redundant `quote_volume -= fee_amount` adjustment in the SELL branch when fee is paid in quote asset
- This code had no effect because `quote_volume` is recalculated for SELL orders later in the method
- Replace with explicit `pass` and clarifying comment addressing the TODO

## Test plan
- Verify existing `inventory_cost_price_delegate` tests pass
- No functional change expected as the removed code was dead code


rapcmia:
@nikspz please inspect and test this PR, else we can close it. Thanks

rapcmia:
Thank you for the cleanup. We confirmed that this line does not change the saved inventory cost because sell orders recalculate the quote volume later in the same flow. Since this does not fix a user-facing behavior, we will close this PR for now.

PR #8016, opened 2026-02-14
fix(executors): pass position_close to budget check for reduce-only perpetual exits

eilyashko:
Closing a perpetual position does not require additional collateral — the existing position is the collateral being released. But `PositionExecutor.validate_sufficient_balance()` treats every order as a new open, requiring fresh margin. This causes standalone exit executors to fail with `INSUFFICIENT_BALANCE` even when the exchange would accept the order.
`PerpetualOrderCandidate` already has a `position_close` flag that zeroes collateral requirements — it just was never threaded from the executor config. This PR connects it.

- Added `position_close: bool = False` to `PositionExecutorConfig` (default preserves existing behavior)
- Passed `position_close=self.config.position_close` to `PerpetualOrderCandidate` in `validate_sufficient_balance()`

1. Created a `PositionExecutor` to open a perpetual position (BUY, 20x leverage)
2. Created a second `PositionExecutor` to close it with `position_close=False` — fails with `INSUFFICIENT_BALANCE` 
3. Same exit with `position_close=True` — passes validation and places the order 
4. New entry with `position_close=False` (default) — still gets balance-checked as before 

**Tips for QA testing**:

Set up an account with margin fully utilized by an open position. Example: $50 margin, all used by a 20x long.
- **New entry** (`position_close=False`): should be blocked — no free margin to open another position 
- **Close existing position** (`position_close=True`): should pass — closing does not require new collateral, it releases it 
This matches exchange behavior: on the exchange web UI, you can always close an open position regardless of available margin.


rapcmia:
Thanks for the contribution! I'll close this PR for now because the current implementation does not match the executor's order flow. The balance check runs before an opening order is placed, so skipping it based on position_close=True can incorrectly bypass validation for new positions.

PR #8024, opened 2026-02-17
fix: prevent market close orders when keep_position=True

rathorvikash010-netizen:
## Summary
This PR fixes an issue where market close orders were still being executed even when `keep_position=True`.

## Problem
When the `keep_position` flag is enabled, the strategy is expected to stop without closing open positions. However, the current behavior still triggers market close orders during stop execution.

## Solution
Added a conditional check to prevent market close orders from being executed when `keep_position=True`.

## Testing
- Verified strategy behavior with `keep_position_on_stop: true`
- Confirmed that open positions remain active after running the `stop` command
- No impact observed when `keep_position=False`

## Impact
This change ensures that strategy stop behavior aligns with expected configuration settings without affecting other execution flows.


nikspz (commented):
Closing since no changes in this PR

PR #8034, opened 2026-02-20
fix: minimum order size calculation for Hyperliquid Perpetual

agnivon:

- [ ] Tests all pass

Minimum order size in the hyperliquid perpetual connector should not be derived from the fractional part of the open interest. The number of digits after the decimal in open interest keeps fluctuating, resulting in different minimum order size calculations and errors due to failed validation during order placement. Implemented minimum order size calculation referring to https://docs.chainstack.com/docs/hyperliquid-order-precision

I ran into errors using grid strike strategy using the hyperliquid perpetual connector. When the minimum order quote value is close to or equal to the minimum notional value of $10, order placements start failing with error messages such as "Order amount x is lower than minimum order size y for the pair . The order will not be created.", when orders with the same size were successfully submitted just a few seconds back. Orders with valid sizes suddenly started failing because of erroneous calculation of minimum order size using open interest, whose fractional part was reduced to fewer digits for a period of time.

**Tips for QA testing**:
Test using the hyperliqid perpetual connector with different strategies by keeping the minimum quote as $10. Orders which can be successfully manually placed on Hyperliquid should not fail during strategy execution.

### DIFF
--- a/hummingbot/connector/derivative/hyperliquid_perpetual/hyperliquid_perpetual_derivative.py
+++ b/hummingbot/connector/derivative/hyperliquid_perpetual/hyperliquid_perpetual_derivative.py
@@ -815,14 +815,16 @@ class HyperliquidPerpetualDerivative(PerpetualDerivativePyBase):
         coin_infos: list = exchange_info_dict[0]['universe']
         price_infos: list = exchange_info_dict[1]
         return_val: list = []
+        min_notional = 10
         for coin_info, price_info in zip(coin_infos, price_infos):
             try:
                 ex_symbol = f'{coin_info["name"]}'
                 trading_pair = await self.trading_pair_associated_to_exchange_symbol(symbol=ex_symbol)
-                step_size = Decimal(str(10 ** -coin_info.get("szDecimals")))
-
+                sz_decimals = coin_info.get("szDecimals")
+                step_size = Decimal(str(10 ** -sz_decimals))
+                price = float(price_info.get("markPx"))
                 price_size = Decimal(str(10 ** -len(price_info.get("markPx").split('.')[1])))
-                _min_order_size = Decimal(str(10 ** -len(price_info.get("openInterest").split('.')[1])))
+                _min_order_size = Decimal(str(round(min_notional / price, sz_decimals)))
                 collateral_token = CONSTANTS.CURRENCY
                 return_val.append(
                     TradingRule(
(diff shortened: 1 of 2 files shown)


cardosofede:
@isreallee82 can you check?

hummingbot/connector/derivative/hyperliquid_perpetual/hyperliquid_perpetual_derivative.py
```diff
@@ -815,14 +815,16 @@ async def _format_trading_rules(self, exchange_info_dict: List) -> List[TradingR
...
                 ex_symbol = f'{coin_info["name"]}'
                 trading_pair = await self.trading_pair_associated_to_exchange_symbol(symbol=ex_symbol)
-                step_size = Decimal(str(10 ** -coin_info.get("szDecimals")))
-
+                sz_decimals = coin_info.get("szDecimals")
+                step_size = Decimal(str(10 ** -sz_decimals))
+                price = float(price_info.get("markPx"))
                 price_size = Decimal(str(10 ** -len(price_info.get("markPx").split('.')[1])))
-                _min_order_size = Decimal(str(10 ** -len(price_info.get("openInterest").split('.')[1])))
+                _min_order_size = Decimal(str(round(min_notional / price, sz_decimals)))
```
isreallee82, line 827: @agnivon you are enforcing $10 minimum size also this can round down to < 10

isreallee82: @cardosofede did cross check hyperliquid minimum order os 10$but this logic coud still fail

isreallee82:
@cardosofede @agnivon will be creating a new pr https://github.com/hummingbot/hummingbot/pull/8061 that addresses new api improvement, docs have been recently updated 12 days ago will push improvement, you can retest this logic for min order size for merge

Screenshot 2026-02-24 at 1 14 19 AM (3)

isreallee82:
@agnivon nice update, but after reviewing https://docs.chainstack.com/docs/hyperliquid-order-precision, noticed its not dynamic enough also The docs say “typically ~$10”, so if Hyperliquid introduces per-market exceptions, we’d need an API field or a configurable value, the last two commit https://github.com/hummingbot/hummingbot/pull/8061 Right now enforces 10 consistently. cc @cardosofede

agnivon:
Thanks for the feedback @isreallee82. PR #8061 seems to includes the proper fix for this by utilizing native min_notional_size validation. That's definitely a much more robust approach than manually calculating the minimum base size from the current price, as it avoids any rounding/price-drop issues.

Closing this PR in favor of #8061

PR #7948, opened 2026-01-15
feat: LP Executor and Controller for StrategyV2

fengtality:
## Summary

This PR introduces LP (Liquidity Provider) strategies for automated liquidity provision on CLMM DEXs (Meteora, Raydium, etc.) with automatic rebalancing when price moves out of range.

**Related PRs:**
- Gateway: https://github.com/hummingbot/gateway/pull/591
- hummingbot-api: https://github.com/hummingbot/hummingbot-api/pull/120

### Architecture

**Controller + Executor Pattern (StrategyV2)**

- **LP Rebalancer Controller** (`controllers/generic/lp_rebalancer/`)
  - Orchestrates LP positions with configurable position width, rebalance timer, and price limits
  - KEEP vs REBALANCE logic based on price limits
  - Folder structure with comprehensive README.md
  
- **LP Executor** (`hummingbot/strategy_v2/executors/lp_executor/`)
  - Manages individual LP position lifecycle
  - State machine: `NOT_ACTIVE → OPENING → IN_RANGE ↔ OUT_OF_RANGE → CLOSING → COMPLETE`
  - Retry logic with user notification on max retries

### Controller Folder Structure

```
controllers/generic/lp_rebalancer/
├── __init__.py
├── lp_rebalancer.py      # Controller implementation
└── README.md             # Comprehensive documentation
```

📖 **Please review the [LP Rebalancer README](https://github.com/hummingbot/hummingbot/blob/feat/lp-strategy-v2-base/controllers/generic/lp_rebalancer/README.md) before testing.** It documents the controller architecture, configuration, rebalancing logic, and troubleshooting guide.

### Additional Changes
- **lphistory command** - View historical LP position performance and P&L
- **Gateway error handling** - Distinguish "Position closed" vs "Position not found" errors
- **Status visualizations** - Position range, price limits (unified scale), rebalance timer
- **WebSocket timeout fix** - Use config values instead of hardcoded 60s
- **Order tracking cleanup** - Stop tracking orders after LP events complete

### Files Added/Modified
- `controllers/generic/lp_rebalancer/` - LP Rebalancer Controller (new folder structure)
- `hummingbot/strategy_v2/executors/lp_executor/` - LP Executor
- `hummingbot/client/command/lphistory_command.py` - History command
- `hummingbot/connector/gateway/gateway_lp.py` - Order tracking fix
- `hummingbot/model/range_position_update.py` - DB schema updates

### Prerequisites
[...]

PR #8045, opened 2026-02-23
fix: clamp order amounts to available balance in LPRebalancer

CrestoSpb:
# fix: clamp LP order amount to available wallet balance with stale-cache guard

- [x] Tests all pass

---

## Description

In `LPRebalancer._calculate_amounts`, the `total_amount_quote` configured by the user was used as-is when sizing LP positions. After a rebalance cycle, the actual wallet balance can be slightly lower than the configured value due to swap fees or impermanent loss, which caused the order to fail with insufficient-funds errors.

This PR clamps the effective order amount to the actual available wallet balance before constructing the position, with a **stale-cache guard** to avoid using an outdated balance reading immediately after a position close.

### Changes

| Side | Behaviour |
|------|-----------|
| **1 — BUY** | Queries the quote token balance (e.g. USDC). If it is less than `total_amount_quote` **and ≥ 80 %** of it, the quote amount is reduced to the available balance. |
| **2 — SELL** | Queries the base token balance (e.g. TRUMP), converts it to quote units at the current price, and reduces the total under the same 80 % guard. |
| **0 — BOTH** | No change — the 50/50 split is unaffected. |

### Stale-cache guard

If the available balance is **below 80 %** of the configured total, the clamp is skipped and the full `total_amount_quote` is used. This handles the window right after a position close where the connector's balance cache is still stale and reports near-zero holdings.

**Without this guard** the bot would open a dust-sized position
(observed: `0.018 TRUMP` instead of `~91 TRUMP`).

> The balance check is wrapped in `try/except` so any provider error falls back
> to the original configured total without crashing.

---

## Tests

Unit tests added in
`test/hummingbot/controllers/generic/lp_rebalancer/test_lp_rebalancer_calculate_amounts.py`:

- **Side 1 (BUY)**
  - No clamping when balance ≥ total
  - Clamping when balance is between 80–100 % of total
  - No clamping when balance is below 80 % (stale-cache path)
  - `None` balance passthrough
  - Exact-equal balance passthrough
- **Side 2 (SELL)**
  - No clamping when base value ≥ total
  - Clamping when base value is between 80–100 % of total
  - No clamping when base value is below 80 % (stale-cache path)
  - `None` balance passthrough
- **Side 0 (BOTH)**
  - `get_balance` is never called
  - 50/50 split is unchanged
- **Exception safety**
  - `RuntimeError` from the provider falls back to configured total for BUY and SELL
- **Token routing**
  - BUY queries the quote token; SELL queries the base token

All tests pass (`pytest` green).

---

## QA Steps

1. Configure `LPRebalancer` with `total_amount_quote` set slightly above the current wallet balance (e.g. configure `100 USDC`, fund wallet with `97 USDC`).

2. Trigger a **BUY-side** position creation — the order should be placed for  `97 USDC` instead of failing with an insufficient-funds error.
   Verify a log line appears:
   ```
   INFO - Clamping quote amount from 100 to available balance 97 USDC
   ```

3. Repeat for **SELL-side**: configure total so that `base_balance × price` is between 80–100 % of `total_amount_quote`; verify the order is placed using the available base amount.

4. **Stale-cache scenario**: ensure the balance API returns a value below 80 % of `total_amount_quote` (e.g. mock near-zero right after a close).
   Verify the bot uses the full configured total and logs:
   ```
   DEBUG - Skipping balance clamp: available ... is below 80% of total ... — balance cache likely stale
   ```
   and does **not** open a dust-sized position.

5. When balance is sufficient (≥ total), verify no clamping log line appears and the full configured amount is used.

### DIFF
--- a/controllers/generic/lp_rebalancer/lp_rebalancer.py
+++ b/controllers/generic/lp_rebalancer/lp_rebalancer.py
@@ -414,17 +414,27 @@ class LPRebalancer(ControllerBase):
         """
         total = self.config.total_amount_quote
 
-        # Clamp to actual available balance to avoid order failures
+        # Clamp to actual available balance to avoid order failures when the balance
+        # is slightly less than expected (e.g. due to fees or impermanent loss).
+        # Only clamp when available >= 80% of total — otherwise the balance cache
+        # is likely stale (e.g. just after a position close) and we should not clamp.
+        MIN_BALANCE_RATIO = Decimal("0.80")
         try:
             if side == 1:  # BUY - needs quote token
                 available_quote = self.market_data_provider.get_balance(
                     self.config.connector_name, self._quote_token
                 )
                 if available_quote is not None and available_quote < total:
-                    self.logger().info(
-                        f"Clamping quote amount from {total} to available balance {available_quote} {self._quote_token}"
-                    )
-                    total = available_quote
+                    if available_quote >= total * MIN_BALANCE_RATIO:
+                        self.logger().info(
+                            f"Clamping quote amount from {total} to available balance {available_quote} {self._quote_token}"
+                        )
+                        total = available_quote
+                    else:
+                        self.logger().debug(
+                            f"Skipping balance clamp: available {available_quote} {self._quote_token} "
+                            f"is below 80% of total {total} — balance cache likely stale"
+                        )
             elif side == 2:  # SELL - needs base token
                 available_base = self.market_data_provider.get_balance(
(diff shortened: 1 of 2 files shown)


fengtality:
Thanks for the suggestion. I implemented a similar fix based on the closed executor amounts in https://github.com/hummingbot/hummingbot/pull/7948/changes/f26a23bacecae70faf2732a01d8fd0c7dc1799d1 that should accomplish the same thing, without dependency on wallet balances which might change.

Issue #8042, opened 2026-02-21
Executors not stored to database when they terminate


fengtality:
## Description

When an executor terminates (e.g., `RunnableStatus.TERMINATED` with `CloseType.EARLY_STOP`), it is not automatically stored to the database. Executors are only persisted when the bot shuts down gracefully via the `stop` command.

## Current Behavior

- `StoreExecutorAction` exists in `hummingbot/strategy_v2/models/executor_actions.py` but is **never emitted** by any controller
- `store_executor()` method exists in `ExecutorOrchestrator` but is never called during normal operation
- Executors are only saved via `store_all_executors()` which is called on bot shutdown

## Expected Behavior

When an executor terminates (status changes to `TERMINATED`), it should automatically be stored to the database so that:
1. Users can analyze completed executors without stopping the bot
2. Data is not lost if the bot crashes or is force-killed
3. Export/analysis scripts can access completed executor data in real-time

## Steps to Reproduce

1. Start an LP Rebalancer controller
2. Wait for an executor to complete (e.g., rebalance or early stop)
3. Check the Executors table in SQLite - it will be empty
4. The executor shows as TERMINATED in the Hummingbot CLI but is not in the database

## Proposed Solution

Controllers should emit `StoreExecutorAction` when an executor terminates. This could be done in:

1. **Controller level**: Each controller emits `StoreExecutorAction` when it detects an executor has terminated
2. **Orchestrator level**: `ExecutorOrchestrator` automatically stores executors when their status changes to `TERMINATED`

Option 2 is cleaner as it handles all executor types centrally.

## Affected Components

- `hummingbot/strategy_v2/executors/executor_orchestrator.py`
- `hummingbot/strategy_v2/controllers/controller_base.py`
- All controllers that manage executors

## Workaround

Run `stop` command in Hummingbot CLI to flush all executors to the database before analyzing.

rafaelrzaev-cmyk:
😂

EdenRochmanSharabi:
The StoreExecutorAction infrastructure exists but is never triggered. Working on wiring it up in the ExecutorOrchestrator.

PR #8048, opened 2026-02-23
fix: store executors to database upon termination (#8042)

ipezygj:
This PR ensures that executors are correctly saved to the database when they are terminated. This fixes the issue where executor data was lost upon stopping. Addresses bounty #8042.


nikspz:
Closed as not inline with https://github.com/hummingbot/hummingbot/blob/master/CONTRIBUTING.md

Issue #8083, opened 2026-02-26
Security: eval() in ws_data_to_dict() allows code execution via WebSocket input


theeggorchicken:
## Summary

`foxbit_utils.ws_data_to_dict()` uses Python's `eval()` to deserialize WebSocket message strings. Because `eval()` executes arbitrary Python code, anything passing through this function that originates from a network-controlled source becomes a code execution path.

I sent a detailed report with reproduction steps and test artifacts to **dev@hummingbot.io**. Happy to share the full writeup here or through whichever channel works best for the team.

## Affected Code

**File:** [`hummingbot/connector/exchange/foxbit/foxbit_utils.py#L70`](https://github.com/hummingbot/hummingbot/blob/master/hummingbot/connector/exchange/foxbit/foxbit_utils.py#L70)

```python
def ws_data_to_dict(data: str) -> Dict[str, Any]:
    return eval(data.replace(":null", ":None").replace(":false", ":False").replace(":true", ":True"))
```

The `.replace()` calls exist because the input is JSON, but `json.loads()` handles `null`, `false`, and `true` natively. There's no need for `eval()` here.

## Impact

Anyone able to inject into the Foxbit WebSocket stream — via MITM, DNS hijack, or a compromised exchange endpoint — can run arbitrary Python code on the machine running the bot. In the official Docker image, the process runs as root.

## Suggested Fix

```python
import json  # already imported in this file

def ws_data_to_dict(data: str) -> Dict[str, Any]:
    return json.loads(data)
```

I've submitted a pull request with this fix and tests as well.

PR #8086, opened 2026-02-26
fix: replace eval() with json.loads() in ws_data_to_dict

theeggorchicken:
## What

Replaces `eval()` with `json.loads()` in `foxbit_utils.ws_data_to_dict()`.

## Why

The function was calling `eval()` on raw WebSocket message strings. The `.replace()` calls before `eval()` were there because the input is JSON — but `json.loads()` handles `null`, `false`, and `true` natively, so `eval()` was never needed. Any caller passing network-received data through this function exposed it to code execution if the input was attacker-controlled.

Reported in issue #8083. Additional context and test artifacts are in the email thread at dev@hummingbot.io.

## Changes

- `hummingbot/connector/exchange/foxbit/foxbit_utils.py` — replace `eval(...)` with `json.loads(data)` on line 70
- `test/hummingbot/connector/exchange/foxbit/test_foxbit_utils.py` — two new tests:
  - `test_ws_data_to_dict_rejects_code_injection`: confirms injection attempts now raise `ValueError`
  - `test_ws_data_to_dict_handles_json_native_types`: confirms `null`/`false`/`true` are handled correctly without the old `.replace()` chain

- [x] Tests all pass

### DIFF
--- a/hummingbot/connector/exchange/foxbit/foxbit_utils.py
+++ b/hummingbot/connector/exchange/foxbit/foxbit_utils.py
@@ -67,7 +67,7 @@ def is_exchange_information_valid(exchange_info: Dict[str, Any]) -> bool:
 
 
 def ws_data_to_dict(data: str) -> Dict[str, Any]:
-    return eval(data.replace(":null", ":None").replace(":false", ":False").replace(":true", ":True"))
+    return json.loads(data)
 
 
 def datetime_val_or_now(string_value: str,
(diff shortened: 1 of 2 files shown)


rapcmia:
Replacing eval() is the right security fix, so this change is needed. However, this PR is not ready to merge yet because some current Foxbit message tests use Python-style values such as True and single quotes. json.loads() correctly rejects those values. Please update the affected message fixtures and confirm that Foxbit sends valid JSON for these user-stream messages then we can review again

PR #8108, opened 2026-03-08
Add System R pre-trade risk gate example script

ashimnandi-trika:
## Summary

- Adds `scripts/systemr_risk_managed_pmm.py`, a pure market making strategy that integrates [System R](https://agents.systemr.ai) pre-trade risk intelligence
- Before placing each bid/ask order, the script calls `POST /v1/compound/pre-trade-gate` with symbol, direction, entry price, stop price, and equity
- Orders are only placed if the risk gate returns `{"allow": true}`; rejected orders are logged with reasons
- Follows the existing `simple_pmm.py` pattern (StrategyV2Base, Pydantic config, `on_tick` lifecycle)

### Key design choices

- **Fail-closed**: if the risk API is unreachable or returns an error, orders are rejected (not placed)
- **Async**: risk gate calls run via `aiohttp` so they don't block the event loop
- **Configurable**: exchange, pair, spreads, stop-loss %, equity, and API key are all Pydantic fields
- **Observable**: `format_status()` shows live approved/rejected counters

### Related

- Demo trading agent: https://github.com/System-R-AI/demo-trading-agent
- System R risk intelligence API: https://agents.systemr.ai

## Test plan

- [ ] Load script in Hummingbot with `create --script-config systemr_risk_managed_pmm`
- [ ] Verify risk gate calls appear in logs (approved and rejected)
- [ ] Verify `status --live` displays risk gate statistics
- [ ] Test with unreachable API endpoint to confirm fail-closed behavior

### DIFF
--- a/scripts/systemr_risk_managed_pmm.py
+++ b/scripts/systemr_risk_managed_pmm.py
@@ -1,0 +1,305 @@
+"""
+System R Risk-Managed Pure Market Making Strategy
+
+A pure market making script that gates every order through System R's
+pre-trade risk intelligence API before execution. Orders that fail the
+risk gate are suppressed, protecting the bot from placing trades that
+violate portfolio risk limits, drawdown constraints, or adverse market
+regime conditions.
+
+System R API docs & demo agent:
+  https://github.com/System-R-AI/demo-trading-agent
+
+Setup:
+  1. Get a free API key at https://agents.systemr.ai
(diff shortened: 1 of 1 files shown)


rapcmia:
Thanks for the contribution! I'll close this PR for now because it introduces a new third-party integration that requires additional design considerations, including secure API key management and handling of trading/account data outside Hummingbot. We can revisit it if the team decides to officially support this provider.

PR #8148, opened 2026-03-27
fix / auto-store executors to database when they terminate

### LINKED ISSUE #8042: Executors not stored to database when they terminate

EdenRochmanSharabi:
## Summary
Fixes #8042

`StoreExecutorAction` exists but is never emitted, so terminated executors are lost from the database. Added `_auto_store_terminated_executors()` in `ExecutorOrchestrator`, called from `get_all_reports()`.

## Changes
- `hummingbot/strategy_v2/executors/executor_orchestrator.py`: new method + wired into reports
- 4 new tests covering termination storage, POSITION_HOLD exclusion, active skip, batch storage


rapcmia:
Hi @EdenRochmanSharabi I’d like to review this PR. If you intend to proceed with it, kindly address the GitHub workflow failure 🙏🏼

EdenRochmanSharabi:
Thanks for looking into this @rapcmia! I've fixed the flake8 indentation issue and pushed the fix. CI should be green now.

rapcmia:
Commit 54e94a9b9dd22356424cd8fd625e2ebdb651dc94

#### Test executor persistence during active runtime
- Ran a `market_making.pmm` V2 controller and monitored executor completion.
- Completed executors did not flush to SQLite while the controller was still running. ❌
- This did not match the PR expectation that terminated executors would be stored without needing `stop`. ❌

Steps to reproduce:
- Create any v2 controller config e.g market_making.pmm_simple
- Start and wait for few trades
- Check `./data/<your controller>.sqlite`
- Observe if terminated executors are flushed to `executior` table

#### Test executor persistence after shutdown ✅
- Ran the `stop` command and waited for the shutdown phase to complete.
- Checked SQLite after shutdown.
- The completed executors were present in SQLite.
- The SQLite records included close type `5` (`EARLY_STOP`) and close type `3` (`TAKE_PROFIT`).

rapcmia (dismissed):
- Test non-POSITION_HOLD executor runtime persistence ✅
	- Verified runtime SQLite persistence for non-`POSITION_HOLD` close types `EARLY_STOP`, `TAKE_PROFIT`, and `INSUFFICIENT_BALANCE`.
	- While the strategy was still running, checked SQLite as soon as the executors reached terminated status.
	- The terminated executors were saved to the `Executors` table without relying on the close buffer.
- Test POSITION_HOLD executor runtime behavior ✅
	- Ran `generic.examples.basic_order_example`, which created filled market `order_executor` entries.
	- While the strategy was still running, confirmed the `Executors` table stayed empty after the executors reached `POSITION_HOLD`.
	- This matched the expected `POSITION_HOLD` exclusion from runtime auto-store.
- Test POSITION_HOLD executor shutdown persistence ✅
	- After stopping the strategy, checked SQLite again.
	- The executor rows were present in SQLite with close type `10` (`POSITION_HOLD`).

hummingbot/strategy_v2/executors/executor_orchestrator.py
```diff
@@ -442,6 +442,32 @@ def stop_executor(self, action: StopExecutorAction):
             return
         executor.early_stop(action.keep_position)
 
+    def _auto_store_terminated_executors(self):
+        """
+        Automatically store executors to the database when they transition to TERMINATED status.
+        This ensures executor data is persisted without requiring an explicit StoreExecutorAction.
+        Executors with POSITION_HOLD close type are excluded as they are handled by position tracking.
```
cardosofede: this is not true... we should also store the executors with position hold, the final position is managed automatically.

cardosofede (changes requested):
first of all thanks for the contribution. review that comment and also one thing that I don't like about this PR is that it will automatically store and delete from memory the terminated executors... and the most of the strategies check what was the last executor terminated wiht this condition (inside a queue of lets say 50 executors), to determine if we need to cooldown or the next price to place an order... with this we lose them, so I think that will be better to auto store them when they are terminated and sent to a deque or something like that so it's automatically cleaned after x executors.

EdenRochmanSharabi:
Thanks for the thorough review @cardosofede! I've pushed a fix addressing both points:

1. Removed the POSITION_HOLD exclusion: all terminated executors are now auto-stored.
2. Replaced immediate removal with a bounded `deque(maxlen=50)` per controller, so strategies can still inspect recently terminated executors for cooldown/pricing logic. The deque is also included in `get_executors_report()`.

Let me know if the maxlen of 50 works or if you'd prefer it configurable.

hummingbot/strategy_v2/executors/executor_orchestrator.py
```diff
@@ -442,6 +443,34 @@ def stop_executor(self, action: StopExecutorAction):
...
         executor.early_stop(action.keep_position)
 
+    def _auto_store_terminated_executors(self):
+        """
+        Automatically store terminated executors to the database and move them to a bounded deque
+        so strategies can still inspect recently terminated executors for cooldown or pricing logic.
+        """
+        for controller_id, executors_list in list(self.active_executors.items()):
+            stored = []
+            for executor in executors_list:
```
cardosofede, line 516: here I will just simplify and check for status in TERMINATED

cardosofede (changes requested):
simplify that condition mentioned and also do a careful review on the PNL reported since when we are storing the executors in the DB, we update the cached performance. I saw that you are using the cached executors in the report of all executors (make sure here that you are not processing the pnl and volume of them twice).
regarding to the 50 executors in the queue I think that is enough

EdenRochmanSharabi:
hey, simplified the condition to just check for status == TERMINATED as you mentioned, pushed that now.

regarding the PNL double counting, I went through the flow carefully and it should be fine... when `_auto_store_terminated_executors` runs, it updates `cached_performance` and moves the executor from `active_executors` to the deque. then in `generate_performance_report` we start from `cached_performance` (which already has the terminated ones) and only add data from `active_executors` (which no longer contains them). the deque is only used in `get_executors_report` so strategies can see recent terminations, but it's not iterated in the performance calculation so nothing gets counted twice.

rapcmia:
Commit 7041b51a58ce24fc7fffbc49eca06a91b6acb5a1

#### Test non-POSITION_HOLD executor runtime persistence using `market_making.pmm_simple`✅
- Kept the bot running and checked SQLite before any stop/shutdown command.
- Confirmed the bot process was still active after more than 15 minutes.
- Confirmed all stored executor rows were non-`POSITION_HOLD`
- Confirmed close types included `1`, `3`, and `5`.
- Confirmed SQLite had matching activity in `TradeFill` and `Order` tables.
- Confirmed `TradeFill` had 18 rows and `Order` had 25 rows during runtime.

---

#### Test POSITION_HOLD runtime executor persistence and held-position tracking
image

- Ran generic.examples.basic_order_example with order_frequency set to 120.
- Kept the bot running and checked SQLite and confirmed the bot process was still active after more than 5 minutes.
- Confirmed 3 completed BUY market orders were recorded in logs at 20:36:40, 20:38:40, and 20:40:41.
- Confirmed SQLite Executors table had 3 runtime rows with close_type = 10 (POSITION_HOLD) and status = 4 (TERMINATED).
- Confirmed SQLite TradeFill and Order rows matched the 3 completed BUY orders.
- However checking the status, i see `no position held` ❌ 
  - Check `executor` table, custom_info has records of `held_position_orders`
  - Yet the `position` table is left blank even when the bot has initializd `stop` command
  - Compared with development branch, not able to reproduce ✅
    image
      - Status shows held position during runtime, and Position table is populated after stop\
      - See attached for the controller config used on this test: [08062026.zip](https://github.com/user-attachments/files/28710066/08062026.zip)

EdenRochmanSharabi:
The POSITION_HOLD issue was the ordering in `get_all_reports()`: `_auto_store_terminated_executors()` was removing POSITION_HOLD executors from `active_executors` before `_update_positions_from_done_executors()` could read them. Swapped the two calls so positions are extracted first. All 37 orchestrator tests pass.

PR #8151, opened 2026-04-02
docs: refresh stale README documentation links

xsfX20:
## Summary
- replace the stale installation quick link with the current installation overview
- point the Condor onboarding link at the current Condor landing page
- switch the Gateway installation reference to HTTPS

## Testing
- verified the replacement documentation URLs return live pages with curl


nikspz (approved):
checked that new links is valid

PR #8152, opened 2026-04-02
docs: sync Cursor test ignores with Makefile

xsfX20:
## Summary
- sync the VS Code/Cursor pytest ignore example with the current `make test` exclusions from `Makefile`
- replace stale injective-specific ignores with the ndax, test/mock, and cross_exchange_market_making paths that are currently skipped in the repo test entrypoint
- clarify that the IDE example should stay aligned with `make test` so local test discovery matches the documented workflow

### DIFF
--- a/CURSOR_VSCODE_SETUP.md
+++ b/CURSOR_VSCODE_SETUP.md
@@ -30,13 +30,14 @@ CONDA_ENV=hummingbot
         "test",
         // "-v",  // optional: verbose output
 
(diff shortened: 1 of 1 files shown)


xsfX20:
Low current-value self-audit: this only syncs editor/Cursor test-ignore documentation with Makefile paths. It is not a core product fix or maintainer-requested change, so I am closing it to avoid reviewer burden.

Thanks for the review bandwidth; I am closing this myself so maintainers do not need to spend time on a low-value PR.

PR #8181, opened 2026-04-17
fix / use total balance for perpetual margin check in validate_sufficient_balance (#8168)

### LINKED ISSUE #8168: Bug Report: Backpack - Wrongly reporting Not enough budget to open position.
### Describe the bug

Hi Hummingbot Team,

I'm getting an issue with Backpack where Hummingbot is reporting `Not enough budget to open position.` erroneously.
In short, my account balance is $20 and I'm using a directional controller with the following parameters
```
  total_amount_cote = 40
  leverage = 5
  max_executors_per_side = 2
  position*mode = ONEWAY
```

The fist buy is OK with an initial margin a bit less of $4 x 5 = $20 position but then, Hummingbot is reporting `Not enough budget to open position.` on the second buy.

I've developed a dumb [buy_v1.py](https://github.com/user-attachments/files/26548000/buy_v1.py) controller to demonstrate the problem. This controller buys on each tick so slow it down with the cooldown_time parameter!
Here is the controller config
```
id: D1BwywWp9TQsniD89xe1woVVCa6LJwGo2viM6dG1s8C
controller_name: buy_v1
controller_type: directional_trading
total_amount_quote: '40'
manual_kill_switch: false
initial_positions: []
connector_name: backpack_perpetual
trading_pair: SOL-USDC
max_executors_per_side: 2
cooldown_time: 30
leverage: 5
position_mode: ONEWAY
stop_loss: '0.03'
take_profit: '0.02'
time_limit: 2700
take_profit_order_type: 2
trailing_stop:
  activation_price: '0.015'
  trailing_delta: '0.005'
candles_config: []
candles_connector: binance_perpetual
candles_trading_pair: SOL-USDC
interval: 5m
topic: solusdc
```

### Steps to reproduce

1. Balance $20
2. Run the buy_v1 controller with the aforementioned config

### Release version

2.13

### Type of installation

Docker

### Attach required files

_No response_

iamvrajpatel:
## Summary

Fixes #8168 — erroneous "Not enough budget to open position" on `backpack_perpetual` when using leverage with multiple executors per side.

## Root Cause

On cross-margin perpetual exchanges like Backpack, the REST balance API reports `available` as near-zero after a position opens—it deducts the full position notional rather than just the initial margin. This caused `validate_sufficient_balance` to incorrectly fail the budget check for the second executor even when sufficient margin existed.

## Fix

Set `from_total_balances=True` on `PerpetualOrderCandidate` in `validate_sufficient_balance` so the `BudgetChecker` uses `get_balance()` (total equity) instead of `get_available_balance()`. Since `_get_collateral_amount()` already divides by leverage (`order_size / leverage`), this correctly computes the required margin from total equity—the correct metric for cross-margin perpetual accounts regardless of how each exchange defines "available" in their balance API.

Also improves error and info logging to include balance values for easier debugging.

## Changes

- `hummingbot/strategy_v2/executors/position_executor/position_executor.py`:
  - Added `from_total_balances=True` to `PerpetualOrderCandidate` in `validate_sufficient_balance`
  - Added info-level logging before every balance check (connector, pair, leverage, required collateral, available and total balance)
  - Improved error message to include actual balance values

- `test/hummingbot/strategy_v2/executors/position_executor/test_position_executor.py`:
  - Added `test_validate_sufficient_balance_perpetual_with_leverage` regression test
  - Simulates the Backpack scenario: `leverage=5`, `total=$20`, `available≈$0`
  - Verifies check passes when total equity is sufficient
  - Verifies check fails when total equity is genuinely insufficient


cardosofede:
@tomasgaudino can you review If the available balance can eb fixed in backpack?

nikspz:
hi @iamvrajpatel please review https://github.com/hummingbot/hummingbot/blob/master/CONTRIBUTING.md 

PR should be to the development branch. (and using latest development branch also (not master))

hummingbot/VERSION
nikspz, line 1: Those changes is not needed in this PR to development @iamvrajpatel

PR #8190, opened 2026-04-20
Fix: Skip Gateway API calls for CEX connectors in market data provider

fengtality:
## Summary
- Only query Gateway's `get_default_swap_provider` for connectors that start with known chain names (ethereum, solana) from the `Chain` enum
- This prevents unnecessary Gateway API calls and error logs for CEX connectors like `hyperliquid_perpetual` during headless startup

## Problem
During v2.14 staging testing with a v2 controller autostart in Docker, Gateway returns errors for CEX connectors:
```
Failed to get default swap provider for hyperliquid_perpetual: Gateway error: Namespace 'hyperliquid_perpetual' not found (NotFoundError)
```

The issue is in `market_data_provider.py` which was calling `get_default_swap_provider()` for ALL connectors to detect gateway connectors, instead of checking first if the connector is a known gateway network.

## Solution
Check if the connector name starts with known chain names from the `Chain` enum (`ethereum`, `solana`) before making any Gateway API calls. CEX connectors like `hyperliquid_perpetual`, `binance`, etc. will skip the Gateway check entirely.

## Test plan
- [x] Start a v2 controller with a CEX connector (e.g., hyperliquid_perpetual) in headless mode
- [x] Verify no Gateway "namespace not found" errors appear in logs
- [x] Verify gateway connectors (solana-mainnet-beta, ethereum-mainnet) still work correctly


fengtality:
@rapcmia I tested this with `bin/quickstart.py` with CEX connectors and don't see  "namespace not found" errors in logs.

nikspz (approved):
- Started a v2 controller with a CEX connector ✅
    - started hyperliquid_perpetual using Client ✅
    - hyperliquid_perpetual in headless mode: no errors ✅
    - binance in quickstart mode: no errors ✅
- Verified no Gateway "namespace not found" errors appear in logs ✅
- Verified gateway connectors (solana-mainnet-beta, ethereum-mainnet) still work correctly
    - gateway swap BUY/SELL: ok ✅
    - gateway lp: added / removed successfully ✅
    - generic.lp_rebalancer.lp_rebalancer: started successfully using quickstart , added / removed successfully ✅

setup.py
```diff
@@ -23,7 +23,7 @@ def build_extensions(self):
 
 def main():
     cpu_count = os.cpu_count() or 8
-    version = "20260302"
+    version = "20260421"
```
cardosofede, line 26: this change should be in dev already right?

cardosofede (approved):

setup.py
nikspz: yes

PR #8205, opened 2026-04-27
fix(gemini): use seconds-based nonce for API authentication

shivanshi-das:
## Summary
Fixes InvalidNonce error in Gemini connector by correcting nonce timestamp format.

## Problem
The connector was generating nonces in milliseconds format (e.g., `1777322173198`), but Gemini's API expects nonces in seconds format (e.g., `1777322173`) to match their server time. This caused all authenticated requests to fail with:
```
InvalidNonce: Nonce '1777322173198' is not within 30 seconds of server time '1777322173'
```

## Changes
- Changed `NonceCreator.for_milliseconds()` to `NonceCreator.for_seconds()` in `gemini_auth.py`

## Test plan
- Tested authenticated API requests (balances endpoint) against Gemini sandbox
- Verified nonce values are now in correct seconds format
- Confirmed requests succeed without InvalidNonce errors


ybapat:
LGTM

PR #8212, opened 2026-05-07
fix / Hyperliquid auth silently accepting wrong private keys (#7866)

### LINKED ISSUE #7866: Hyperliquid - no error returned for incorrect private key
### Describe the bug

Hyperliquid and hyperliquid_perpetual - no error returned for incorrect private key

### Steps to reproduce

1. clone latest development branch
2. connect hyperliquid or hyperliquid_perpetual
3. use wrong private key

Actual:
connected successfully with incorrect private key

Expected:
returns error wrong key

### Release version

dev-2.11.0

### Type of installation

Source

### Attach required files

Image

Image

Image

ipezygj:

- [x] Tests all pass

Fixes #7866. Both `HyperliquidAuth` (spot) and `HyperliquidPerpetualAuth` (perpetual) constructors silently accepted any well-formed 32-byte hex string as `api_secret` without verifying it derived to the supplied `api_address`. Users supplying a wrong key reached a "connected" state in the CLI; subsequent signed requests were rejected by the exchange, but the local error surface gave no indication of misconfiguration. This PR adds construction-time validation so the misconfiguration is surfaced immediately at connect time.

**Root cause**

`eth_account.Account.from_key()` accepts any 32-byte hex value as a valid key (cryptographically, any 32-byte integer can be a private key) and derives a wallet address from it. The auth classes never compared the derived address against the user-supplied `api_address`, so a key-for-different-account passed silently. Subsequent `sign_l1_action` calls signed payloads with the wrong key; the exchange rejected those signatures, but Hummingbot's connect-time UX showed success.

**Fix**

In both auth `__init__` methods:

1. Validate `api_secret` is a non-empty string and parseable by `eth_account.Account.from_key()`. Wrap any underlying exception in a `ValueError` with a clear message.
2. Validate `api_address` is a non-empty string and a valid hex Ethereum address (`eth_utils.is_hex_address`).
3. In non-vault mode (`use_vault=False`), checksum-case both the address derived from the private key and the supplied `api_address`, then require them to be equal. On mismatch, raise `ValueError` with both addresses in the message so the user can see exactly what is wrong.
4. In vault mode (`use_vault=True`), skip the derive-and-compare check because the supplied `api_address` is the vault address rather than the API wallet address; the API wallet derived from `api_secret` is the agent.

**Files changed**

| File | Change |
|---|---|
| `hummingbot/connector/exchange/hyperliquid/hyperliquid_auth.py` | Add validation block in `__init__`; extend `eth_utils` import. |
| `hummingbot/connector/derivative/hyperliquid_perpetual/hyperliquid_perpetual_auth.py` | Same change applied to the perpetual auth. |
| `test/hummingbot/connector/exchange/hyperliquid/test_hyperliquid_auth.py` | Update `setUp` address; add `HyperliquidAuthValidationTests`. |
| `test/hummingbot/connector/derivative/hyperliquid_perpetual/test_hyperliquid_perpetual_auth.py` | Same change applied to the perpetual auth tests. |
| `test/hummingbot/connector/exchange/hyperliquid/test_hyperliquid_exchange.py` | Replace placeholder `"someAddress"` fixture with key-derived address. |
| `test/hummingbot/connector/derivative/hyperliquid_perpetual/test_hyperliquid_perpetual_derivative.py` | Same fixture update. |
| `test/hummingbot/connector/exchange/hyperliquid/test_hyperliquid_user_stream_data_source.py` | Same fixture update. |
| `test/hummingbot/connector/derivative/hyperliquid_perpetual/test_hyperliquid_perpetual_user_stream_data_source.py` | Same fixture update. |

**Backwards compatibility**

This is a behavioral change: configurations that previously appeared to connect with a mismatched key now raise `ValueError` at connect time. This is the intended fix per the issue. Vault-mode users are unaffected because vault mode bypasses the derive-and-compare check.

Five new test cases per auth class cover the validation paths and are included in the diff:

- `test_invalid_private_key_format_raises` — non-hex string for `api_secret`
- `test_invalid_address_format_raises` — non-address string for `api_address`
- `test_address_does_not_match_key_raises` — well-formed key and address but mismatched
- `test_vault_mode_bypasses_address_match_check` — `use_vault=True` accepts mismatched address
- `test_empty_inputs_raise` — empty string for either field

The pre-existing test suites in both connectors used a placeholder `"someAddress"` for `api_address` together with a real test private key. Those four fixtures were updated to use the address derived from the existing test private key (`0x836eE2b55d173245832995082a8600709c38D099`) so the new validation passes; no behavioral assertion in the existing tests changes.

**Tips for QA testing**:

1. **Positive case (vault mode unchanged)**: configure a Hyperliquid perpetual connector with `use_vault=True`, any vault address, and the API wallet's private key. Connection should succeed exactly as before.
2. **Positive case (non-vault, matching pair)**: configure with `use_vault=False`, the wallet address that the private key derives to, and that private key. Connection should succeed.
3. **Negative case (this PR's fix)**: configure with `use_vault=False`, a wallet address that does NOT match the private key (e.g. swap the address from a different wallet). Previously this reached "connected" state and only failed on first signed request. After this PR, `connect` raises `ValueError` immediately with both the derived and the supplied address in the message.
4. **Negative case (malformed inputs)**: empty string or non-hex value for either field raises `ValueError` at construction time with a clear message.


ipezygj:
@rapcmia ready for review when convenient — fixes #7866. Thanks!

rapcmia:
Thanks @ipezygj for the fix, ill have my colleague look into it
@nikspz can you please check this PR

ipezygj:
Pushed `6ec59de` — fixture-only follow-up: replaced placeholder addresses (`"testkey"`, `"testaddress"`, `"test-address"`) with the already-valid `0x836eE2b5...D099` pair used in the auth-test fixtures so the construction-time validation passes. No production-code changes from `ca9e0afb`. Should clear the 55 stable-test failures from the previous run.

ipezygj:
Heads-up for review: the latest commit (370d25cef) changes the behaviour of this PR based on a correctness issue I found in the original strict check.

**Problem:** the original `derived != provided` check raised `ValueError` whenever the private key didn't derive to the supplied wallet address unless `use_vault=True`. That rejects the **documented `api_wallet` (Hyperliquid API / agent wallet) flow**, where the agent key by design does *not* derive to the user's trading address even with `use_vault=False` (see the `hyperliquid_mode` config: prompt *"Enter your API wallet private key (from https://app.hyperliquid.xyz/API)"*). As written it would have been a broader regression than the silent-wrong-key bug #7866 it fixes.

**Change:** threaded the connector's connection mode into `HyperliquidAuth` / `HyperliquidPerpetualAuth` via a new `connection_mode` param (default `"arb_wallet"`, backward compatible). The derive-and-compare check is now skipped when `connection_mode == "api_wallet"` (in addition to vault mode). It still applies in `arb_wallet` mode, so the #7866 protection is preserved exactly there. Error message updated to also point users at api_wallet mode.

**Tests:** added `test_api_wallet_mode_bypasses_address_match_check` to both auth suites; existing arb_wallet/vault tests are unchanged (default mode preserves prior behaviour).

Note: I was unable to run the unit suite locally (Cython core not built in my env), so I'm relying on CI for the test run — flake8/isort/py_compile are clean on the changed files.

@nikspz thank you

nikspz (commented):
commit 370d25c

Steps to reproduce:
1. Clone and install PR8212
2. connect hyperliquid or connect hyperliquid_perpetual
3. Select connection mode (arb_wallet/api_wallet) >>> arb_wallet
4. Do you want to use the Vault address? (Yes/No) >>> no
5. use wallet address (0x08940dc9B5a..............................8067E6843 ) for both Arbitrum wallet address and Arbitrum wallet private key. Or use random text for private key

Expected:
not connected or error

Actual:
You are now connected



image
image
image

ipezygj:
@nikspz thank you for the precise reproduction — it exposed why the earlier validation didn't fire in your steps, and I've pushed a fix (`63427eae1`).

**Root cause:** the validation I'd added lived in `HyperliquidPerpetualAuth.__init__` / `HyperliquidAuth.__init__`, but that code never runs during `connect`:

- `UserBalances.connect_market` builds the connector via `ConnectorSetting.conn_init_parameters(...)`, whose `trading_required` defaults to **`False`** (and `connect_market` doesn't override it).
- The connector's `authenticator` property is `if self._trading_required: return HyperliquidPerpetualAuth(...)` → with `trading_required=False` it returns `None`, so the auth class (and its validation) is **never constructed** at connect time.
- The connect "success" check is `_update_balances()`, which POSTs to the **public** `clearinghouseState` info endpoint keyed by address (no signature). Any well-formed address returns data → no error → "You are now connected". The private key is never exercised on this path.

So the auth-class validation only protected the trading path (strategy run, `trading_required=True`), not the `connect` path you tested — exactly your observation on `370d25c`.

**Fix:** moved the validation into the connector **config maps** (`hyperliquid_perpetual_utils.py` + `hyperliquid_utils.py`), which *are* parsed during `connect` (`prompt_for_model_config` → `Security.update_secure_config`; the pre-existing `validate_mode`/`validate_address` validators confirm this path runs):

- `field_validator` on the secret key → rejects non-parseable / empty keys with a clear message.
- `model_validator(mode="after")` → in `arb_wallet` non-vault mode, requires the key to derive (checksummed) to the supplied Arbitrum address. Bypassed for `api_wallet` (agent key, by design doesn't derive to the trading address) and vault mode — same semantics as the auth class. Defers cleanly during field-by-field prompting until both address and key are populated.
- Applied to all four config maps (spot + perpetual, mainnet + testnet). Auth-class validation kept as defense-in-depth for the trading path.

**Tests:** added `HyperliquidKeyValidationTests` / `HyperliquidPerpetualKeyValidationTests` (incl. testnet) covering random key, empty key, mismatched address, matching pair, and the `api_wallet`/vault bypasses.

Re-test of your steps after this commit (`arb_wallet`, Vault = No, random text or a non-matching address/key) now raises a `ValueError` at connect time instead of reporting connected. flake8/isort/py_compile are clean; I verified the validator logic in isolation but couldn't run the full unit suite locally (Cython core not built in my env), so relying on CI for the suite. Thanks again 🙏

isreallee82:
Reviewed this and ran it locally (applied the branch, ran the new tests + a
simulation of the interactive `connect` prompt flow). The `arb_wallet` fix is
solid a wrong/malformed key now errors at connect instead of silently
"connecting" also  Nice catch that the config-map validators are what actually run at
connect time (`trading_required=False` skips the auth path), and the
`model_construct`/"not provided yet" handling is correct.

**Main thing to resolve before this matches "errors like the rest of the
connectors": `api_wallet` and vault modes still accept a wrong-but-well-formed
key silently.** The derive-match check is (correctly) skipped for them, so the
original #7866 behavior remains in those two modes. Verified:

| mode | wrong-but-valid key | result |
|---|---|---|
| arb_wallet | mismatched/ malformed | ⛔ rejected (this PR ✅ OK |
| api_wallet | wrong agent key | ⚠️ connects, no error |
| vault | wrong vault address | ⚠️ connects, no error |

**Suggestion that also lets you drop the `connection_mode` plumbing:** instead of
branching by mode, validate the same question for both modes *can this key
trade for this wallet?* which is true if the key **derives to the wallet**
(arb_wallet, offline) **or** its address is an **approved agent** of the wallet
(api_wallet). The owner-shortcut means a correct arb_wallet key still costs zero
network calls; api_wallet does one `extraAgents` lookup. This is mode-agnostic,
so the `connection_mode` param added to the auth classes isn't needed for
validation (signing never used it). Hyperliquid can't catch this the usual
"authenticated balance call fails" way because the balance query is an
unauthenticated `/info` request keyed by address — hence the explicit check.

I have a working prototype of this (both connectors + tests, runs the check once
through `_update_balances` so it surfaces at connect) and can push it / open a
follow-up if useful.

`extraAgents` is what Hyperliquid's official
Python SDK uses (`Info.extra_agents`) and is mirrored by QuickNode/Chainstack,
alternative is to sign a cheap L1 action with the agent key and treat the
`"User or API Wallet 0x… does not exist"` error as "not approved" (matches the
"authenticated call fails" pattern, at the cost of a signed write at connect).

Screenshot 2026-06-17 at 6 14 55 PM (2)

https://docs.chainstack.com/docs/hyperliquid-debugging-signature-errors#:~:text=Use%20agent%20wallets%20(-,recommended


Happy to approve for `arb_wallet`; let's close the `api_wallet`/vault gap or
document it explicitly since the stated goal is connector parity.

isreallee82:
@ipezygj you can review https://docs.chainstack.com/docs/hyperliquid-debugging-signature-errors#:~:text=Use%20agent%20wallets%20(-,recommended if that solve api_wallet part

isreallee82:
One addition node @ipezygj 

pydantic_core._pydantic_core.ValidationError: 1 validation error for hyperliquid_perpetual
hyperliquid_perpetual_secret_key
Value error, Invalid Hyperliquid private key: not a valid 32-byte hex private key.
[input_value='7b2263727970746f223a207b...', input_type=str]

Steps to reproduce:

Clone and install PR #8212
connect hyperliquid (or connect hyperliquid_perpetual)
Select connection mode (arb_wallet/api_wallet) >>> arb_wallet
Do you want to use the Vault address? (Yes/No) >>> no
Enter your Arbitrum wallet address >>> 0x08940dc9B5a...8067E6843
Enter your Arbitrum wallet private key >>> <the real 32-byte hex key for that address>
You are now connected to hyperliquid_perpetual


Quit Hummingbot, relaunch it, and log in with your password
Expected:
Logs in normally and loads the saved Hyperliquid keys (as on master)

Actual:
Crashes during startup decryption, before any prompt:

  pydantic_core._pydantic_core.ValidationError: 1 validation error for hyperliquid_perpetual
  hyperliquid_perpetual_secret_key
    Value error, Invalid Hyperliquid private key: not a valid 32-byte hex private key.
    [type=value_error, input_value='7b2263727970746f223a207b...', input_type=str]

(input_value 7b2263727970746f... is hex for {"crypto": ...}, i.e. the still-ENCRYPTED key.)

Notes:
- Happens in every mode (arb_wallet and api_wallet, vault or not) — the failing
  check is an unconditional field validator on the secret key, so the mode never
  matters.
- Simpler repro: anyone who already has a saved Hyperliquid config just needs to
  check out PR #8212 and log in — instant crash, no connect step needed.

Root cause: load_connector_config_map_from_file() runs model_validate() on the raw
yml BEFORE decrypting it, so validate_private_key_format() calls
eth_account.Account.from_key() on the encrypted blob and raises.



```python
# config_helpers.py — load_connector_config_map_from_file
hb_config = ...model_validate(config_data)   # validators run here, secret still ENCRYPTED
config_map = ClientConfigAdapter(hb_config)
config_map.decrypt_all_secure_data()         # decryption happens AFTER

**Comment by ipezygj** [comment]
@isreallee82 Thanks for the detailed review and the crash report on encrypted key validation.

**Encrypted key crash (2026-06-17 20:02 comment):** The current commit (ad5be41) already avoids this — there are no `@field_validator` decorators on the `*_secret_key` fields in the config maps, so `model_validate()` never calls `eth_account.Account.from_key()` on the encrypted blob. Validation only runs in the auth `__init__`, which receives decrypted keys at runtime.

**api_wallet/vault validation:** You're right that these modes don't validate at connect time. I've added docstrings (commit 65d943d) documenting that:
- `arb_wallet`: validated at connect (offline key-to-address check) ✅
- `api_wallet` / `vault`: validated at first authenticated request (would require network call in sync `__init__`)

This matches the pattern you described: "surface the misconfiguration at the earliest feasible point." For `arb_wallet` that's connect time (offline check); for `api_wallet`/`vault` it's first-request time (needs async API call or test-sign, which `__init__` can't do).

If you have a working prototype with the `extraAgents` check integrated into the connector (async, post-connect), I'm happy to incorporate it — just let me know or push it as a follow-up commit to this branch. Otherwise this documents the current behavior clearly per your "document it explicitly" suggestion.

CI running now to confirm no regressions. Thanks again 🙏

**Comment by ipezygj** [comment]
Fixed test failures from the latest CI run:

1. **`test_authenticator_when_not_required`** - Updated to expect auth creation even when `trading_required=False`, since post-#7866 auth is always created when secret_key is provided (so validation runs at connect time)

2. **Builder test fixtures** - Updated `_build_connector()` to use the derived address `0x836eE2b55d173245832995082a8600709c38D099` instead of placeholder `0x1111...`, since arb_wallet mode now validates key-to-address match

Also rebased on latest development to pick up recent test suite additions.

Pushed commits:
- `97f93fe` - test: fix test_authenticator_when_not_required 
- `7be12a6` - test: fix builder test fixtures to use derived address

CI running now.

**Comment by ipezygj** [comment]
Cherry-picked the flaky test fix from #8318 (commit `54e8ae1`) so CI should pass now without waiting for #8318 to be merged.

The timeout issue was causing random failures unrelated to this PR's auth validation changes.

**Comment by ipezygj** [comment]
@isreallee82 Thanks again for the thorough review and the prototype offer — I've implemented the mode-agnostic authority check you suggested (commit `f8ee05596`), closing the `api_wallet` gap.

**Approach (as you described):** validate the same question for every mode — *can this key trade for this account?* — true if the key **derives to the account** (arb_wallet, offline, zero network calls) **or** its address is an **approved agent** of the account (api_wallet, one `extraAgents` `/info` lookup). This is mode-agnostic, so validation no longer needs to branch on `connection_mode`.

- New predicate `HyperliquidAuth.is_key_authorized(derived, account, approved_agents)` / `HyperliquidPerpetualAuth.is_key_authorized(...)` — pure, checksum-normalised.
- The connector runs it once at connect via `_update_balances` (`_verify_key_authority`), so a wrong agent key now errors at connect instead of silently "connecting" — the owner case still costs zero network calls, api_wallet does one `extraAgents` lookup.
- Applied to **both** spot and perpetual connectors.

| mode | wrong-but-valid key | before | after |
|---|---|---|---|
| arb_wallet | mismatched/malformed | ⛔ rejected | ⛔ rejected |
| api_wallet | unapproved agent key | ⚠️ connects | ⛔ rejected |

**Two deliberate conservative choices — flag if you'd tighten either:**
- **Vault mode is skipped** (not hard-checked). I couldn't confirm the vault↔agent-set semantics of `extraAgents` against a real vault without a test account, so rather than risk false-rejecting a valid vault config I left it as first-request validation (documented). If your prototype confirms vault agents surface via `extraAgents`, I'm glad to extend the same check to vault.
- **Network / unexpected-response errors don't block connect** (logged) — a truly wrong key still fails at the first signed request; this only makes the common case explicit and avoids false-rejects on an endpoint hiccup.

**Tests:** predicate cases (owner / approved agent / unapproved / empty-agents / checksum) plus connect-time cases (owner-no-network / approved / unapproved-raises / vault-skip / malformed-response) on both connectors. I ran the auth + new `KeyAuthority` suites locally (green); the `aioresponses`-based balance tests don't run in my local env (an aiohttp/aioresponses version mismatch, unrelated to this change), so I'm relying on CI for those.

Happy to reconcile with your prototype if it differs — especially on the vault path. Thanks! 🙏

**Comment by rapcmia** [comment]
Commit d3b6ee34b105be147c0746562e757529c593b949
- Successfully connected valid api and arb wallets
- All tested with vaults set to false 

#### Test with API wallet invalid pkey or wallet address ❌ 
image
- hmm I’m not sure why the error refers to an agent wallet. Since the issue occurs during the API setup, perhaps we should describe it as an API-related wallet error instead? 

---

#### Test with arb_wallet invalid pkey or wallet address ❌ 
image

- This looks like a valid error for invalid `pkey` or` wallet address` on arb_wallet test, however think it would been better if we can:
  image
  - Add this line after the connector setup prompt
      ```
      Verify the private key matches the wallet address you intend to trade from, set use_vault=True if the supplied address is a vault address, or select the api_wallet connection mode if you are using a Hyperliquid API/agent wallet key.
      ```
  - Then also quietly hide the unhandled error? i think the error is valid but it does not needed to be showing on the logs
  - Because of the `unhandled error`, user cant `connect {hyperliquid, hyperliquid_perpetual}` anymore
    -  The  `ctrl+x` does not work and have to relaunch hummingbot 

---

All invalid tests at the same time successfully created a connector.yml file which resulted to the connector assuming that the API keys were added successfully which is not. ❌

image

  - Expected behavior is should not save it to a e.g `./conf/connector/hyperliquid.yml`

**Comment by ipezygj** [comment]
Thanks for the thorough testing @rapcmia — all three findings are addressed in `aa3e55221`:

1. **API wallet error wording** — the error you saw in the api_wallet test now speaks of the API wallet, matching the setup flow's terminology: *"The supplied private key is not a valid API wallet key for account 0x…: its address 0x… is not in the account's approved API (agent) wallet list and is not the account owner. Approve this API wallet at https://app.hyperliquid.xyz/API, verify you supplied the correct API wallet key, or use the account owner's key with the arb_wallet connection mode."* Applied to both the spot and perpetual connectors.

2. **arb_wallet unhandled error / stuck client** — found the root cause: the key/address mismatch is raised while the connector object is being constructed, and that exception escaped `UserBalances.add_exchange` uncaught, which is why you got a raw traceback and a dead prompt. `add_exchange` now catches connector-construction errors and returns them through the normal connection-error path, so the flow prints a single clean message — including the guidance line you asked for ("Verify the private key matches the wallet address you intend to trade from, set use_vault=True if the supplied address is a vault address, or select the api_wallet connection mode if you are using a Hyperliquid API/agent wallet key.") — and returns to a usable prompt. No traceback in the logs, no relaunch needed, and `connect hyperliquid` works again immediately.

3. **Config file created despite failed validation** — on a fresh connect the config is now rolled back when validation fails: `Security.remove_secure_config` deletes the just-written `conf/connectors/hyperliquid.yml`, so the client no longer reports keys as added. When replacing existing keys, the previous working config is restored as before.

Added regression tests for both connect-flow paths (fresh connect rolls back, replace-keys restores) and tightened the message assertions in both Hyperliquid key-authority suites. Happy to adjust the wording further if you'd like.

**Comment by rapcmia** [comment]
image

- Tested on both arb_wallet and api_wallet with invalid wallet and pkey ok
- Invalid connector does not automatically save into `./conf/connectors` ok

**Review by rapcmia — APPROVED** [review]

File: hummingbot/user/user_balances.py [hunk 1]
```diff
@@ -71,7 +71,15 @@ async def add_exchange(self, exchange, client_config_map: ClientConfigMap, **api
...
         if not is_gateway_market:
-            market = UserBalances.connect_market(exchange, client_config_map, **api_details)
+            try:
+                market = UserBalances.connect_market(exchange, client_config_map, **api_details)
+            except Exception as e:
+                # Connector construction can fail credential validation (e.g. Hyperliquid rejects a
+                # private key that does not derive to the supplied wallet address). Surface it as a
+                # normal connection error message instead of an unhandled exception that leaves the
+                # client prompt unusable.
+                logging.getLogger().debug(f"Failed to create connector for {exchange}", exc_info=True)
```
cardosofede: this is not the standard logger

cardosofede (approved):
just update the logger

cardosofede:
thanks for the contribution @ipezygj !

ipezygj:
Thanks @cardosofede, @rapcmia — fixed in 5c7c5ff.

`UserBalances` now has the standard class-level `logger()` accessor and both debug calls go through it:

```python
class UserBalances:
    __instance = None
    _logger: Optional[HummingbotLogger] = None

    @classmethod
    def logger(cls) -> HummingbotLogger:
        if cls._logger is None:
            cls._logger = logging.getLogger(__name__)
        return cls._logger
```

One line beyond your comment: the pre-existing `logging.getLogger().debug(...)` in `_update_balances` was the call I had copied, so I moved it over too rather than leave the same class using two conventions. Happy to revert that line if you'd rather keep the diff to the new code.

`test/hummingbot/client/command/test_connect_command.py` — 7 tests, all passing.

cardosofede (approved):

PR #8222, opened 2026-05-12
fix / exponential backoff for MQTT autostart reconnect (#8012)

### LINKED ISSUE #8012: Client (headless) -W MQTT bridge continuously logs "Reconnecting in 5.0 seconds"
### Describe the bug

When testing https://github.com/hummingbot/hummingbot/pull/7986 and use hummingbot in headless mode, the mqtt bridge repeatedly prints reconnect logs every few seconds, creating logs spam making difficult to monitor/inspect

Image

### Steps to reproduce

1. Run `./bin/hummingbot_quickstart.py controller-config.yml --headless true`

### Release version

dev-2.13.0

### Type of installation

Source

### Attach required files

n/a

ImMike:
Fixes #8012

## Summary

When `mqtt_autostart=true` and the broker is unreachable, the retry loop in `MQTTCommand.start_mqtt_async` reattempts every 10s forever, producing a continuous stream of identical error logs and notifications. In headless mode this drowns out other output (see the screenshot on #8012).

## Change

Replace the fixed retry sleep with **exponential backoff**, doubling each attempt and capped at a new class attribute `_mqtt_max_sleep_rate_autostart_retry` (default `300.0` seconds = 5 min).

- First retry: still 10s (unchanged default behavior).
- Spam profile over 1 hour: ~360 logs ➜ ~15 logs.
- Infinite-retry semantics preserved — this only changes the cadence, not the give-up policy.
- Backwards compatible: the existing `_mqtt_sleep_rate_autostart_retry` knob still controls the initial wait.

## Test

- Existing `test_start_mqtt_command_retries_with_autostart` still passes (first retry log unchanged).
- New `test_start_mqtt_command_uses_exponential_backoff` verifies the advertised sleep interval doubles each attempt (0.1 ➜ 0.2 ➜ 0.4) and saturates at the configured cap.

## Notes

This does not fix the underlying connectivity failure (broker config / config-load race condition mentioned in the same issue) — those are separate concerns. It does substantially reduce the log spam complained about in the report.


cardosofede:
seems like the issue is in commlib library that does not have management for exception

PR #8225, opened 2026-05-13
fix(xrpl): carry custom markets through startup config

dhairyajangir:

  This PR carries the XRPL `custom_markets` configuration through connector startup so that custom market definitions
  are available when the exchange is initialized.

  Changes:
  - `XrplExchange` now uses the connector configuration when available.
  - `XRPLConfigMap` forwards the connector configuration into the exchange startup path.
  - Added regression coverage for resolving a custom XRPL market from connector configuration during startup.

  - `python -m py_compile hummingbot/connector/exchange/xrpl/xrpl_exchange.py hummingbot/connector/exchange/xrpl/
  xrpl_utils.py test/hummingbot/connector/exchange/xrpl/test_xrpl_utils.py`

  **Tips for QA testing**:

  - Configure XRPL with a custom market such as `BBRL-RLUSD`.
  - Start the connector from the config path used by Hummingbot.
  - Verify the custom market resolves correctly at startup and is available for trading pair lookups.
  - Confirm the behavior matches the original XRPL startup regression repro.

### DIFF
--- a/hummingbot/connector/exchange/xrpl/xrpl_exchange.py
+++ b/hummingbot/connector/exchange/xrpl/xrpl_exchange.py
@@ -62,6 +62,7 @@ from hummingbot.connector.exchange.xrpl.xrpl_utils import (  # AddLiquidityReque
     PoolInfo,
     QuoteLiquidityResponse,
     RemoveLiquidityResponse,
+    XRPLConfigMap,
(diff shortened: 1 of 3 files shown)


cardosofede:
this was already fixed in the previous release

Issue #7953, opened 2026-01-16
Bybit - order_book bid and ask size not matched with exchange


nikspz:
### Describe the bug

Bybit spot - order_book bid and ask size not matched with exchange

### Steps to reproduce

1. connect bybit
2. create v1 pureMM strategy 
3. start
4. order_book --live
5. review bid and ask size not matched with exchange

### Release version

dev-2.12.0

### Type of installation

Source

rapcmia:
Bybit's bid and ask quantities through without converting or scaling them, and its order-book tests cover the raw values. We could not reproduce the mismatch from the current report because it does not include the trading pair, timestamp, or the differing price and size levels

PR #8226, opened 2026-05-14
(fix) Bybit WS order book snapshot timestamp scaling

ipezygj:

- [x] Tests all pass

In `BybitAPIOrderBookDataSource._process_ob_snapshot` the websocket snapshot path passed `json_msg["ts"]` directly as the `OrderBookMessage` timestamp. Bybit V5 delivers all timestamps in milliseconds, and the three other paths in this file already convert to seconds:

| Path | Conversion |
|---|---|
| `_order_book_snapshot` (REST) | `float(snapshot["ts"]) * 1e-3` |
| `_parse_order_book_diff_message` (WS diff) | `raw_message["ts"] * 1e-3` |
| `_take_full_order_book_snapshot` (REST refresh) | `float(snapshot["ts"]) * 1e-3` |
| `_process_ob_snapshot` (WS snapshot) | `json_msg["ts"]` ← missing `* 1e-3` |

The websocket snapshot path was the only one missing the `1e-3` scaling, so WS-pushed snapshots ended up with timestamps roughly three orders of magnitude larger than every other source. This breaks ordering against REST snapshots (which run hourly) and any downstream consumer that mixes WS snapshot timestamps with WS diff or REST snapshot timestamps.

The fix applies the same `* 1e-3` scaling already used by the other three paths.

This may be related to #7953 (Bybit order book bid/ask size not matching the exchange UI), since stale or out-of-order snapshots can cause the local book to drift from the exchange state. Filing as a standalone consistency fix rather than claiming a definitive close, since reproducing #7953 requires comparing live exchange data against the local book.

The existing `test_listen_for_order_book_snapshots_successful_ws` test in `test_bybit_api_order_book_data_source.py` did not actually exercise the websocket snapshot path:

- it routed the snapshot-shaped message into the `order_book_diff` queue and ran `listen_for_order_book_diffs`, so it exercised the diff code path instead of `_process_ob_snapshot`
- the bid/ask fixture contained `Ellipsis` (`...`) entries mixed with real `[price, size]` rows, which was an accidental placeholder rather than valid order book data
- it asserted only `update_id` with `assertTrue`, so it would have passed for any truthy value

The rewritten test feeds the snapshot event into the `order_book_snapshot` queue, calls `_process_ob_snapshot` directly, and asserts both `update_id` and the scaled timestamp (`ts * 1e-3 == msg.timestamp`).

**Tips for QA testing**:

1. Subscribe to a Bybit spot order book in Hummingbot (e.g. `BTC-USDT`).
2. After receiving WS snapshots, inspect any logged or downstream-consumed `OrderBookMessage.timestamp` for that pair. Before this PR the value is on the order of `1.7e12` (milliseconds-as-seconds); after this PR it should be a normal Unix timestamp in seconds (`~1.7e9`).
3. The REST snapshot path (taken on startup and every hour) and the WS diff path are unchanged.

### DIFF
--- a/hummingbot/connector/exchange/bybit/bybit_api_order_book_data_source.py
+++ b/hummingbot/connector/exchange/bybit/bybit_api_order_book_data_source.py
@@ -234,7 +234,7 @@ class BybitAPIOrderBookDataSource(OrderBookTrackerDataSource):
                 trading_pair = await self._connector.trading_pair_associated_to_exchange_symbol(
                     symbol=data["s"])
                 order_book_message: OrderBookMessage = BybitOrderBook.snapshot_message_from_exchange_websocket(
-                    data, json_msg["ts"], {"trading_pair": trading_pair})
+                    data, json_msg["ts"] * 1e-3, {"trading_pair": trading_pair})
                 snapshot_queue.put_nowait(order_book_message)
             except asyncio.CancelledError:
                 raise
--- a/test/hummingbot/connector/exchange/bybit/test_bybit_api_order_book_data_source.py
+++ b/test/hummingbot/connector/exchange/bybit/test_bybit_api_order_book_data_source.py
@@ -585,26 +585,12 @@ class TestBybitAPIOrderBookDataSource(IsolatedAsyncioWrapperTestCase):
             "data": {
                 "s": f"{self.ex_trading_pair}",
                 "b": [
-                    ...,
-                    [
-                        "16493.50",
-                        "0.006"
-                    ],
-                    [
-                        "16493.00",
-                        "0.100"
-                    ]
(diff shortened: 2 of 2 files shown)


ipezygj:
For reviewer context: this is part of a small series of order-book-snapshot timestamp-scaling fixes (exchange `ts`/`time` returned in ms, `OrderBookMessage.timestamp` expected in seconds, REST/WS path inconsistency). They're independent but near-identical in shape and can be reviewed together:

- #8226 — Bybit spot WS snapshot (this PR)
- #8229 — Bybit Perpetual REST snapshot
- #8230 — OKX Perpetual (all paths)
- #8231 — Hyperliquid REST snapshot (spot + perpetual)

PR #8291, opened 2026-06-11
(fix/perf) binance_perpetual: connector-review correctness & rate-limit fixes

cardosofede:
## What

Correctness and rate-limit fixes for the `binance_perpetual` connector, found by auditing the connector against the official Binance USDⓈ-M Futures API docs. Each fix is small, isolated, and covered by tests.

### Correctness
- **Order book diff sequencing** — `@depth` diffs were applied unconditionally with no `pu`/`u` chain validation; a dropped WS message left the local book corrupt until the hourly snapshot reset. Now the connector validates that each diff's `pu` matches the previously applied `u`, drops the gapped diff, and forces an immediate REST snapshot resync of the affected pair (served on-demand by `listen_for_order_book_snapshots` alongside the hourly reset). Per the official *"How to manage a local order book correctly"* guide.
- **WS available balance** — the `ACCOUNT_UPDATE` handler derived available balance from `cw` (Cross Wallet Balance), which does not subtract margin used by open positions/orders, overstating available funds between REST polls. The event exposes no true available balance, so available is now left to the REST poll (`_update_balances`, which uses `availableBalance`).
- **Listen-key loop** — a bare `raise` in the listen-key refresh-failure branch raised a misleading `RuntimeError("No active exception to re-raise")`; replaced with an explicit `IOError`. Behaviour (reset key + retry) unchanged.

### Performance / rate limits
- **`userTrades` by `orderId`** — `_all_trade_updates_for_order` fetched all recent trades for a symbol and filtered client-side; it now passes `orderId` so the exchange returns only that order's fills (no more missed fills past the 500-trade window).
- **`premiumIndex` auth** — `_request_complete_funding_info` signed a public endpoint; auth removed (consistent with `_fetch_last_fee_payment`).
- **`ORDER_URL` rate limit per verb** — POST/GET/DELETE shared a single `RateLimit` linked to the order-count pools, so status polling burned the order-creation quota. Split into per-verb limit ids; only the POST (create) consumes `ORDERS_1MIN`/`ORDERS_1SEC`.
- **Order-fills polling window** — `_update_order_fills_from_trades` downloaded every `userTrades` page per symbol each tick; it now bounds the query with `startTime` based on the previous poll (overlapping windows to avoid missing fills).

## Testing
- Full `binance_perpetual` connector suite green: **108 passed**.
- New unit tests added for each fix (diff sequence gap → resync, `first_update_id` propagation, on-demand snapshot serving, `orderId` filtering, per-verb rate limits, `startTime` windowing, balance handling, listen-key error).
- `pre-commit` (flake8 / isort / autopep8) clean on all touched files.

## Notes
- Each fix carries its audit/spec card under `improvements/binance_perpetual/done/` for traceability (problem → solution → acceptance criteria → commit).
- **Known pending (not in this PR):** the user-stream `listenKey` is passed as a query param (`/private/ws?listenKey=`) while the docs document the path form (`/private/ws/<listenKey>`). Left out because it needs validation against testnet before changing — tracked as `CORR-002`.


nikspz (approved):
- commit 2ec147a
    - cloned and installed PR8291 successfully ✅
    - connected binance_perpetual successfully ✅
    - checked balance matched with exchange ✅
    - created/started perpetualMM strategy successfully ✅
    - placed / refreshed orders successfully ✅
        - Placed / filled BUY limit order successfully
        - Placed / filled SELL limit order successfully
        - Placed / filled MARKET order successfully
    - Order Status Updates: ok ✅
    - Partial Fill Handling ✅
        - Every fill detected.
        - Correct filled quantity.
        - Correct average price.
        - No duplicate fills.
    - status, order_book, ticker —live matched with exchange ✅
    - Balance Update After Trade: ok ✅
    - Reconnect Test: ok, Trading continues normally ✅
    - Longrun: was able to run v2 strategy over the weekend successfully✅
    - Long-running Stability
        - No order book freezes.
        - No repeated warnings/errors.
        - No user stream disconnect loop.

PR #8265, opened 2026-05-29
feat: add builder code support to Hyperliquid connectors (HGP-87)

fengtality:
## Summary

Implements the reference builder-code track from **HGP-87** for both Hyperliquid connectors: `hyperliquid` (spot) and `hyperliquid_perpetual` (perp). Every mainnet order carries a Foundation builder attribution at a zero fee rate (attribution only); the field is omitted on testnet and vault orders. No approval action is invoked from the connector.

The work is additive — no structural changes to either connector.

## What's included

**Constants** (`hyperliquid_constants.py`, `hyperliquid_perpetual_constants.py`)
- `BUILDER_SUPPORTED`, `FOUNDATION_BUILDER_ADDRESS`, `FOUNDATION_BUILDER_FEE_TENTHS_BPS`
- Protocol fee caps in tenths-of-a-bp: perp `100` (= 10 bps), spot `1000` (= 100 bps)
- `FOUNDATION_BUILDER_ADDRESS` is set to the onboarded Foundation builder `0x10BA451e6439Efc6a17dc20d21121Aa838100705` at `0` bps, so every mainnet order is attributed by default.

**Order payload injection**
- `_place_order` attaches the builder field `{"b": <lowercase_address>, "f": <tenths_of_bps>}` to the order action.
- The auth classes (`_sign_order_params`) carry the field into the signed L1 action **before** EIP-712 signing, so the signature covers it. The address is lowercased — Hyperliquid rejects mixed-case addresses in the builder field.

**Omit invariants** (`_should_inject_builder`) — cannot be overridden by config:
- Vault orders (venue rejects on wallet-identity mismatch)
- Testnet (stricter check, rejects regardless of fee value)

**Override validation** (`_load_builder_override`)
- Validates an application-provided `fee_bps` against the per-venue protocol cap and raises `ValueError` at load time; lowercases the address.

**`/builder-info` handler** (`get_builder_info`)
- Returns `{"supported": False}` on testnet / vault / unconfigured.
- Otherwise queries `maxBuilderFee` and returns `{supported, builder_address, fee_bps, approved, approval_expiry_ms}`. At the 0-bps default `approved` is trivially `True` (approved max ≥ 0 for any user); `approval_expiry_ms` is `None` (Hyperliquid approvals do not expire).

Connect flow is unchanged — users provide existing credentials and the connector never invokes `ApproveBuilderFee`.

## Tests

[...]

PR #8292, opened 2026-06-11
fix(ci): stop push-event runs failing on get-diff-action ref resolution

fengtality:
## Problem

The post-merge `push` CI runs on `development`/`master` go red even when the PR that was merged passed all checks. Example: [run 27359558988](https://github.com/hummingbot/hummingbot/actions/runs/27359558988) (the merge of #8265).

The `Check if client files changed` gate fails at the `technote-space/get-diff-action@v6` step:

```
command [git diff] exited with code 128.
fatal: ambiguous argument 'get-diff-action/staging...get-diff-action/pull/8290/merge':
unknown revision or path not in the working tree.
```

`get-diff-action` diffs against a `pull/<n>/merge` ref. That ref only exists for **`pull_request`** events — on a **`push`** event (the merge commit landing on the branch) it's absent, so `git diff` exits 128. That fails the gate job, and because `build_hummingbot` has `needs: run_client`, the build + stable tests job is **skipped**. Re-running doesn't help — the failure is deterministic, not flaky.

## Fix

Only run the file-change filter for `pull_request` events; on `push`, always set `is_set=true` so the post-merge run exercises the full suite.

- **PR events:** unchanged — `get-diff-action` still gates the build so it's skipped when no `py/pyx/pyd/yml` files changed.
- **Push events:** skip `get-diff-action` entirely (no merge ref to resolve) and build unconditionally.

## Verification

- YAML parses cleanly; pre-commit hooks pass.
- Logic: on push, the `get-diff-action` step is skipped, `check_files` runs via `github.event_name == 'push'`, sets `is_set=true`, gate passes, build runs. On PR, behaviour is identical to today.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -16,15 +16,19 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
+      # get-diff-action diffs against a pull/<n>/merge ref that only exists for pull_request
+      # events; on a push (e.g. post-merge to master/development) that ref is absent and the
+      # action exits 128, failing this gate and skipping the build. Only run the file-change
+      # filter for PRs; on push always build, since post-merge runs should exercise the suite.
       - uses: technote-space/get-diff-action@v6
+        if: github.event_name == 'pull_request'
         with:
           PATTERNS: |
             **/*.+(py|pyx|pyd|yml)
       - name: Check if client files are modified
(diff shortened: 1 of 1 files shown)


fengtality:
Merging this in to test if it fixes CI/CD issues in https://github.com/hummingbot/hummingbot/pull/8290

PR #8310, opened 2026-06-18
Fix/ fix LP executor get_net_pnl_pct return value to conform with other executor

mlguys:

- [x] Tests all pass

- Fixed LP executor `net_pnl_pct` to return a ratio instead of whole percentage points, conforming with other executor return types.


cardosofede (approved):

rapcmia (approved):
- Tested LP executor PnL and PnL% against Orca portfolio values.
- Initial issue: PnL amount was correct, but PnL% did not match Orca.
- Condor displayed PnL: -$0.01 and PnL%: -0.17%.
- Orca also showed Total PnL: -$0.01 and -0.17%.
- PnL% is now aligned with Orca and appears to be calculated/displayed correctly.

PR #8313, opened 2026-06-19
fix: correct kucoin_perpetual trade stream WebSocket topic and timestamp field

### LINKED ISSUE #7482: kucoin_perpetual: missing trade updates
### Describe the bug

The discussion started [here](https://discord.com/channels/530578568154054663/557343500908363786/1343588621760073818)

I realized that I don't receive any trade updates with **kucoin_perpetual**.

Then I found the error and [fixed it](https://github.com/hummingbot/hummingbot/pull/7463), but my PR did not pass the automatic tests.

Unfortunately, I don't have time to write all the tests now, so I closed the PR.

### Steps to reproduce

1. Connect to kucoin_perpetual
2. Run the `download_order_book_and_trades.py` script
3. Analyze (un-)received trades

4. Apply fixes from [my PR](https://github.com/hummingbot/hummingbot/pull/7463)
5. Run the script again
6. Compare trades with kucoin web-site.

### Release version

2.3.0, 2.4.0

### Type of installation

Source

### Attach required files

_No response_

ipezygj:

- [x] You are using approved title (feat/, fix/, docs/, refactor/)

The kucoin_perpetual connector was not receiving any trade updates via WebSocket because:

1. **Wrong topic**: `WS_TRADES_TOPIC` was set to `/contractMarket/tradeOrders` (deprecated/non-existent for public trades). The correct topic for public trade execution data is `/contractMarket/execution`.

2. **Wrong timestamp field**: The trade message parser referenced `data[time]` which does not exist in the `/contractMarket/execution` payload. The correct field is `data[ts]` (nanosecond timestamp).

3. **Hardcoded topic strings**: The `_subscribe_channels` method used hardcoded `/contractMarket/ticker` and `/contractMarket/level2` strings instead of the defined `CONSTANTS.WS_TRADES_TOPIC` and `CONSTANTS.WS_ORDER_BOOK_EVENTS_TOPIC` references.

**Files changed:**
- `kucoin_perpetual_constants.py` — fix `WS_TRADES_TOPIC` value
- `kucoin_perpetual_api_order_book_data_source.py` — use CONSTANTS refs + fix timestamp field
- `test_kucoin_perpetual_api_order_book_data_source.py` — update test assertions to match

- Updated unit tests to reflect the corrected topic and timestamp field
- Verified the fix matches the approach in the previously-validated PR #7463 (which was closed only due to missing test updates)

**Related issue**: Fixes #7482

**Tips for QA testing**:

Use the `download_order_book_and_trades.py` script (attached in issue #7482) to verify trades are received. Compare received trades with the KuCoin Futures web interface for validation.


ipezygj:
CI failure is **unrelated to this PR's changes**. 

The failing test `test_cancel_two_orders_with_cancel_all_and_one_fails` is a known flaky test with a timeout race condition. I've submitted #8318 to fix it.

The failure occurs because the test wrapper `async_run_with_timeout()` defaults to 1s timeout while `cancel_all(10)` needs 10s to complete, causing random `CancelledError` → `TimeoutError`.

This PR's code changes are correct. Once #8318 is merged, a rebase will pull in the fix.

ipezygj:
Updated this PR to tighten scope and complete the fix:

- **Removed the two unrelated Hyperliquid test tweaks.** Increasing the `cancel_all` timeout addresses a flaky test that is independent of this change; that fix belongs in #8318. This PR is now scoped strictly to `kucoin_perpetual`.
- **Extended the fix to the dynamic per-pair paths.** `subscribe_to_trading_pair` / `unsubscribe_from_trading_pair` still hardcoded `/contractMarket/ticker` for the trade feed, so any pair added after startup received no public trade prints (same root cause as `_subscribe_channels`). They now use `WS_TRADES_TOPIC` (`/contractMarket/execution`) and `WS_ORDER_BOOK_EVENTS_TOPIC`.
- **Added regression tests for #7482:** `_channel_originating_message` routes the execution topic to the trade queue (and never routes the ticker topic there), plus an assertion that the per-pair subscription actually uses the execution topic.

Summary of the full fix now in this PR:
1. Public trade feed subscribed to `/contractMarket/execution` instead of `/contractMarket/ticker` (initial + per-pair paths).
2. Trade timestamp read from `ts` (execution feed) instead of `time`.
3. `WS_TRADES_TOPIC` corrected to `/contractMarket/execution`; the private order-change subscription is preserved via `WS_ORDER_CHANGE_TOPIC` so the user stream is unaffected.

Note: I am aware of #8341, which targets the same root cause. This PR predates it (opened Jun 19) and additionally covers the per-pair subscribe/unsubscribe paths and preserves the private order-change subscription. Happy to consolidate if the maintainers prefer one.

wuxusen:
Read through this end-to-end against `development` — the fix is correct. Public trades now go to `/contractMarket/execution` in `_subscribe_channels` and both per-pair `subscribe_to_trading_pair` / `unsubscribe_from_trading_pair` paths, `_channel_originating_message` picks them up via the repurposed `WS_TRADES_TOPIC`, and reading the timestamp from `ts` matches the execution payload. The private order-change stream stays intact on `WS_ORDER_CHANGE_TOPIC` (same `/contractMarket/tradeOrders`).

Two small non-blocking things you might tidy while you're in here:

1. `constants.py` already had `WS_EXECUTION_DATA_TOPIC = "/contractMarket/execution"` sitting unused. Using that existing constant for the public trade subscription/routing — instead of repurposing `WS_TRADES_TOPIC` — would keep the change inside the order-book source and avoid having to touch the user-stream file. As it is now, `WS_TRADES_TOPIC` and `WS_EXECUTION_DATA_TOPIC` are two constants holding the same value.
2. In `test_listen_for_trades_successful` the sample message still uses the spot envelope (`topic: /market/match:…`, `subject: trade.l3match`) while `data` now carries the futures `ts`. The parser only reads `data`, so it passes, but switching the topic/subject to `/contractMarket/execution` / `match` would make it a faithful execution-feed regression.

Neither blocks merge. Good catch on the root cause.

ipezygj:
Thanks for the thorough review @wuxusen — applied both tidy-ups in the latest commit:

1. Public trade subscription/routing now uses the existing `WS_EXECUTION_DATA_TOPIC` constant instead of repurposing `WS_TRADES_TOPIC`. The change stays inside the order-book source and the user-stream file is untouched — the private order-change stream keeps using `WS_TRADES_TOPIC` (`/contractMarket/tradeOrders`, its original value). `WS_ORDER_CHANGE_TOPIC` is dropped since the two constants no longer hold duplicate values.
2. `test_listen_for_trades_successful` now uses a faithful execution-feed envelope (`/contractMarket/execution` topic, `match` subject) instead of the spot `trade.l3match` envelope.

The topic assertions and #7482 regression tests were repointed at `WS_EXECUTION_DATA_TOPIC` accordingly.

ipezygj:
Gentle nudge here when a maintainer has bandwidth. CI is green, @wuxusen reviewed the change end-to-end against `development` and confirmed it, both of his suggested tidy-ups are applied in the latest commit, and the competing #8341 was closed in favor of this PR. Happy to rebase if needed.

ipezygj:
Pushed `22ccaf655` hardening the corrected feed against reconnect distortion, per review feedback.

A websocket reconnect can replay recent executions, and the order-book snapshot taken on re-subscribe overlaps the live stream — so the same match can arrive twice. With the feed now correctly subscribed, processing that overlap twice would double-count the trade and distort anything built from it. The data source now tracks the last execution `sequence` emitted per trading pair and skips any trade at or below it. Since KuCoin's per-symbol execution sequence is monotonic, this drops both duplicates (reconnect replay) and stale/out-of-order matches, and keeps the emitted trade timestamps monotonically non-decreasing.

Added a reconnect regression (`test_parse_trade_message_dedupes_and_orders_on_reconnect`) that feeds an overlapping replay plus a late/out-of-order match and asserts each match is emitted exactly once and the timestamps stay ordered. flake8 clean; new test green locally.

nikspz (commented):
- Test performed:
    - Cloned and installed PR8313
    - connected kucoin_perpetual
    - checked balance successfully
    - created/started perpetualMM strategy using kucoin_perpetual
        - Public market data subscription succeeds
        - Order book initializes correctly
        - Trade messages are now being received
    - Before this PR (issue #7482), trade messages would typically remain at **0** because the connector subscribed to the wrong topic.
        Now:
        - ✅ Trade messages processed
        - ✅ No rejected messages
        - ✅ Parser successfully handled incoming trades
    - Verify order book still works
        - order_book —live matched with exchange
    - Order submission and cancellation: ok
    - Filled events: OPEN and CLOSE: ok
    - History: ok
    - ✅ Public WebSocket connected successfully
    - ✅ Order book initialized correctly
    - ✅ Order book diffs continue processing (`23030`)
    - ✅ **Trade messages processed (`10`)** — this directly validates the bug fix
    - ✅ No trade parsing errors
    - ✅ No `KeyError('time')`
    - ✅ No WebSocket subscription failures

nikspz (approved):

ipezygj:
Glad this landed and closed out #7482 — trade streams on kucoin_perpetual now flow correctly. Out of interest: does Hummingbot have a contributor-rewards or bounty process a fix like this could qualify for, or is that limited to pre-assigned board issues? Either way I'm happy to pick up assigned bounty issues going forward — point me at the board and I'll take some on.

PR #8316, opened 2026-06-19
fix: validate dca_amounts sum is non-zero in dman_maker_v2 (#8109)

### LINKED ISSUE #8109: [Bug] ZeroDivisionError in DMAN Controller when DCA parameters are set to 0 - Causes Infinite Loop
## Description
While using the `dman_maker_v2` controller on Hyperliquid, setting DCA-related parameters to `0` or leaving them effectively at zero causes a `ZeroDivisionError`. This error enters an infinite loop within the strategy tick, freezing the Hummingbot client entirely.

## Steps to Reproduce
1. Use a V2 Strategy with `dman_maker_v2` controller.
2. In the controller YAML config, set DCA spreads or amounts to `0`.
3. Run the script.
4. As soon as the controller attempts to calculate the next DCA level or position rebalance, the client freezes.

## Expected Behavior
The controller should either:
- Validation: Prevent the script from starting if parameters lead to a division by zero.
- Logic: Skip DCA calculations if the user intends to disable it by setting values to 0.

## Actual Behavior
The bot enters a non-responsive state. Commands like `stop` or `exit` do not work. The only way to terminate is via `kill -9 [PID]`, which leaves open orders on the exchange.

## Environment
- Hummingbot Version: [Your Version]
- Controller: dman_maker_v2
- Connector: hyperliquid_perpetual

### Steps to reproduce

1.
2.
3.

### Release version

2.13.0

### Type of installation

Source

### Attach required files

_No response_

ipezygj:

- [x] Tests all pass

Fixes #8109. When all `dca_amounts` are set to `0` in the `dman_maker_v2` controller config, `sum(self.config.dca_amounts)` evaluates to `0` in `DManMakerV2.__init__()` line 71, causing a `ZeroDivisionError` that enters an infinite loop within the strategy tick and freezes the client.

**Fix**: Add validation in `parse_and_validate_dca_amounts` to reject configs where the sum of `dca_amounts` is <= 0, raising a clear `ValueError` at config load time.

**Files changed**:

| File | Change |
|---|---|
| `controllers/market_making/dman_maker_v2.py` | Add sum validation in `parse_and_validate_dca_amounts` |

- `py_compile` clean
- Verified that `DManMakerV2Config(dca_amounts="0,0,0,0", dca_spreads="0.01,0.02,0.04,0.08")` now raises `ValueError: The sum of dca_amounts must be greater than zero.`
- Verified that valid configs like `"0.1,0.2,0.4,0.8"` still pass validation

**Tips for QA testing**:

1. Set `dca_amounts` to all zeros in a dman_maker_v2 config → should get clear validation error
2. Set `dca_amounts` to valid values → should work as before
3. Leave `dca_amounts` empty/None → should default to `[1,1,1,1]` as before

### DIFF
--- a/controllers/market_making/dman_maker_v2.py
+++ b/controllers/market_making/dman_maker_v2.py
@@ -57,10 +57,13 @@ class DManMakerV2Config(MarketMakingControllerConfigBase):
         if v is None or v == "":
             return [1 for _ in validation_info.data['dca_spreads']]
         if isinstance(v, str):
-            return [float(x.strip()) for x in v.split(',')]
-        elif isinstance(v, list) and len(v) != len(validation_info.data['dca_spreads']):
+            v = [float(x.strip()) for x in v.split(',')]
+        if isinstance(v, list) and len(v) != len(validation_info.data['dca_spreads']):
             raise ValueError(
                 f"The number of dca amounts must match the number of {validation_info.data['dca_spreads']}.")
+        if isinstance(v, list) and sum(v) <= 0:
+            raise ValueError(
+                "The sum of dca_amounts must be greater than zero.")
         return v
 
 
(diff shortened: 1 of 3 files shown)


ipezygj:
CI failure is **unrelated to this PR's changes**. 

The failing test `test_cancel_two_orders_with_cancel_all_and_one_fails` is a known flaky test with a timeout race condition. I've submitted #8318 to fix it.

The failure occurs because the test wrapper `async_run_with_timeout()` defaults to 1s timeout while `cancel_all(10)` needs 10s to complete, causing random `CancelledError` → `TimeoutError`.

This PR's code changes are correct. Once #8318 is merged, a rebase will pull in the fix.

PR #8328, opened 2026-06-28
fix / rate oracle treat USD as USDT-equivalent for balance conversions

fengtality:
## Problem

The `balance` command showed `$0` totals when an asset was priced against a rate source that quotes in USDT (e.g. the default `gate_io` source) but the asset/quote symbol was `USD`:

- **USD-quoted collateral** (e.g. `hyperliquid_perpetual` shows a `USD` balance) failed to convert.
- **Single-market tokens** like `HYPE` (gate.io only lists `HYPE-USDT`) returned `None` → `$0`, even with `USD` mapped to USDT, because `find_rate` skipped the direct pair after normalizing the quote and fell through to bridge logic that built the non-existent `USDT-USDT` pair.

## Root causes

1. `find_rate` did not re-check the direct pair after normalizing the quote (`USD → USDT`), so `HYPE-USD` never matched `HYPE-USDT`.
2. The `USD → USDT` equivalence list was hardcoded (`["USC"]`) and not user-configurable, and the surrounding helper carried brittle, hand-maintained wrapped-token regexes (`W…`, `w…`, `.e`) in a misnamed `core/gateway/utils.py` file unrelated to gateway.

## Changes

- **`find_rate`** (`core/rate_oracle/utils.py`): re-check the normalized direct pair before reverse/bridge lookups. Fixes single-market tokens and yields direct (more accurate) prices instead of multi-hop bridge prices.
- **Token normalization**: replaced the brittle `unwrap_token_symbol` (W/w/.e regex handling + `USC` default) with a simple `normalize_token_symbol` that maps USD-equivalent symbols to USDT. Renamed since it no longer "unwraps" wrapped tokens.
- **Removed `core/gateway/utils.py`**: it was unrelated to gateway and consumed only by `find_rate`. Moved `USD_EQUIVALENT_TOKENS` + the helper into `core/rate_oracle/utils.py` next to its only caller and deleted the file (imports repointed, no compat shim).
- **New config field** `global_token.usd_equivalent_tokens` (default `["USD"]`), user-settable via `config global_token` or `conf_client.yml`, wired into the rate-oracle utils list at startup alongside the quote token. Accepts comma-delimited input and uppercases symbols.
- **Regression test** covering USD-quoted lookups and single-market tokens.

### Automated
```bash
# from repo root, in the hummingbot conda env
pytest test/hummingbot/core/rate_oracle/test_rate_oracle.py -v
```
All `test_find_rate*` cases should pass, including `test_find_rate_unwraps_usd_quote_to_usdt` and `test_find_rate_usd_equivalence_is_a_fallback_not_an_override`.

### Manual (reproduces the original report)
Prereq: at least one account with a USD-denominated balance — e.g. a `hyperliquid_perpetual` connector (collateral shows as `USD`) and/or a token that only lists against USDT on the rate source (e.g. `HYPE`).

1. Start the Hummingbot client on this branch.
2. Set a USDT-quoting rate source and the USD global token:
   ```
   config rate_oracle_source gate_io
   config global_token USD
   ```
3. Run `balance`.
   - **Before this PR:** `Total ($)` is `0` for every asset; `hyperliquid_perpetual` `USD` and `HYPE` show `0`.
   - **After this PR:** every asset shows a non-zero `Total ($)`, including the `USD` collateral and `HYPE`.
4. (Optional) Confirm it also works with `config global_token USDT` — totals should match.
5. (Optional) Verify the new config field:
   ```
   config global_token   # shows usd_equivalent_tokens: [USD]
   ```
   Remove `USD` from `usd_equivalent_tokens` (set it to empty) and re-run `balance` with `global_token USD` → USD/HYPE go back to `0`, confirming the list drives the behavior.

### Quick scripted check (no client needed)
```bash
python -c "
import asyncio
from hummingbot.core.rate_oracle.sources.gate_io_rate_source import GateIoRateSource
from hummingbot.core.rate_oracle.utils import find_rate
async def main():
    prices = await GateIoRateSource().get_prices(quote_token='USD')
    for b in ['USD','USDT','USDC','SOL','HYPE']:
        print(f'{b}-USD ->', find_rate(prices, f'{b}-USD'))
asyncio.run(main())
"
# expect non-None values for all (e.g. HYPE-USD ~ 60+, USD/USDT -> 1)
```

## Verification (already run)

- 15 rate-oracle tests + config tests pass.
- Live gate.io check, `global_token = USD`: `HYPE-USD → 62.99`, `SOL-USD → 71.85` (direct), `USD/USDT → 1`, `USDC-USD → 1.001` (de-peg preserved). Same results with `global_token = USDT`.

## Behavior change to note

Wrapped tokens (`WBTC`, `WETH`, `*.e`, …) are no longer unwrapped to their underlying for rate lookups — they now require the rate source to list them directly. This is intentional (the regex allowlist was brittle), but flagged for reviewers.


rapcmia:
Commit 886095479b208b8d5135cd20d3773d898d0d5530
- Passed ✅ 
  - USD balance conversion work with `gate_io` and `global_token: USD`, USD-denominated collateral and `HYPE` no longer show `0`.
     - For this test, i use `rate --pair <pair>` command to test the behavior
  - `usd_equivalent_tokens` is working as intended. Removing it makes the USD fallback disappear, and adding `USD` back restores the values.
  - Hyperliquid USD collateral follows the same config-gated behavior and recovers correctly when `USD` is added back.
  - Comma-delimited `usd_equivalent_tokens` values persist in `conf_client.yml` and reload in uppercase format.
  - The Gate.io script check returned valid values for `USD`, `USDT`, `USDC`, `SOL`, and `HYPE`.
  - My agent help to create a wrapped-token smoke-test script to confirmed that wrapped and `.e` tokens no longer inherit the base token price automatically, but still price correctly when their own direct market exists.
    ```python
    python - <<'PY'
    import sys, types
    
    # `find_rate()` imports a helper from `hummingbot.connector.utils`,
    # and that file imports `hexbytes`. If the local env does not have
    # `hexbytes` installed, this tiny stub is enough for this smoke test
    # because we are not using any real `hexbytes` behavior here.
    m = types.ModuleType('hexbytes')
    
    class HexBytes(bytes):
        pass
    
    m.HexBytes = HexBytes
    sys.modules['hexbytes'] = m
    
    from decimal import Decimal
    from hummingbot.core.rate_oracle.utils import find_rate
    
    # Each test case is:
    # 1. a short label
    # 2. a fake price map that imitates what the rate source would return
    # 3. the pair we want to value
    tests = [
        (
            "wrapped token should NOT fall back to base asset anymore",
            # Only BTC has a market here, not WBTC.
            # Old behavior could unwrap WBTC -> BTC and use this anyway.
            {"BTC-USDT": Decimal("100000")},
            "WBTC-USD",
        ),
        (
            ".e token should NOT fall back to base asset anymore",
            # Only AVAX has a market here, not AVAX.e.
            # Old behavior could unwrap AVAX.e -> AVAX and use this anyway.
            {"AVAX-USDT": Decimal("25")},
            "AVAX.e-USD",
        ),
        (
            "wrapped token should still work when its own market exists",
            # Here WBTC itself has a direct market, so the lookup should succeed.
            {"WBTC-USDT": Decimal("101500")},
            "WBTC-USD",
        ),
        (
            ".e token should still work when its own market exists",
            # Same idea for AVAX.e: direct wrapped-token market exists,
            # so this should return the direct price.
            {"AVAX.e-USDT": Decimal("24.5")},
            "AVAX.e-USD",
        ),
    ]
    
    for label, prices, pair in tests:
        # Run the same pricing helper used by the rate-oracle path.
        result = find_rate(prices, pair)
    
        print(label)
        print(f"  available prices: {prices}")
        print(f"  requested pair: {pair}")
        print(f"  result: {result}")
        print()
    PY
    ```
  - Regression coverage looks good. The targeted pytest file passed.

---

However the `usd_equivalent_tokens` is saved and works correctly, but it is not exposed through the interactive `config global_token` command. ❌ 
  image
   - I still mark this PR as passed for the intended behavior ✅

rapcmia (approved):

PR #8356, opened 2026-07-14
Fix Hyperliquid perpetual order price quantization to exchange tick size

chevoisiatesalvati:
## Summary

`quantize_order_price` on the Hyperliquid perpetual connector rounds to a fixed 6 decimals via a `.5g` float heuristic, ignoring the exchange's tick size. Hyperliquid's `limitPx` rule is: **at most 5 significant figures and at most `MAX_DECIMALS - szDecimals` decimal places** (`MAX_DECIMALS` = 6 for perpetuals; integer prices are always allowed). Six decimals satisfies that only for `szDecimals=0`.

ARB-USD is the clearest case. It carries `szDecimals=1`, so `limitPx` accepts 5 decimals — but a market order at `BestBid * 1.05` off a `0.0901` mid quantizes to `0.094605`, and the exchange rejects it with **"Order has invalid price."**

This change rounds to `min_price_increment` from the trading rules after the 5-significant-figure pass, falling back to the 5-sig-fig result when trading rules aren't loaded yet.

## Validation

Pulled the live universe from `api.hyperliquid.xyz/info` and ran every pair's market-order price (`mid * 1.05` and `mid * 0.95`) through the old and new implementations, checking each result against the `limitPx` rule. `_place_order` sends `"limitPx": float(price)`, so validation is against the float repr that actually reaches the exchange.

| | before | after |
|---|---|---|
| perpetual prices violating `limitPx` | **15 of 232 assets** | **0 of 232** |

The 15: ARB, OP, APT, STRK, DYM, MANTA, ME, MELANIA, PNUT, SAGA, TNSR, USUAL, W, ZETA, ACE. Across a ±20% band around the live mid, 283 of 400 realistic ARB market prices are rejected today.

This is safe because `_format_trading_rules` derives `min_price_increment` from the number of decimals in `markPx`, and for perpetuals that is never *finer* than `szDecimals` allows (verified 0 of 232), so rounding to it can only land on a price the exchange accepts.

## Scope: perpetuals only

The spot connector has the same shape of bug but **must not** take the same fix. On spot `MAX_DECIMALS` is 8, and the invariant above does not hold: 23 of 269 spot pairs report a `markPx` with more decimals than `limitPx` accepts — RIP quotes `0.0000124` against `szDecimals=2`, a 6-decimal limit. Rounding to the derived increment there produces 7- and 8-decimal prices the exchange rejects, taking spot from 9 invalid pairs to 23. The existing `round(..., 6)` happens to clamp to the spot limit for the common `szDecimals=2` case.

Fixing spot requires deriving `min_price_increment` from `szDecimals` rather than from the `markPx` string, in `_format_trading_rules`. That is worth doing, but it is a different change and is left out of this PR.

## Notes

- `(price / tick).quantize(...) * tick` inflates the Decimal scale, turning `Decimal(\"10000\")` into `Decimal(\"10000.0000\")`. The value and the serialized `limitPx` are unchanged, but it alters how prices render in logs and order events. The result is normalized, with a guard against `normalize()` choosing exponent form (`1E+4`) for integers.
- BTC-USD is unaffected either way: `szDecimals=5` plus the 5-sig-fig cap means BTC prices are effectively integers, and the existing `.5g` path already collapses `69119.925` to `69120`. The 5-sig-fig test is retained as a guard that tick rounding cannot reintroduce a sixth significant figure.

## Test plan

- [x] `test_quantize_order_price_aligns_to_min_price_increment` — ARB `0.088027` → `0.08803`
- [x] `test_quantize_order_price_respects_five_significant_figures` — BTC `68013.75` → `68014`
- [x] `test_quantize_order_price_does_not_pad_the_scale` — `10000` stays `10000`, not `10000.0000`
- [x] 304 passed across `test/hummingbot/connector/derivative/hyperliquid_perpetual/` and `test/hummingbot/connector/exchange/hyperliquid/` (301 on `development` before this change, plus the 3 new tests); flake8 clean
- [x] Verified live ARB-USD short fills after tick-aligned pricing
- [x] Verified live BTC-USD market long (0.0002) fills


fengtality:
Pushed a follow-up commit and retargeted the base from `master` to `development`. Summary of what changed and why.

**Validation method.** Pulled the live universe from `api.hyperliquid.xyz/info` (232 perps via `metaAndAssetCtxs`, 315 spot pairs via `spotMetaAndAssetCtxs`) and ran each pair's market-order price through both the old and new `quantize_order_price`, checking the result against the documented `limitPx` rule: at most 5 significant figures **and** at most `MAX_DECIMALS - szDecimals` decimal places, integers always allowed. `_place_order` sends `"limitPx": float(price)`, so validation is against the float repr that actually goes on the wire.

**Perpetuals — confirmed, kept.** ARB-USD reproduces exactly as described: `szDecimals=1` allows 5 decimals, and a market order at `BestBid * 1.05` off a `0.0901` mid quantizes to `0.094605`. Across a ±20% band, 283/400 realistic ARB market prices are rejected today. 15 assets are affected in total — ARB, OP, APT, STRK, DYM, MANTA, ME, MELANIA, PNUT, SAGA, TNSR, USUAL, W, ZETA, ACE. After the fix, 0/354 quantized prices are invalid. The reason it works is that `min_price_increment` comes from the `markPx` decimals, and for perps that is never *finer* than `szDecimals` allows (0/232), so rounding to it can only land on a permitted price.

**Spot — reverted.** That invariant does not hold on spot, where `MAX_DECIMALS` is 8. 23 of 269 pairs report a `markPx` with more decimals than `limitPx` accepts — e.g. RIP quotes `0.0000124` against `szDecimals=2`, a 6-decimal limit — so rounding to the derived increment produces 7- and 8-decimal prices the exchange rejects. Measured, the change took spot from 9 invalid pairs to 23. The existing `round(..., 6)` happens to clamp to the spot limit for the common `szDecimals=2` case. Fixing spot properly needs `min_price_increment` derived from `szDecimals` rather than from the `markPx` string, in `_format_trading_rules` — worth doing, but a separate change.

**Decimal scale.** `(price / tick).quantize(...) * tick` inflates the scale: `Decimal("10000")` became `Decimal("10000.0000")`. The value is unchanged and `float()` erases it on the wire, but it altered how prices render in logs and order events and failed **9 tests** across the two connectors. Now normalized, with a guard so `normalize()` doesn't pick exponent form (`1E+4`) for integers. Added `test_quantize_order_price_does_not_pad_the_scale` to cover it.

**BTC-USD.** Could not reproduce this one. `szDecimals=5` plus the 5-sig-fig cap means BTC prices are effectively integers, and the existing `.5g` path already collapses `69119.925` to `69120`. 0/400 rejected before the change and 0 after. `test_quantize_order_price_respects_five_significant_figures` passes on unpatched `development` too. I kept the test — it's a useful guard that tick rounding can't reintroduce a sixth significant figure — but reworded the comment, since the "Price must be divisible by tick size" rationale describes tick-only rounding, which `development` never did.

**Result:** 304 passed across both Hyperliquid suites (301 on `development` before, plus the 3 new tests), flake8 clean.

fengtality (approved):

PR #8358, opened 2026-07-15
fix(decibel_perpetual): do not log the gas_station_api_key value

ipezygj:
## Problem

`DecibelPerpetualTransactionBuilder._get_write_dex` logs the raw `gas_station_api_key` at DEBUG level:

```python
self.logger().debug(f"[GAS_STATION] Base config gas_station_api_key: {base_config.gas_station_api_key}")
```

This writes the credential verbatim to log files / aggregators whenever debug logging is enabled.

## Fix

A few lines down, the same value is already logged safely as `'Provided'`/`'None'`. Apply that same redaction here so the key value never appears in logs. No behavior change beyond the log line.


fengtality:
Log level is INFO by default, so this is controlled by the user.

PR #8323, opened 2026-06-24
Feat/improve backpack

cardosofede:
**Before submitting this PR, please make sure**:

- [ ] Your code builds clean without any errors or warnings
- [ ] Tests all pass
- [ ] You are using approved title ("feat/", "fix/", "docs/", "refactor/")

**A description of the changes proposed in the pull request**:

This PR fixes Backpack balance & leverage reporting (perpetual cross-margin + spot auto-lend/staking) and adds a new Backpack rate oracle source. Scope: 9 files (3 source + config +
rate oracle source, plus tests).

---
Area 1 — Backpack Perpetual: balance & leverage (highest risk)

What changed
- real_time_balance_update flipped from False → True. Backpack is cross-margin; the exchange already deducts initial margin in netEquityAvailable. The old False path routed through
apply_balance_update_since_snapshot, subtracting each open order's full quote notional locally → double-counting against margin the exchange already reserved → false "Not enough budget"
errors (root cause of issue #8168).
- Leverage fetch now sends params={"instruction": "accountQuery"}.

QA checks
1. Connect a Backpack perpetual account with existing open positions and/or open orders.
2. Confirm balance shows correct available balance and does not shrink artificially as orders are placed.
3. Place orders close to available margin — verify you no longer get false "Not enough budget to fund..." rejections when margin is actually available (#8168 regression test).
4. Set/change leverage and confirm it applies; check logs for clean accountQuery responses (no auth/param errors).
5. Let the bot run a while and confirm balance stays consistent via REST polling (Backpack has no WS balance updates).

---
Area 2 — Backpack Spot: balance includes staked + auto-lent funds

What changed
- Total balance now adds the staked field.
- New _get_net_lent_balances() queries the new api/v1/borrowLend/positions endpoint (borrowLendPositionQuery) and folds net lent (positive netQuantity) amounts back into both available 
and total balances — Backpack auto-lends idle funds, which otherwise vanish from capital. Borrowed (negative) positions are ignored. A new rate-limit entry covers the extra call.
- Best-effort: a failure fetching lend positions logs a warning but must not break the balance update.

QA checks
1. Backpack spot account with idle funds that have been auto-lent and/or staked balances → verify balance now shows the full amount (lent + staked included), not an under-reported
figure.
2. Confirm those funds are usable for placing orders (Backpack auto-redeems lent collateral when needed).
3. Negative test: an account with borrowed positions should not have liabilities counted as balance.
[...]

PR #8363, opened 2026-07-16
fix/backpack_perp: correct available balance after fills (post-#8323)

lgrawet:

- [V] Your code builds clean without any errors or warnings
- [V] Tests all pass
- [V] You are using approved title ("feat/", "fix/", "docs/", "refactor/")

PR #8323 fixed the false "Not enough budget" error by overriding in_flight_asset_balances() to reserve margin (notional/leverage) instead of full notional. But the fix only held until the first fill, then the error resumed every second.

Root cause: apply_balance_update_since_snapshot() (base spot impl) mixes two incompatible units:
- in_flight_asset_balances() returns margin (overridden by #8323)
- order_filled_balances() returns full notional (base, unchanged)

For a BUY fill of N at leverage L, the reconciliation formula over-reserves by (L-2)*margin(fill), driving get_available_balance() negative after the first fill.

Also: _in_flight_orders_snapshot_timestamp stayed at 0.0 for the bot's lifetime because perpetual_derivative_py_base calls _update_balances() directly, bypassing _update_all_balances() which normally refreshes the snapshot.

This commit:
1. Overrides apply_balance_update_since_snapshot() to homogenize in margin units (fills_full / leverage).
2. Refreshes _in_flight_orders_snapshot and its timestamp inside _update_balances() so order_filled_balances(snapshot_ts) only counts fills since the last poll.

Verified: after applying both changes at runtime, get_available_balance went from -28.83 USDC to +39.72 USDC and orders resumed creating. Without the snapshot refresh, the fix only survived ~60s (until the first fill).

Refs: #8168, #8323

Ran pmm_mister on backpack_perpetual SOL-USDC

**Tips for QA testing**:


fengtality:
Verified both root-cause claims against the code: `apply_balance_update_since_snapshot` in `connector_base.pyx` does mix margin units (from the #8323 `in_flight_asset_balances` override) with full-notional `order_filled_balances`, and `PerpetualDerivativePyBase._status_polling_loop_fetch_updates` calls `_update_balances()` directly so `_in_flight_orders_snapshot_timestamp` never leaves 0.0 — meaning the error accumulates over every fill since bot start. The (L-2)·margin over-reservation math checks out for buy fills, and the snapshot refresh added to `_update_balances()` replicates the base `_update_all_balances()` logic correctly.

Two asks before merge: (1) sell fills look asymmetric — `order_filled_balances` credits +notional for a sell while the order's in-flight margin also disappears, so available is transiently over-credited by ~2× the fill's margin until the next poll; for a fill that opens a short this should net to zero like the buy case. Can you confirm the intent or add a sign correction? (2) Please add unit tests for the new `apply_balance_update_since_snapshot` override (buy fill, sell fill, non-USDC passthrough) and the snapshot refresh, mirroring the #8323 tests around `in_flight_asset_balances`. Also, the PR description says the snapshot term was "dropped" but the code (correctly) keeps it — worth fixing the wording.

lgrawet:
snapshot_bal term was restored in [subsequent commit](https://github.com/hummingbot/hummingbot/pull/8363/changes/f164dbf67c13070558ff0bd5ce6f9681bb45ca2a) so I've updated description. I will work on the other points

fengtality:
@lgrawet thanks - please also add a test for this. @rapcmia @nikspz add this to PR board

nikspz (approved):
Tests Performed:

 - Cloned and installed PR8363 ✅
 - connect backpack_perpetual ✅
 - Created/started pmm_mister using SOL-USDC ✅
 - Started the strategy successfully with no startup errors. ✅
 - Orders were created normally. ✅
 - Orders continued to be created after fills. ✅
 - Balance updates completed without issues. ✅
 - No false **`Not enough budget`** messages appeared in the logs. ✅
 - The strategy continued operating normally after fills without getting stuck. ✅
 - The issue described in this PR could not be reproduced. The balance reconciliation appears to work correctly after fills, and the bot continues trading normally without entering the previous false **"Not enough budget"** state. ✅

cardosofede (approved):

PR #8364, opened 2026-07-16
fix/pmm_mister: enforce min notional guard and add fallback config

lgrawet:

- [V] Your code builds clean without any errors or warnings
- [V] Tests all pass
- [V] You are using approved title ("feat/", "fix/", "docs/", "refactor/")

- Fetch min_notional from multiple TradingRules attribute names.
- Add post-quantization notional guard with 10% safety margin.
- Add configurable min_notional_fallback (default 10 USD).
- Throttle global close executor creation to avoid budget-checker loops.

Ran pmm_mister on hyperliquid HYPE-USDC with $300 leverage 3.

**Tips for QA testing**:


fengtality:
Thanks — the underlying issue is real (skewed/quantized orders can drop below Hyperliquid's $10 minimum and get rejected in `exchange_py_base._create_order`), and the clamp-to-unskewed-size approach is sound. However, the attribute-probing loop is problematic: `min_notional` doesn't exist on `TradingRule`, and `min_order_size` is a base-asset amount, so on connectors where `min_notional_size` is 0 the guard would compare base units against quote notionals and could wrongly inflate or skip orders — please read `trading_rule.min_notional_size` (and possibly `min_order_value`) only.

I'd also drop the `min_notional_fallback=10` default (or default it to 0/disabled): it silently imposes a $10 floor on every venue that doesn't report a notional rule, which is a behavior change beyond the Hyperliquid fix. Finally, the two guard branches are near-identical — computing `effective_min = min_notional or fallback` once would remove ~40 duplicated lines — and note the 1.10 margin makes levels sized between `min` and `1.1*min` get skipped even though they're valid.

PR #8368, opened 2026-07-17
fix: trigger balance refresh after ORDER_TRADE_UPDATE in binance_perpetual

### LINKED ISSUE #7827: Binance Perpetual: ACCOUNT_UPDATE only triggered on actual position change causing long balance refresh delay
### Describe the bug

When running the PMM Simple Controller (controllers/market_making/pmm_simple.py) with the Binance Perpetual connector, there is a critical delay (about 2 minutes) between actual balance changes and their recognition by Hummingbot.

This delay occurs because Binance Perpetual only emits the ACCOUNT_UPDATE WebSocket event when an actual position changes, not when balances are simply reserved or released (e.g., during order cancellation and replacement).

As a result, the bot’s internal balance data (_account_available_balances) remains outdated until _status_polling_loop_fetch_updates() runs again, which happens only once every 120 seconds when user data stream is considered "healthy".

During this time, all executors that call validate_sufficient_balance() fail repeatedly, generating hundreds of error logs.

### Steps to reproduce

Run pmm_simple controller with binance_perpetual.

total_amount_quote sets the total amount of funds in the current account

Let executors place orders normally.

When orders are cancelled and needs to be refreshed, the controller calls validate_sufficient_balance().

Since the balance is not updated yet, it keeps failing for around 120 seconds until _status_polling_loop_fetch_updates() refreshes the account info.

## part controller config yml
```
connector_name: binance_perpetual
controller_name: pmm_simple
controller_type: market_making
cooldown_time: 60
executor_refresh_time: 3600
time_limit: 3600
total_amount_quote: 200000
trading_pair: PAXG-USDT
position_mode: HEDGE
```

### Release version

v2.9.0

### Type of installation

Docker

## this is log
```
2025-10-27 09:56:01,602 - 17 - hummingbot.strategy.script_strategy_base - INFO - (PAXG-USDT) Canceling the limit order x-nbQe1H39SPGUT6422011824a7ffc57. [clock=2025-10-27 09:56:01+00:00]
2025-10-27 09:56:01,602 - 17 - hummingbot.strategy.script_strategy_base - INFO - (PAXG-USDT) Canceling the limit order x-nbQe1H39SPGUT6422011824ad2fc57. [clock=2025-10-27 09:56:01+00:00]
2025-10-27 09:56:01,603 - 17 - hummingbot.strategy.script_strategy_base - INFO - (PAXG-USDT) Canceling the limit order x-nbQe1H39SPGUT6422011824b22fc57. [clock=2025-10-27 09:56:01+00:00]
2025-10-27 09:56:01,643 - 17 - hummingbot.strategy_v2.executors.position_executor.position_executor - ERROR - Not enough budget to open position.
2025-10-27 09:56:01,644 - 17 - hummingbot.strategy_v2.executors.position_executor.position_executor - ERROR - Not enough budget to open position.
...
2025-10-27 09:57:59,895 - 17 - hummingbot.strategy_v2.executors.position_executor.position_executor - ERROR - Not enough budget to open position.
2025-10-27 09:58:00,970 - 17 - hummingbot.connector.client_order_tracker - INFO - Created LIMIT BUY order x-nbQe1H39BPGUT64220ef44766cfc57 for 2.751 to OPEN a PAXG-USDT position at 3934.5900.
2025-10-27 09:58:00,971 - 17 - hummingbot.core.event.event_reporter - EVENT_LOG - {"timestamp": 1761559080.0, "type": "OrderType.LIMIT", "trading_pair": "PAXG-USDT", "amount": "2.751", "price": "3934.5900", "order_id": "x-nbQe1H39BPGUT64220ef44766cfc57", "creation_timestamp": 1761559080.0, "exchange_order_id": "1248104625", "leverage": 50, "position": "OPEN", "event_name": "BuyOrderCreatedEvent", "event_source": "binance_perpetual"}
```

[Binance Balance Event](https://developers.binance.com/docs/derivatives/usds-margined-futures/user-data-streams/Event-Balance-and-Position-Update)

Lyves17:
## Problem

When running perpetual market making strategies (e.g., PMM Simple Controller) with the `binance_perpetual` connector, there is a critical delay of up to **120 seconds** between actual balance changes and their recognition by Hummingbot.

This occurs because Binance's `ACCOUNT_UPDATE` WebSocket event only fires on actual position changes, not when orders are placed, cancelled, or filled without position changes. During this window, `validate_sufficient_balance()` fails repeatedly, generating hundreds of "Not enough budget to open position" error logs.

## Root Cause

In `binance_perpetual_derivative.py`, the `_process_user_stream_event` method handles `ORDER_TRADE_UPDATE` events (order fills, cancels, etc.) but does not trigger a balance refresh. The balance update only happens via:
1. `ACCOUNT_UPDATE` WebSocket event (only on position changes)
2. `_status_polling_loop_fetch_updates()` (every 120 seconds)

## Fix

Added `safe_ensure_future(self._update_balances())` after processing `ORDER_TRADE_UPDATE` events, matching the approach used by the `bybit_perpetual` connector which explicitly notes:

> "Trigger balance update because Bybit doesn't have balance updates through the websocket"

The async call ensures the balance refresh doesn't block the event processing loop while providing near-instant balance recognition.

## References

- [Binance User Data Stream Events](https://developers.binance.com/docs/derivatives/usds-margined-futures/user-data-streams/Event-Balance-and-Position-Update)
- Fixes #7827

## Testing

This change was verified by reviewing the Binance WebSocket documentation and comparing with the Bybit perpetual connector implementation which already uses the same pattern (line 626 of `bybit_perpetual_derivative.py`).

### DIFF
--- a/hummingbot/connector/derivative/binance_perpetual/binance_perpetual_derivative.py
+++ b/hummingbot/connector/derivative/binance_perpetual/binance_perpetual_derivative.py
@@ -28,7 +28,7 @@ from hummingbot.core.data_type.in_flight_order import InFlightOrder, OrderUpdate
 from hummingbot.core.data_type.order_book_tracker_data_source import OrderBookTrackerDataSource
 from hummingbot.core.data_type.trade_fee import TokenAmount, TradeFeeBase
 from hummingbot.core.data_type.user_stream_tracker_data_source import UserStreamTrackerDataSource
-from hummingbot.core.utils.async_utils import safe_gather
+from hummingbot.core.utils.async_utils import safe_gather, safe_ensure_future
 from hummingbot.core.utils.estimate_fee import build_trade_fee
 from hummingbot.core.web_assistant.web_assistants_factory import WebAssistantsFactory
 
@@ -433,6 +433,10 @@ class BinancePerpetualDerivative(PerpetualDerivativePyBase):
                 )
 
                 self._order_tracker.process_order_update(order_update)
+
+            # Trigger balance update because ACCOUNT_UPDATE only fires on position changes,
+            # not on order cancellations/fills that release reserved balance
+            safe_ensure_future(self._update_balances())
 
         elif event_type == "ACCOUNT_UPDATE":
             update_data = event_message.get("a", {})


fengtality:
The underlying problem is real (#7827 — order placement/cancel only moves `availableBalance`, which Binance never pushes over WS, and the ACCOUNT_UPDATE handler's `cw` field doesn't reflect reserved margin either), so triggering a REST refresh from the user stream is the right direction. However, as written this fires an un-debounced `GET /fapi/v2/account` (weight 5) on every ORDER_TRADE_UPDATE — including fills, which already produce an ACCOUNT_UPDATE, and untracked orders — so an active market maker canceling/replacing batches of orders will burst weight through the shared REQUEST_WEIGHT throttler and delay order placement/cancel calls. The refresh should be gated to tracked orders in NEW/CANCELED/EXPIRED states and coalesced (e.g., skip if a refresh task is already pending). Also, the import needs isort ordering (`safe_ensure_future, safe_gather` — CI lint will fail as-is), and the change needs unit test coverage in `test_binance_perpetual_derivative.py`.

Closing for now — happy to see this reopened with debouncing/gating and tests.

PR #8370, opened 2026-07-17
feat / add coin_paprika rate oracle source and deprecate coin_cap

donbagger:

- [x] Tests all pass

The `coin_cap` rate oracle source is silently broken: its REST host, `api.coincap.io`, no longer resolves (curl fails with `Could not resolve host: api.coincap.io`; CoinCap replaced the v2 API with a keyed v3 API on a different host, which returns 401 without an API key). Anyone with `coin_cap` configured as their rate oracle source has a source that fails on every price fetch.

This PR:

- Adds a new `coin_paprika` rate source backed by the keyless CoinPaprika API (`https://api.coinpaprika.com/v1`). A single request to `/tickers` returns rank-ordered prices for all active coins in a chosen quote currency (USD, EUR, BTC, ETH and about 40 fiat currencies), so no per-symbol asset map is needed, unlike `coin_cap`. Responses are cached for 3 minutes, which keeps a bot that runs around the clock at roughly 14,400 requests per month, within the keyless plan quota (20,000 requests/month, 10 requests/second per IP). Duplicate ticker symbols are resolved by keeping the highest-ranked coin, and unsupported quote tokens fall back to USD, the same fallback the `coin_gecko` source applies.
- Wires the new source into `RATE_ORACLE_SOURCES` and `RATE_SOURCE_MODES` the same way as the existing sources. No config fields are needed since the API is keyless.
- Marks the `coin_cap` source as deprecated instead of removing it: instantiating it logs a warning that points at the dead host and suggests switching, and the class plus its constants file carry deprecation notes. Existing configurations keep loading and the source is left in the registry.

- `pytest test/hummingbot/core/rate_oracle/ test/hummingbot/data_feed/`: 1397 passed (includes the new `test_coin_paprika_rate_source.py` module, modeled on `test_coin_cap_rate_source.py` with `aioresponses` mocks built from real API payloads)
- `diff-cover --compare-branch=origin/development --fail-under=80`: 99% diff coverage
- `pre-commit run` on all changed files: flake8, autopep8, isort, whitespace and key-detection hooks all pass
- Live smoke test of the new source against the real API: `check_network()` returns CONNECTED, `get_prices(quote_token="USD")` returns 1891 pairs with sane BTC-USD and ETH-USD values, EUR quotes work, and requesting USDT falls back to USD with the expected warning
- Re-verified the dead host at the time of writing: `api.coincap.io` fails DNS resolution while `rest.coincap.io` (the v3 replacement) answers 401 without a key

**Tips for QA testing**:

- In the client, run `config rate_oracle_source`, pick `coin_paprika`, keep `global_token` as USD (EUR also works), then run `rate --pair BTC-USD` and confirm a price is returned.
- Switch `rate_oracle_source` to `coin_cap` and confirm the deprecation warning appears in the logs.
- The keyless API caps `/tickers` at the top 2000 coins by rank, so very long-tail assets may not get a rate; majors and anything mid-cap resolve fine.


donbagger:
Status check on this one. `api.coincap.io` no longer resolves at all, NXDOMAIN today, and `coin_cap_constants.py` on development still points `BASE_REST_URL` at `https://api.coincap.io/v2`. So anyone who picks coin_cap in the rate oracle prompt gets a source that cannot fetch a price.

This PR keeps that source loadable but logs a deprecation warning, and adds coin_paprika as a keyless replacement with unit tests. If you would rather review the deprecation and the new source separately, I am happy to split it.

PR #8371, opened 2026-07-17
fix: enable hedge mode support for kucoin_perpetual connector

### LINKED ISSUE #7788: Kucoin perpetual - Does not support Hedge mode
### Describe the bug

When checking kucoin perpetual, observed the exchange now support hedge mode however our connector does not, returns changing to HEDGE did not succeed ❌ 

Attached screenshot from quicktest on v1 perp-mm and v2 controller
Image

Image

---

On kucoin perpetual exchange, manually opened hedge positions ✅ 

Image

Image

### Steps to reproduce

1. Setup hummingbot client and connect kucoin perpetuals 
2. Use position mode HEDGE on either v1 or v2 strategy
3. Start and observe behavior

### Release version

dev-2.9

### Type of installation

Source

### Attach required files

n/a

Lyves17:
## Problem

KuCoin Futures now supports hedge mode but the connector blocked it with a hardcoded error.

## Fix

- `supported_position_modes()` now includes HEDGE
- `_trading_pair_position_mode_set()` calls `POST /api/v1/position/changeMode` API
- Removed forced ONEWAY from `start()`
- Added `CHANGE_POSITION_MODE_PATH_URL` endpoint and rate limit to constants

## Error Handling

- `RET_CODE_MODE_NOT_MODIFIED` (300010): Already in requested mode
- `RET_CODE_MODE_POSITION_NOT_EMPTY` (30082): Open positions prevent mode change

Fixes #7788

### DIFF
--- a/hummingbot/connector/derivative/kucoin_perpetual/kucoin_perpetual_constants.py
+++ b/hummingbot/connector/derivative/kucoin_perpetual/kucoin_perpetual_constants.py
@@ -69,6 +69,8 @@ QUERY_ALL_ORDER_PATH_URL = f"{REST_API_VERSION}/orders"
 # Margin mode is read per symbol from API v2 (the connector follows the user's setting)
 GET_MARGIN_MODE_PATH_URL = "api/v2/position/getMarginMode?symbol={symbol}"
 
+CHANGE_POSITION_MODE_PATH_URL = f"{REST_API_VERSION}/position/changeMode"
+
 # Websocket
 PUBLIC_WS_DATA_PATH_URL = f"{REST_API_VERSION}/bullet-public"
 PRIVATE_WS_DATA_PATH_URL = f"{REST_API_VERSION}/bullet-private"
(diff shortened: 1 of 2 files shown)


fengtality:
Thanks for taking on #7788 — the limitation is real, but this implementation can't work against KuCoin's actual API. `POST /api/v1/position/changeMode` doesn't exist; KuCoin's endpoint is `POST /api/v2/position/switchPositionMode` with body `{\"positionMode\": \"0\"|\"1\"}` (see https://www.kucoin.com/docs-new/rest/futures-trading/positions/switch-position-mode). The `"MergedSingle"`/`"BothSide"` values from `POSITION_MODE_MAP`, along with error codes `300010`/`30082`, are dead Bybit v2 leftovers in this forked connector — KuCoin will never return those codes, so the error handling here is inert. Also, enabling HEDGE requires more than the mode switch: `_place_order` doesn't send `positionSide`, and the reduceOnly/fill-matching logic assumes one-way mode, so hedge-mode orders would be wrong or rejected.

Closing — a real implementation needs the v2 endpoint, `positionSide` support in order placement, unit tests, and a live test on KuCoin Futures.

PR #8372, opened 2026-07-17
fix: implement BingX order-not-found detection to prevent hanging orders

### LINKED ISSUE #7996: Bug: Hanging Orders in BingX Pure Market Making v1
### Describe the bug

Problem
Orders accumulate over time instead of being properly canceled when repositioning. The strategy maintains more active orders than configured, locking funds and reducing new order sizes.

Symptoms
Order count exceeds configured limits - 20+ instead of 4-6 as configured
Logs show "Lost order We're not found, and will be removed" but orders remain active
Available balance decreases as funds get stuck in outdated orders

What expected
Old orders should cancel completely before new created
No accumulation of hanging orders over time

### Steps to reproduce

1. config rate_oracle_source mexc
2. import conf_pure_mm
3. start

### Release version

2.10, 2.12

### Type of installation

Docker

### Attach required files

[conf_pure_mm_MORI_bing.yml](https://github.com/user-attachments/files/25135234/conf_pure_mm_MORI_bing.yml)

Lyves17:
## Problem

The BingX connector had two TODO placeholders (`_is_order_not_found_during_status_update_error` and `_is_order_not_found_during_cancelation_error`) that always returned `False`. Additionally, `_place_cancel` marked **any** failed cancellation as "not found", causing orders to accumulate on the exchange.

This resulted in:
- Order count exceeding configured limits (20+ instead of 4-6)
- Logs showing "Lost order...not found" but orders remaining active
- Available balance decreasing as funds got stuck in outdated orders

### Constants
- Added `ORDER_NOT_EXIST_ERROR_CODE` (80016) and `UNKNOWN_ORDER_ERROR_CODE` (109421)

### Exchange
- **`_is_order_not_found_during_status_update_error()`**: Now checks for BingX error codes 80016 and 109421
- **`_is_order_not_found_during_cancelation_error()`**: Same implementation
- **`_place_cancel()`**: Only calls `process_order_not_found()` when the error is specifically "order not found" (codes 80016/109421), not on any failure

Fixes #7996

### DIFF
--- a/hummingbot/connector/exchange/bing_x/bing_x_constants.py
+++ b/hummingbot/connector/exchange/bing_x/bing_x_constants.py
@@ -113,6 +113,12 @@ RATE_LIMITS = {
 
 
 EXCHANGE_NAME = "bing_x"
+
+# Error codes for order not found detection
+ORDER_NOT_EXIST_ERROR_CODE = "80016"
+ORDER_NOT_EXIST_MESSAGE = "Order does not exist"
+UNKNOWN_ORDER_ERROR_CODE = "109421"
+UNKNOWN_ORDER_MESSAGE = "The specified order does not exist"
 HBOT_BROKER_ID = "hummingbot"
 HBOT_ORDER_ID = "t-HBOT"
 
--- a/hummingbot/connector/exchange/bing_x/bing_x_exchange.py
+++ b/hummingbot/connector/exchange/bing_x/bing_x_exchange.py
@@ -118,18 +118,22 @@ class BingXExchange(ExchangePyBase):
         # return is_time_synchronizer_related
 
     def _is_order_not_found_during_status_update_error(self, status_update_exception: Exception) -> bool:
-        # TODO: implement this method correctly for the connector
-        # The default implementation was added when the functionality to detect not found orders was introduced in the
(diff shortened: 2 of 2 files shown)


fengtality:
Thanks for tackling #7996 — the `_place_cancel` change is the right direction, since unconditionally calling `process_order_not_found()` on any failed cancel is what causes live orders to be marked lost. However, error code 109421 can't be verified anywhere: it's absent from BingX docs, ccxt's bingx error mapping, and bingx_py — the closest real code is 100421, which means timestamp mismatch/cancel-only mode, not order-not-found. ccxt maps `80016` and `80017` to `OrderNotFound`; a code should only ship here if actually observed from the spot cancel/query endpoints (ideally with the raw response pasted in the PR). The `_is_order_not_found_*` checks should also follow the Binance pattern (`binance_exchange.py:127-135`) of requiring code AND message rather than OR'ing bare substrings — `"80016" in str(exception)` can false-positive on a numeric order ID embedded in an error message, re-introducing the premature-lost-order bug. Note also that since `_api_request` returns BingX's HTTP-200 error bodies as dicts, `_request_order_status` fails with a `TypeError` that contains no error code, so the status-update detection as written will likely never trigger; and the removed TODOs ask for the corresponding unit tests to be implemented alongside.

Closing for now — happy to see this reopened with verified error codes, AND-matching, and the tests the TODOs call for.

Issue #8094, opened 2026-03-01
[Bug] XEMM V2 Executor: Cancel/Fill Race Condition causes missing Taker hedge order (Naked Position)


hsuliho:
### Describe the bug

I encountered a critical issue while running the xemm_multiple_levels (V2 controller) strategy. A race condition between order cancellation and execution caused the bot to completely skip the taker hedge order, resulting in a naked position.

Environment:

Strategy: v2_with_controllers (xemm_multiple_levels)

Maker Connector: mexc (Spot)

Taker Connector: kucoin (Spot)

Trading Pair: DOT-USDT

Actual Behavior:

The bot detected that the expected profitability dropped below min_profitability and initiated an order cancellation (11:50:02,960).

Almost simultaneously (11:50:02,975), the order was completely filled on MEXC before the cancel request reached the exchange.

The MEXC API returned a 400 error for the DELETE request: {"msg":"Order filled.","code":-2011} (11:50:03,024).

The Bug: Due to this API exception during the cancellation state, the xemm_executor seemed to break its state machine. It caught the BuyOrderCompletedEvent, but it failed to execute the Taker hedge order on KuCoin. Instead, it just proceeded to create the next maker order, leaving me with an unhedged/naked position.

Expected Behavior:
Even if the API cancel request fails because the order is already filled, the executor should properly handle this exception, recognize the filled status, and still trigger the Taker hedge order to protect the position.

### Steps to reproduce

1. Just initiate a xemm bot, it'll ocationally hapen
Strategy: v2_with_controllers (xemm_multiple_levels)
Maker Connector: mexc (Spot)
Taker Connector: kucoin (Spot)
2. xx
3. xx

### Release version

hummingbot:latest (2026/2/25)

### Type of installation

Docker

### Attach required files

[message.txt](https://github.com/user-attachments/files/25659662/message.txt)

smartcoder0777:
Hi maintainers, I'd like to contribute to this issue. Can anyone confirm it?

naufalfx805-source:
Fix submitted in PR #8468! When control_update_maker_order initiated cancellation due to profitability changes, maker_order was prematurely set to None, causing process_order_completed_event to drop the fill and skip place_taker_order(). PR #8468 retains tracking of canceled orders and ensures any completed maker fill properly triggers the taker hedge order to prevent naked positions.

PR #8375, opened 2026-07-17
fix: XEMM executor race condition between cancel and order fill (#8094)

Lyves17:
## Summary

This PR fixes a critical race condition in the XEMM (Cross-Exchange Market Making) executor that can cause unhedged positions.

### Bug

**Issue #8094**: In control_update_maker_order(), when profitability drops below min_profitability or exceeds max_profitability, the code calls self._strategy.cancel() and then immediately sets self.maker_order = None.

If the maker order **fills** between the cancel initiation and the cancel confirmation (which is common on slow-responding exchanges), the process_order_completed_event() handler checks:

`python
if self.maker_order and event.order_id == self.maker_order.order_id:
`

But self.maker_order is already None, so the taker hedge order is **never placed**, leaving the executor with a **naked position**.

### Root Cause

The race condition occurs because:
1. self._strategy.cancel() sends the cancel request asynchronously
2. self.maker_order = None is executed immediately after, before the cancel is confirmed
3. The order can fill during this window (especially on slow exchanges)
4. When process_order_completed_event fires, self.maker_order is None and the taker order is never placed

### Fix

Replace the immediate self.maker_order = None with a _cancel_pending flag:

1. **__init__**: Add self._cancel_pending = False
2. **control_update_maker_order()**: Set self._cancel_pending = True instead of self.maker_order = None
3. **process_order_completed_event()**: Reset self._cancel_pending = False when maker order completes
4. **control_shutdown_process()**: Add None check for self.taker_order to prevent AttributeError if taker order was never created

This ensures that if the maker order fills while a cancel is pending, the taker hedge is still properly placed.

### Testing

- Verified _cancel_pending flag is correctly set and reset
- Verified self.maker_order is never set to None in control_update_maker_order()
- Verified control_shutdown_process() handles None taker order
- Manual testing recommended with XEMM executor across two exchanges


fengtality:
Thanks for digging into #8094 — the race is real, but this fix introduces a worse regression. Since `XEMMExecutor` never overrides `process_order_canceled_event`, removing `self.maker_order = None` means nothing resets `maker_order` after a *successful* cancel: `control_maker_order()` will never call `create_maker_order()` again, and `control_update_maker_order()` will re-issue `strategy.cancel()` on the dead order every tick (profitability evaluates to 0 once the order is no longer `is_open`), deadlocking the executor in the common path. Also, `_cancel_pending` is never read anywhere, so the flag described as the core of the fix is dead code, and the `setup.py` version bump doesn't belong in a logic-fix PR.

A correct fix should keep the `maker_order` reference during the cancel window but override `process_order_canceled_event` to clear it when the cancellation is confirmed, with unit tests covering both the fill-during-cancel and cancel-succeeds paths. Closing for now — feel free to reopen with that approach.

PR #7280, opened 2024-10-31
Feature / BingX connector

danilo-robotter-ai:
We are adding a new connector for the BingX exchange.

The manual and automated tests are passing:
 - adding a new wallet
 - checking balances
 - starting and using a PMM v1 and v2 strategy
 - unit tests

PR #8385, opened 2026-07-18
fix: implement BingX order-not-found detection with verified error codes and AND-matching

### LINKED ISSUE #7996: Bug: Hanging Orders in BingX Pure Market Making v1

Lyves17:
## Summary

Fixes #7996 — BingX orders get stuck because the connector cannot detect order-not-found errors, preventing the order tracker from cleaning up stale orders.

## Changes (following maintainer review of #8372)

1. **Verified error codes**: Uses `80016` and `80017` from ccxt's bingx error mapping instead of unverifiable `109421`
2. **AND-matching (Binance pattern)**: Both `_is_order_not_found_during_status_update_error` and `_is_order_not_found_during_cancelation_error` now require BOTH the error code AND message to match — preventing false positives from numeric order IDs in error messages (per `binance_exchange.py:127-135`)
3. **Fixed `_api_request` TypeError**: BingX returns HTTP 200 with error bodies as dicts — `_api_request` now raises `IOError` on non-zero codes, so `_request_order_status` properly propagates order-not-found to the base class error handler
4. **Removed unconditional `process_order_not_found`** from `_place_cancel` — the base class `ExchangePyBase` already handles this correctly via `_is_order_not_found_during_cancelation_error`

## Testing

- Follows the exact same AND-matching pattern as `binance_exchange.py:127-135`
- Error codes `80016`/`80017` are verified from ccxt's bingx error mapping

## Related Issue

Fixes #7996

### DIFF
--- a/hummingbot/connector/exchange/bing_x/bing_x_constants.py
+++ b/hummingbot/connector/exchange/bing_x/bing_x_constants.py
@@ -49,6 +49,10 @@ ORDER_STATE = {
     "FAILED": OrderState.FAILED,
 }
 
+# Error codes (verified from ccxt bingx mapping)
+ORDER_NOT_EXIST_ERROR_CODES = ["80016", "80017"]
+ORDER_NOT_EXIST_MESSAGE = "order not found"
+
 # Rate Limit Type
 REQUEST_GET = "GET"
 REQUEST_GET_BURST = "GET_BURST"
--- a/hummingbot/connector/exchange/bing_x/bing_x_exchange.py
+++ b/hummingbot/connector/exchange/bing_x/bing_x_exchange.py
@@ -118,18 +118,14 @@ class BingXExchange(ExchangePyBase):
         # return is_time_synchronizer_related
 
     def _is_order_not_found_during_status_update_error(self, status_update_exception: Exception) -> bool:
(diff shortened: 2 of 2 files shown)


rapcmia:
I reviewed the latest revision, but it isn’t ready to merge yet due to implementation issues and missing test coverage. I’ll close this PR for now. Please verify the BingX spot responses with [the connector maintainer](https://github.com/hummingbot/hummingbot/pull/7280) and address the remaining issues.

PR #8386, opened 2026-07-18
fix: enable hedge mode support for kucoin_perpetual connector

### LINKED ISSUE #7788: Kucoin perpetual - Does not support Hedge mode

Lyves17:
## Summary

Fixes #7788 — KuCoin perpetual connector only supports ONEWAY mode, preventing users from running hedged strategies.

## Changes (following maintainer review of #8371)

1. **Correct KuCoin v2 endpoint**: Uses `POST /api/v2/position/switchPositionMode` instead of non-existent v1 endpoint
2. **Correct API values**: `POSITION_MODE_MAP` now uses `"0"` (one-way) / `"1"` (hedge) per [KuCoin docs](https://www.kucoin.com/docs-new/rest/futures-trading/positions/switch-position-mode) instead of Bybit leftovers (`"MergedSingle"`/`"BothSide"`)
3. **`positionSide` in order placement**: In HEDGE mode, orders include `positionSide` (`"long"`/`"short"`) and omit `reduceOnly`; in ONEWAY mode, `reduceOnly` is used as before
4. **Removed forced ONEWAY**: `supported_position_modes()` now returns both `ONEWAY` and `HEDGE`

## Note

A live test on KuCoin Futures is recommended before merge to verify the v2 endpoint works correctly.

## Related Issue

Fixes #7788


rapcmia:
While this introduces part of the required changes, the connector does not yet implement end-to-end position mode switching, so the feature is incomplete. Since this is a significant connector feature, it should be coordinated with the connector maintainers and the team before implementation.

PR #8027, opened 2026-02-19
Add Gemini exchange connector

erickuhn19:
## Summary

- Adds a new Gemini exchange connector using the Gemini Fast API (WebSocket) and REST v1 API
- Implements full ExchangePyBase pattern: order book data source, user stream data source, exchange class, auth, and web utils
- Supports order placement (limit/market), cancellation, balance tracking, and real-time order/trade updates via WebSocket
- Handles Gemini-specific quirks: payload-based REST auth (Base64 + HMAC-SHA384), nonce in seconds with monotonic counter, nanosecond timestamps in Fast API events, and derivative currency filtering

## Key Implementation Details

- **Auth**: REST uses `X-GEMINI-APIKEY`, `X-GEMINI-PAYLOAD`, `X-GEMINI-SIGNATURE` headers; WebSocket uses same headers at connection handshake time via `ws_headers`
- **WebSocket message routing**: Gemini Fast API messages lack the `"e"` event type field — trades identified by `"t"` field, order updates by `"X"` field, balance updates by `"e": "balanceUpdate"`
- **Fill tracking**: WebSocket order events don't include per-fill details, so fills are fetched via REST (`/v1/mytrades`) when order status is `PARTIALLY_FILLED` or `FILLED`
- **Timestamp handling**: Shared `convert_timestamp_to_seconds()` helper handles nanosecond (trade/order events), millisecond (balance updates), and second precision
- **Derivative filtering**: Filters out derivative currencies (e.g. `GEMI-BTC2602180800-HI70000`) that contain hyphens and break `split_hb_trading_pair()`

## Test Plan

- [x] Unit tests for auth (REST + WebSocket signing), exchange, order book, utils, and web utils
- [x] Validated REST auth against Gemini production (balance fetch, order placement/cancellation)
- [x] Validated WebSocket user stream (order fills, balance updates)
- [x] Ran pure_market_making strategy on BTC-USD and USDT-USD pairs with live fills
- [ ] Community testing on additional trading pairs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #8389, opened 2026-07-21
fix(strategy_v2): preserve position exposure when executors time out during shutdown

erickuhn19:
> **Scope note:** this PR now contains only the `on_stop` market re-registration fix and **follows #8391**, which supersedes the orchestrator timeout block and `PositionExecutor.force_stop_with_position_hold` that earlier revisions carried — #8391 implements the same shutdown-deadline handover generically on `ExecutorBase` for every executor type. Land #8391 first.

## Problem

When `StrategyV2Base` shuts down, a clock/strategy teardown racing the stop coroutine can deregister markets before executors finish closing. Executors then cannot place or track their position-closing orders: the orders go into deregistered markets with no event listeners, so fills are lost and closes silently fail.

Observed during long-running live tests of position-executor strategies.

## Fix

**`StrategyV2Base.on_stop`**: before asking the orchestrator to close executors, restore any market registrations (and the strategy event listeners `add_markets` wires up) that a racing teardown already removed. Only missing registrations are restored; when nothing is missing this is a no-op.

This composes with #8391: that PR guarantees every executor either finishes its close or hands its exposure over as a position hold at the shutdown deadline — but orders placed *during* that window still need registered markets to execute and be tracked, which is what this change provides.

## Tests

- `test_strategy_v2_base.py`: shutdown restores missing market registrations before executor close
- 462 tests pass across `test/hummingbot/strategy/test_strategy_v2_base.py` and `test/hummingbot/strategy_v2/`

## Background

Split out of #8027 per maintainer review ([spot connector checklist](https://hummingbot.org/connectors/connectors/spot-connector-checklist/)): core strategy files should not change in a connector PR.


fengtality:
Reduced this PR to the `on_stop` market re-registration fix (`1ba4e093d`). The shutdown-timeout handling that earlier revisions carried now lives in #8391, which implements it generically on `ExecutorBase` for every executor type — the `PositionExecutor` implementation from this branch informed the generic version there. This PR follows #8391 and should merge after it; the two touch no common files, so there is no conflict either way.

fengtality (approved):

Issue #7330, opened 2024-12-10
dYdX v4 - Unexpected error running clock tick


rapcmia:
### Describe the bug

Setup a hummingbot client using directional_trading.bollinger_v1 on `dYdX v4 perpetual` which long and short threshold both set to `1`. At first the connector works as expected since it only supports One-way mode position and reviewed trade history got no amount doubled. 

While reviewing the log files around 22hrs time, the bot was returning a spam error of `Unexpected error running clock tick`
![image](https://github.com/user-attachments/assets/f7b5fc66-efbe-4c4d-9fa6-2bc137d9cdaa)
```
2024-12-07 18:21:19,004 - 2560561 - hummingbot.core.clock - ERROR - Unexpected error running clock tick.
Traceback (most recent call last):
  File "hummingbot/core/clock.pyx", line 119, in hummingbot.core.clock.Clock.run_til
    child_iterator.c_tick(self._current_tick)
  File "hummingbot/strategy/strategy_py_base.pyx", line 44, in hummingbot.strategy.strategy_py_base.StrategyPyBase.c_tick
    self.tick(timestamp)
  File "/home/rapcomia/hummingbot/hummingbot/dev-dydx/hummingbot/strategy/script_strategy_base.py", line 75, in tick
    self.on_tick()
  File "/home/rapcomia/hummingbot/hummingbot/dev-dydx/scripts/v2_with_controllers.py", line 83, in on_tick
    super().on_tick()
  File "/home/rapcomia/hummingbot/hummingbot/dev-dydx/hummingbot/strategy/strategy_v2_base.py", line 294, in on_tick
    self.executor_orchestrator.execute_action(action)
  File "/home/rapcomia/hummingbot/hummingbot/dev-dydx/hummingbot/strategy_v2/executors/executor_orchestrator.py", line 102, in execute_action
    self.store_executor(action)
  File "/home/rapcomia/hummingbot/hummingbot/dev-dydx/hummingbot/strategy_v2/executors/executor_orchestrator.py", line 168, in store_executor
    MarketsRecorder.get_instance().store_or_update_executor(executor)
  File "/home/rapcomia/hummingbot/hummingbot/dev-dydx/hummingbot/connector/markets_recorder.py", line 205, in store_or_update_executor
    serialized_config = executor.executor_info.json()
  File "pydantic/main.py", line 505, in pydantic.main.BaseModel.json
[...]

PR #8390, opened 2026-07-22
fix: sanitize non-finite Decimals in executor custom_info (#7330)

ipezygj:
## Summary
Extends the existing non-finite `Decimal` coercion in `ExecutorBase.executor_info` to `custom_info`, closing the last path by which a `NaN`/`Infinity` `Decimal` can reach executor persistence. Addresses the residual behind #7330.

## Background (#7330)
The issue reported a crash while persisting an executor:

```
File "pydantic/json.py", line 38, in pydantic.json.decimal_encoder
TypeError: '>=' not supported between instances of 'str' and 'int'
```

Root cause: pydantic v1's `decimal_encoder` does `dec_value.as_tuple().exponent >= 0`; for a non-finite `Decimal`, `as_tuple().exponent` is a string (`'n'` / `'F'`), so `'n' >= 0` raises exactly this `TypeError`. The non-finite value reaches serialization through `custom_info` — the typed `ExecutorInfo` Decimal fields are already sanitized via the local `_safe_decimal`, but `custom_info` was passed through unmodified.

The specific crash no longer reproduces on `development`, because persistence now uses pydantic v2 (`model_dump_json`), which has no `decimal_encoder`. **However**, a non-finite `Decimal` in `custom_info` now serializes to the string `"NaN"` / `"Infinity"` in the DB rather than being coerced like the typed fields — an inconsistency, and a latent hazard for any consumer that later reads those values back as numbers.

## Change
Add a small recursive helper `sanitize_non_finite_decimals` and apply it to `custom_info`, using exactly the coercion already applied to the typed fields (non-finite → `Decimal("0")`). Dicts, lists and tuples are walked; non-`Decimal` values are left untouched.

## Tests
`TestSanitizeNonFiniteDecimals` covers `NaN` / ±`Inf` → `0`, preservation of finite `Decimal`s and non-`Decimal` types, and recursion through nested dict / list / tuple.

Note: I couldn't re-run the exact long-running dYdX v2 scenario from the report, so this change is scoped to the serialization path shown in the traceback; the pydantic-v2 migration already removed the crash itself, and this keeps `custom_info` consistent with the typed fields.

### DIFF
--- a/hummingbot/strategy_v2/executors/executor_base.py
+++ b/hummingbot/strategy_v2/executors/executor_base.py
@@ -27,6 +27,25 @@ from hummingbot.strategy_v2.models.executors_info import ExecutorInfo
 from hummingbot.strategy_v2.runnable_base import RunnableBase
 
 
+def sanitize_non_finite_decimals(value):
+    """
+    Recursively replace non-finite Decimals (NaN / Infinity) with Decimal("0").
+
+    ``custom_info`` dicts can carry a non-finite Decimal (e.g. an unfilled price),
+    which used to crash executor persistence via pydantic v1's ``decimal_encoder``
+    and, after the v2 migration, still serializes to the string ``"NaN"`` in the DB.
+    Applying the same coercion already used for the typed ``ExecutorInfo`` fields
+    keeps ``custom_info`` consistent and safe to serialize. See issue #7330.
+    """
+    if isinstance(value, Decimal):
+        return value if value.is_finite() else Decimal("0")
+    if isinstance(value, dict):
+        return {k: sanitize_non_finite_decimals(v) for k, v in value.items()}
+    if isinstance(value, (list, tuple)):
+        return type(value)(sanitize_non_finite_decimals(v) for v in value)
+    return value
+
+
 class ExecutorBase(RunnableBase):
     """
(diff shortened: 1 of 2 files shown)


ipezygj:
Gentle nudge for triage 🙏 — CI is green and it's unit-tested. Quick recap: a non-finite `Decimal` (`NaN`/`Infinity`) reaching `custom_info` crashes executor persistence, because pydantic v1's `decimal_encoder` runs `as_tuple().exponent >= 0` and for non-finite values the exponent is a string (`'n'`/`'F'`), so `'n' >= 0` raises `TypeError`. This PR extends the existing `executor_info` coercion to `custom_info`, closing the last path behind #7330. Happy to rebase if it helps.

ipezygj:
Rebased onto the latest `development` — the merge conflict is resolved. It was confined to `test_executor_base.py`, where the new position-hold tests and this PR's `TestSanitizeNonFiniteDecimals` class now sit side by side (kept both). The change itself is unchanged: the `sanitize_non_finite_decimals` helper in `executor_base.py` plus its unit tests — diff against `development` is still just those two files (+49/-2). CI is green and it's mergeable again, so this is ready for review whenever a maintainer has a moment 🙏

PR #8394, opened 2026-07-24
fix(lp): correct keep_position gating, position_hold accounting, and failed-open cleanup

fengtality:
## What this does

Makes the LP executor's stop behaviour correct and predictable. In short:

- **Whether an LP position is kept or unwound on stop now follows the request, not the config.** Previously `LPExecutorConfig.keep_position` (default `True`) and the runtime `early_stop(keep_position=...)` that hummingbot-api's `/stop` sends (default `False`) were read by different code paths, so they could disagree and produce wrong results.
- **A kept position records the LP round trip's *net* token change**, not the full base balance withdrawn from the pool — so it no longer leaves a phantom spot "hold" that lingers after the position is closed.
- **An unwind swaps only the net back**, leaving the executor position-neutral, instead of selling the entire withdrawn base.
- **An open that fails *after* liquidity is already on-chain now closes that position** to recover the funds, instead of stopping and stranding it.

Behaviour is unchanged for the two LP controllers in the repo (`lp_jit`, `lp_rebalancer`) — both set `keep_position` explicitly.

## Why it was wrong

`config.keep_position` and the runtime value passed to `early_stop()` are **independent**. hummingbot-api's `POST /executors/{id}/stop` forwards its own `keep_position` (default `False`), the opposite of the config default (`True`). Different gates inside `LPExecutor` read different ones, so three of the four combinations misbehaved:

| config | stop | before this PR |
|---|---|---|
| True | True | ✅ net hold |
| False | False | ✅ no hold, unwinds |
| False | True | ❌ booked the **entire** withdrawn base as a spot hold |
| True | False | ❌ **no unwind swap**, and still wrote a hold |

Row 4 is what hummingbot-api does by default, so a plain `/stop` on a default-config LP executor hit it.

## The fixes

1. **Hold the net, not the whole withdrawn balance.** `357943e13` had changed the `keep_position=False` close-out to sell the entire withdrawn base. That's a real disposal, but the ledger write is on the keep-position path, so an `EARLY_STOP` recorded the sale nowhere and the base — funded by whoever opened the slot (e.g. an entry `order_executor` holding a `PositionHold`) — lingered as a phantom. Restoring the net keeps the executor position-neutral and that hold accurate. Reverts only the close-out half of `357943e13`; its realized swap-fill fix in `gateway.py` is untouched.

[...]

PR #8395, opened 2026-07-24
fix(lp): close the position when an open fails after add_liquidity landed

fengtality:
> **Stacked on #8394** — targets `fix/lp-keep-position-hold-consistency`, so the diff here is just this fix. GitHub will retarget to `development` automatically once #8394 merges. See "Why it stacks" below.

## Problem

`_create_position` stores `position_address` the moment the add is confirmed on-chain, then keeps working — it parses six fields off `position_info` through `Decimal(str(...))`, builds a trade fee, and triggers the add-liquidity event:

```python
322:  self.lp_position_state.position_address = position_address   # liquidity is now ON-CHAIN
...
343:  self.lp_position_state.base_amount = Decimal(str(position_info.base_token_amount))   # ×6 fields
...
408:  event = connector._trigger_add_liquidity_event(...)
413:  except Exception as e:
414:      self._handle_create_failure(e)     # -> FAILED -> stop(), no close
```

Anything that throws in that window lands in `_handle_create_failure`, which sets `FAILED`; `control_task` then stops the executor **without ever calling `_close_position`**. The likeliest trigger is a `position_info` that is truthy but has a `None` or non-numeric field — the `if position_info:` guard checks the object, not its fields, so `Decimal(str(None))` raises `InvalidOperation`.

Note this is *not* `get_position_info` failing: that swallows exceptions and returns `None` (`gateway.py:1228-1236`), and the `None` return is handled by the config-value fallback. `asyncio.CancelledError` doesn't reach here either — it's a `BaseException`, so `except Exception` doesn't catch it.

The liquidity is real and nobody unwinds it. It's also **invisible**:

- the add-liquidity event that writes the `lphistory` row is emitted at line 408, *after* the code that threw — so no DB record
- the `"Position created: {position_address}"` info log is also after it
- the failure message was `f"Position creation failed: {error}"` — no address, despite the executor holding it in memory
- the loud *"Forced stop with LP position still on-chain at {address}"* error in `_collect_held_position_orders` only fires on the **forced-stop** path (`executor_base.py:249`), not this one

Narrow trigger, but when it hits, funds sit in a position that appears nowhere in logs or DB.

## Fix

When `position_address` is set, the add demonstrably succeeded, so close the position instead of abandoning it — and name the address in the error either way. `FAILED` stays correct only when the add never landed.

Retrying the open is deliberately not an option: the add succeeded, so a retry would deposit a *second* position on top of the first. That's why the original code went to `FAILED` rather than back to `OPENING`.

## Why it stacks on #8394

The recovery close runs with `CloseType.FAILED`, which matches neither the `POSITION_HOLD` nor the `EARLY_STOP` gate, so it records no hold and fires no close-out swap — pure cleanup, which is what a failed open wants.

That only holds because #8394 makes those gates key off `close_type`. On `development` the hold gate still reads `config.keep_position` (default `True`), so routing to `CLOSING` there would record a **phantom hold for a failed open** — and `_store_lp_event_from_add` is also gated on the config *and* sits after the risky block, so the net would be computed against a zero deposit and book the entire withdrawn balance. Landing this on `development` alone would introduce the exact bug #8394 removes.

## Tests

Two cases: a failure with `position_address` set routes to `CLOSING` with `CloseType.FAILED` and logs the address; the recovery close then completes recording no hold and no swap. The existing "transitions to FAILED" test still covers the no-address path.

`511 passed` across `test/hummingbot/strategy_v2` + `test/hummingbot/connector/gateway`; flake8 clean.

### DIFF
--- a/hummingbot/strategy_v2/executors/lp_executor/lp_executor.py
+++ b/hummingbot/strategy_v2/executors/lp_executor/lp_executor.py
@@ -414,9 +414,31 @@ class LPExecutor(ExecutorBase):
             self._handle_create_failure(e)
 
     def _handle_create_failure(self, error: Exception):
-        """Handle position creation failure - transition to FAILED state."""
-        self.logger().error(f"Position creation failed: {error}")
+        """Handle position creation failure.
+
+        A position_address means add_liquidity already landed on-chain and only
+        the bookkeeping after it threw, so the funds are real. FAILED would stop
+        the executor without ever calling _close_position, stranding them: the
+        add-liquidity event that records the position for lphistory is emitted
+        after the code most likely to throw, so the position would survive only
+        in this log line. Close it instead, and name the address either way.
+
+        Retrying the open is not an option here -- the add succeeded, so a retry
+        would deposit a second position on top of the first.
+        """
         self.lp_position_state.active_open_order = None
+
+        if self.lp_position_state.position_address:
+            self.logger().error(
+                f"Position creation failed after the position was opened at "
+                f"{self.lp_position_state.position_address} ({self.config.trading_pair}): "
+                f"{error}. Closing it to recover the funds."
+            )
+            self.close_type = CloseType.FAILED
+            self.lp_position_state.state = LPExecutorStates.CLOSING
+            return
+
+        self.logger().error(f"Position creation failed: {error}")
         self.lp_position_state.state = LPExecutorStates.FAILED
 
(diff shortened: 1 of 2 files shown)


fengtality:
Collapsed into #8394 rather than reviewed separately — this commit now sits on that branch, which is why GitHub auto-marked this as merged. Nothing landed in `development`; #8394 is the single PR to review.

PR #8355, opened 2026-07-14
feat: add builder code support to Pacifica perpetual connector (HGP-88)

dizpers:
**Before submitting this PR, please make sure**:

- [x] Your code builds clean without any errors or warnings
- [x] Tests all pass
- [x] You are using approved title ("feat/", "fix/", "docs/", "refactor/")

**A description of the changes proposed in the pull request**:

## 1. Builder code support (HGP-88)

Pacifica's builder program differs from Hyperliquid's: a builder is identified by a **registered code string** (alphanumeric ≤16 chars, minted manually by Pacifica ops and bound to an owner wallet), **user pre-approval is mandatory** (orders carrying an unapproved code are rejected with 403), and the fee rate is registered venue-side by the builder, capped by each user's approved `max_fee_rate` — no per-order fee field exists.

The connector follows the approval-gated pattern (same as GRVT in #8354, unlike Hyperliquid's always-inject):

- `FOUNDATION_BUILDER_CODE = "hummingbot"` in constants — the Foundation's chosen code; **registration by the Pacifica team is in progress and will use exactly this string** (codes are case-sensitive). Until it's live, no user can approve it, so the connector simply omits the field and behavior is unchanged.
- On `start_network` (trading only), one call to `GET /account/builder_codes/approvals?account=…`; the code is injected into the signed order payload **only if it appears in the user's approvals** (exact match).
- Fails safe to omission on any error; with no approval, order payloads are byte-identical to today's.
- Hummingbot core **never** calls approve/revoke — per HGP-88 that belongs to the application UI layer.

### Follow-up work in Condor (separate repo)

Full monetization requires a `ConnectPacifica` flow in Condor — the Solana analog of the Hyperliquid wallet-connect flow (hummingbot/condor#122):

1. Browser wallet discovery via **Wallet Standard** (MetaMask/Phantom Solana accounts; EIP-6963 doesn't apply).
2. The user's **main wallet** signs the `approve_builder_code` payload — canonical recursively-sorted compact JSON per Pacifica's signing spec — with `builder_code: "hummingbot"` and Condor's chosen `max_fee_rate`; POST to `/account/builder_codes/approve`. Important constraint verified live: **agent-signed approvals are rejected** (400 "signature does not match signer and canonical payload") — approval is strictly a main-wallet action, which is exactly why it can't live in the connector.
3. Agent-wallet setup + credential handoff to hummingbot-api as usual (verified live: agent-signed *orders* do honor the account-level approval).
[...]

Issue #8201, opened 2026-04-24
Pacifica - Spot collateral, Perp PnL, and USDC now margin into one account


nikspz:
### Describe the bug

Spot collateral, Perp PnL, and USDC now margin into one account
For the context:
Unified Margin
https://pacifica.gitbook.io/docs/trading-on-pacifica/unified-margin

Unified Margin
https://pacifica.gitbook.io/docs/api-documentation/changelog

### Steps to reproduce

1.
2.
3.

### Release version

2.14.0

### Type of installation

Source

PR #8397, opened 2026-07-27
fix: Pacifica connector crash on spot instruments and lost order fills

dizpers:

- [x] Tests all pass

Split out of #8355 (Pacifica builder-code support) so bugfixes land separately from the feature. The first two bugs below were found during that PR's live mainnet verification, the third during QA of this branch; #8355 currently still contains the earlier fixes and its branch will be slimmed to builder-only separately.

Three live-found `pacifica_perpetual` bugs, plus two smaller hardening fixes and one venue-independent dependency pin:

1. **Spot-instrument crash — breaks every Pacifica user on `development` since ~April 19.** Pacifica listed its first spot market (`SOL-USDC`, `instrument_type: "spot"`) in the same `/api/v1/info` list and price feeds the connector consumes; perp-only pair parsing built `SOL-USDC-USDC` and crashed trading-rule updates, leaving the connector permanently not-ready. Fix: keep only `instrument_type == "perpetual"` entries in trading rules and the symbol map (the Pacifica team confirmed this is the officially supported discriminator; future instrument types are skipped automatically), and skip unmapped symbols in the REST and WS price feeds.
2. **A perp listed while the bot is running permanently broke trading-rule updates — found by QA on this branch after ~14h of uptime.** Pacifica listed three new perpetuals (`VVV`, `KAITO`, `kSHIB`) at 2026-07-29 07:03 UTC — confirmed by their `created_at` in `/api/v1/info` — and the next 30-minute trading-rules poll died with `KeyError: 'VVV'`. `_update_trading_rules()` formats the rules *before* it refreshes the symbol map, so a perp listed since the last poll is present in the fresh `/info` response but not yet in the map; the unguarded lookup raised, and because the exception escaped before the refresh line, the map never learned the new symbol and the failure could not self-heal. The polling loop's error path then sleeps 0.5s rather than 30 minutes, so the bot hot-spun on `/info` for the rest of the session (3928 tracebacks in 56 minutes) with trading rules frozen at their startup values. Existing pairs kept trading throughout — `self._trading_rules.clear()` only runs after formatting succeeds — so the user-visible damage was log spam, permanently stale rules, and ~2 req/s of avoidable load on the venue. Fix: skip the offending entry instead of discarding the whole batch (matching the Binance connector's `_format_trading_rules`), so the map refresh still runs and the new listing is picked up on the following poll — recovery within one interval instead of never. Same class of defect as (1): unguarded parsing of an instrument list the venue changes underneath us, so guarding the loop also covers future malformed or partial entries, not just unmapped symbols.
3. **Market-order fills could be lost, completing orders with zero amounts.** Observed live: the order-updates WS delivered `filled` instantly, but the `account_trades` WS event never arrived within the order tracker's 5s grace — the order completed with `base_asset_amount=0` and `OrderFilled` never fired. Fix: on a filled/partially-filled WS order update, if the tracker's fills don't cover the reported `f` amount, recover them via REST (zero extra requests when the WS delivered; trade-id dedup makes overlap safe).
4. **Private WS keepalive was broken since the connector shipped**: `{"op": "ping"}` is rejected by the venue with a 400 frame; correct is `{"method": "ping"}` (the public data source already used it).
5. **Trade-history query window** used the tick timestamp (floored to the whole second) as `end_time`, which can exclude just-executed fills; now wall-clock + a 2s skew buffer, deduped by trade id.
6. **Pinned a working Brotli — latent, venue-independent, and worth fixing on its own merits.** `aiohttp` advertises `br` in `Accept-Encoding` whenever the `brotli` module is importable, and **`aiohttp>=3.13.3` (2026-01-03) calls `Decompressor.process(data, max_length)`** — a second argument Brotli only accepted from **1.2.0 (2025-11-05)**. On 1.0.9 and 1.1.0 that raises `TypeError`, which aiohttp's blanket `except Exception` in `DeflateBuffer.feed_data` converts into `ContentEncodingError`, surfacing as `Error: 400, message: Can not decode content-encoding: br`. The trap is that `urllib3` pulls `brotli-python` in *transitively* with a floor of `>=1.0.9`, so the constraint is already satisfied and neither `pip install` nor `conda env update` ever upgrades it — **any environment solved before Nov 2025 decodes every `br` response into an error**, for every venue that serves brotli, not just Pacifica. Pacifica happens to make this very easy to hit because its CDN br-encodes *everything*, down to 60-byte 404 bodies. Fix: `Brotli>=1.2.0` in `setup.py` and `brotli-python>=1.2.0` in `setup/environment.yml` — note the conda package named `brotli` is the C library and CLI (~20 KB, no Python ABI tag) and does **not** provide `import brotli`; the bindings package is `brotli-python`.

   To be explicit about provenance, since this one is easy to misattribute: it was found while diagnosing a QA report of that exact error string, but **it was not the cause of that report** — QA's error disappeared when they switched to a different region, and a broken Brotli would have failed identically from every region. That error was the *other* branch of the same exception: Brotli present but the body not decodable, i.e. a mangled or mislabelled `br` response on the blocked-region path. Worth knowing when triaging, because the two causes are indistinguishable from the message alone — a **missing** Brotli gives the longer `Can not decode content-encoding: brotli (br). Please install \`Brotli\`` from `DeflateBuffer.__init__`, whereas the short `... content-encoding: br` always means Brotli was importable and the *stream* failed. This item is cleanly separable from the connector fixes if a reviewer would rather it went in its own PR.

Also refreshes stale account-info docstrings (unified-margin fields; `data` is an object) and removes a duplicated balance assignment. Balance *logic* needs no change under unified margin — `account_equity` / `available_to_spend` are venue-aggregated. Related to #8201 (suggest closing with this PR after QA).

- `pytest test/hummingbot/connector/derivative/pacifica_perpetual/` — **110 passed** (8 new: 3 spot filtering, 2 fills recovery incl. a no-extra-HTTP guard test, 1 ping format, 2 mid-session listings). flake8 + pre-commit clean.
- Bugs 1 and 3 reproduced and their fixes exercised live on mainnet (2026-07-17) during #8355's verification.
- Bug 2 was hit live by QA on 2026-07-29 and is covered by two unit tests — one on `_format_trading_rules` in isolation, one driving `_update_trading_rules()` end-to-end to assert the symbol map still gets refreshed with the new listing. Both fail with `KeyError: 'VVV'` against the previous revision. The fix has not been re-exercised against a real listing, since that needs the venue to list a perp mid-session.
- Bug 6 verified end-to-end against live `api.pacifica.fi/api/v1/info` on aiohttp 3.13.5, holding everything else constant: Brotli **1.0.9 → fails**, **1.1.0 → fails** (both with the reported message), **1.2.0 → succeeds** (22932 bytes decoded), and **Brotli absent → succeeds** via gzip, since aiohttp then does not advertise `br`. The aiohttp boundary was bisected to 3.13.3 by inspecting `BrotliDecompressor.decompress_sync` across 3.13.0–3.13.5. Brotli 1.2.0's `process()` accepts both the 1-arg and 2-arg forms, so the pin is safe across the whole `aiohttp>=3.8.5,<3.14` range.

**Tips for QA testing**:

- Connect `pacifica_perpetual` with any funded account and start any strategy. Before this fix the connector loops "Could not fetch new trading rules… `ValueError: too many values to unpack`" and never becomes ready; after it, it starts cleanly against current mainnet.
- Place a small market order (min $10 notional): `OrderFilled` fires and the completed event carries real amounts (the bug signature was `base_asset_amount=0` after a "fill updates did not arrive on time" warning).
- Bug 2 can't be triggered on demand. On a long-running bot the signature is a repeating "Could not fetch new trading rules from Pacifica_perpetual" warning with `KeyError: '<SYMBOL>'` — that should no longer appear, and a perp listed mid-session should become available within one 30-minute poll.
- If `Can not decode content-encoding: br` ever shows up again, check the environment before suspecting the connector: `python -c "import aiohttp,brotli;print(aiohttp.__version__,brotli.__version__)"`. Brotli ≥1.2.0 with aiohttp ≥3.13.3 rules out bug 6, which points at the response itself (region block, proxy, or truncated body) rather than the codec.
- Note: #8355 (builder codes) requires this fix to pass live QA — suggest merging this one first.

### DIFF
--- a/hummingbot/connector/derivative/pacifica_perpetual/pacifica_perpetual_api_order_book_data_source.py
+++ b/hummingbot/connector/derivative/pacifica_perpetual/pacifica_perpetual_api_order_book_data_source.py
@@ -380,7 +380,11 @@ class PacificaPerpetualAPIOrderBookDataSource(PerpetualAPIOrderBookDataSource):
         Next funding timestamp = :00 of next hour
         """
         for price_entry in raw_message["data"]:
-            trading_pair = await self._connector.trading_pair_associated_to_exchange_symbol(price_entry["symbol"])
+            try:
+                trading_pair = await self._connector.trading_pair_associated_to_exchange_symbol(price_entry["symbol"])
+            except KeyError:
+                # spot instruments are not in the perp symbol map
+                continue
             if trading_pair not in self._trading_pairs:
                 continue
 
--- a/hummingbot/connector/derivative/pacifica_perpetual/pacifica_perpetual_derivative.py
+++ b/hummingbot/connector/derivative/pacifica_perpetual/pacifica_perpetual_derivative.py
@@ -21,7 +21,7 @@ from hummingbot.connector.trading_rule import TradingRule
 from hummingbot.connector.utils import combine_to_hb_trading_pair
 from hummingbot.core.api_throttler.data_types import RateLimit
 from hummingbot.core.data_type.common import OrderType, PositionAction, PositionMode, PositionSide, TradeType
-from hummingbot.core.data_type.in_flight_order import InFlightOrder, OrderUpdate, TradeUpdate
+from hummingbot.core.data_type.in_flight_order import InFlightOrder, OrderState, OrderUpdate, TradeUpdate
 from hummingbot.core.data_type.order_book_tracker_data_source import OrderBookTrackerDataSource
 from hummingbot.core.data_type.trade_fee import TokenAmount, TradeFeeBase, TradeFeeSchema
 from hummingbot.core.data_type.user_stream_tracker_data_source import UserStreamTrackerDataSource
@@ -345,16 +345,27 @@ class PacificaPerpetualDerivative(PerpetualDerivativePyBase):
         rules = []
 
         for pair_info in exchange_info_dict.get("data", []):
-            rules.append(
-                TradingRule(
-                    trading_pair=await self.trading_pair_associated_to_exchange_symbol(symbol=pair_info["symbol"]),
-                    min_order_size=Decimal(pair_info["lot_size"]),
-                    min_price_increment=Decimal(pair_info["tick_size"]),
-                    min_base_amount_increment=Decimal(pair_info["lot_size"]),
-                    min_notional_size=Decimal(pair_info["min_order_size"]),
-                    min_order_value=Decimal(pair_info["min_order_size"]),
+            # Pacifica lists spot instruments (e.g. "SOL-USDC") in the same market-info response;
+            # this connector only handles perpetuals.
+            if pair_info.get("instrument_type", "perpetual") != "perpetual":
+                continue
+            # A single malformed or not-yet-mapped entry must not discard the whole batch. The symbol
+            # map is only refreshed *after* this method returns (see `_update_trading_rules`), so a perp
+            # listed by the venue since the last poll is still unknown here; raising would also skip
+            # that refresh and leave the map permanently stale.
+            try:
+                rules.append(
+                    TradingRule(
+                        trading_pair=await self.trading_pair_associated_to_exchange_symbol(symbol=pair_info["symbol"]),
+                        min_order_size=Decimal(pair_info["lot_size"]),
+                        min_price_increment=Decimal(pair_info["tick_size"]),
+                        min_base_amount_increment=Decimal(pair_info["lot_size"]),
+                        min_notional_size=Decimal(pair_info["min_order_size"]),
+                        min_order_value=Decimal(pair_info["min_order_size"]),
+                    )
                 )
-            )
+            except Exception:
+                self.logger().exception(f"Error parsing the trading pair rule {pair_info}. Skipping.")
 
         return rules
 
@@ -436,27 +447,34 @@ class PacificaPerpetualDerivative(PerpetualDerivativePyBase):
     async def _update_balances(self):
         """
         https://docs.pacifica.fi/api-documentation/api/rest-api/account/get-account-info
+
+        Since the unified-margin rollout, account_equity and available_to_spend are computed
+        venue-side to include LTV-adjusted spot collateral and exclude spot order locks, so they
+        remain the correct totals for a perp connector.
         ```
         {
           "success": true,
-          "data": [{
+          "data": {
             "balance": "2000.000000",
             "fee_level": 0,
             "maker_fee": "0.00015",
             "taker_fee": "0.0004",
             "account_equity": "2150.250000",
+            "cross_account_equity": "2150.250000",
+            "spot_market_value": "0",
+            "spot_collateral": "0",
             "available_to_spend": "1800.750000",
             "available_to_withdraw": "1500.850000",
             "pending_balance": "0.000000",
+            "pending_interest": "0",
             "total_margin_used": "349.500000",
             "cross_mmr": "420.690000",
             "positions_count": 2,
             "orders_count": 3,
             "stop_orders_count": 1,
-            "updated_at": 1716200000000,
-            "use_ltp_for_stop_orders": false
-          }
-        ],
+            "spot_balances": [],
+            "updated_at": 1716200000000
+          },
           "error": null,
           "code": null
         }
@@ -488,7 +506,6 @@ class PacificaPerpetualDerivative(PerpetualDerivativePyBase):
         self._account_balances.clear()
         self._account_available_balances.clear()
 
-        self._account_balances[asset] = Decimal(str(data["account_equity"]))
         self._account_balances[asset] = Decimal(str(data["account_equity"]))
         self._account_available_balances[asset] = Decimal(str(data["available_to_spend"]))
         self._fee_tier = data.get("fee_level", 0)
@@ -651,8 +668,12 @@ class PacificaPerpetualDerivative(PerpetualDerivativePyBase):
         else:
             start_time = int(order.creation_timestamp * 1000)
 
-        current_time = self.current_timestamp
-        end_time = int(current_time * 1000)
+        # current_timestamp is the clock tick, floored to the whole second — a fill that happened
+        # later within the same second would fall outside the window. Use the wall clock plus a
+        # small buffer for venue clock skew instead; overlapping windows are safe because the
+        # order tracker dedups fills by trade_id.
+        current_time = time.time()
+        end_time = int((current_time + 2) * 1000)
 
         params = {
             "account": self.user_wallet_public_key,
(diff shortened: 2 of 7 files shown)


nikspz:
commit d7b1007

reviewed 
- that issue with order placement fixed (could place orders now). ✅
- filled order events: ok ✅

@dizpers Could you please check why perpetualMM flooded with 

[logs_conf_perpetual_market_making_1.log](https://github.com/user-attachments/files/30503454/logs_conf_perpetual_market_making_1.log)

[8397pacifica.zip](https://github.com/user-attachments/files/30503521/8397pacifica.zip)

```
2026-07-29 07:54:49,356 - 1979575 - hummingbot.connector.derivative.pacifica_perpetual.pacifica_perpetual_derivative.PacificaPerpetualDerivative - NETWORK - Unexpected error while fetching trading rules.
  File "/home/hummingbot/nikita/pacifica8397/hummingbot/connector/exchange_py_base.py", line 772, in _trading_rules_polling_loop
    await safe_gather(self._update_trading_rules())
  File "/home/hummingbot/nikita/pacifica8397/hummingbot/core/utils/async_utils.py", line 22, in safe_gather
    return await asyncio.gather(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/hummingbot/nikita/pacifica8397/hummingbot/connector/exchange_py_base.py", line 895, in _update_trading_rules
    trading_rules_list = await self._format_trading_rules(exchange_info)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/hummingbot/nikita/pacifica8397/hummingbot/connector/derivative/pacifica_perpetual/pacifica_perpetual_derivative.py", line 354, in _format_trading_rules
    trading_pair=await self.trading_pair_associated_to_exchange_symbol(symbol=pair_info["symbol"]),
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "hummingbot/connector/exchange_base.pyx", line 108, in trading_pair_associated_to_exchange_symbol
    return symbol_map[symbol]
  File "/home/hummingbot/anaconda3/envs/hummingbot/lib/python3.13/site-packages/bidict/_base.py", line 524, in __getitem__
    return self._fwdm[key]
           ~~~~~~~~~~^^^^^
KeyError: 'VVV'
```

nikspz:
> 
> @dizpers Could you please check why perpetualMM flooded with
> 
> ```
> 2026-07-29 07:54:49,356 - 1979575 - hummingbot.connector.derivative.pacifica_perpetual.pacifica_perpetual_derivative.PacificaPerpetualDerivative - NETWORK - Unexpected error while fetching trading rules.
> Traceback (most recent call last):
>   File "/home/hummingbot/nikita/pacifica8397/hummingbot/connector/exchange_py_base.py", line 772, in _trading_rules_polling_loop
>     await safe_gather(self._update_trading_rules())
>   File "/home/hummingbot/nikita/pacifica8397/hummingbot/core/utils/async_utils.py", line 22, in safe_gather
>     return await asyncio.gather(*args, **kwargs)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/hummingbot/nikita/pacifica8397/hummingbot/connector/exchange_py_base.py", line 895, in _update_trading_rules
>     trading_rules_list = await self._format_trading_rules(exchange_info)
>                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/hummingbot/nikita/pacifica8397/hummingbot/connector/derivative/pacifica_perpetual/pacifica_perpetual_derivative.py", line 354, in _format_trading_rules
>     trading_pair=await self.trading_pair_associated_to_exchange_symbol(symbol=pair_info["symbol"]),
>                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "hummingbot/connector/exchange_base.pyx", line 108, in trading_pair_associated_to_exchange_symbol
>     return symbol_map[symbol]
>   File "/home/hummingbot/anaconda3/envs/hummingbot/lib/python3.13/site-packages/bidict/_base.py", line 524, in __getitem__
>     return self._fwdm[key]
>            ~~~~~~~~~~^^^^^
> KeyError: 'VVV'
> ```


I think should be related to something with VVV token

image

dizpers:
image

###  **tl;dr So basically a lot of connectors are using Brotli (not gzip) to encode the responses. Brotli decoding isn't working correctly (for all connectors), package updates will fix it.**

@nikspz — on the error you hit, because the message is actively misleading:

`Can not decode content-encoding: br` does **not** mean Brotli was missing. aiohttp raises two different errors here. A missing Brotli gives the longer `Can not decode content-encoding: brotli (br). Please install \`Brotli\`` from `DeflateBuffer.__init__`. The short form you got comes from `feed_data`, and only fires when Brotli *was* importable and the response body itself failed to decompress.

Pacifica's CDN br-encodes every response, including small error bodies — even a 60-byte 404 comes back `content-encoding: br`. So when the request from the original region returned something other than a normal 200 (a geo-block page, a WAF challenge, or a body truncated in transit), it still arrived labelled `br`, the stream didn't decompress, and you saw a codec error instead of the real reason. That fits it disappearing once you ran on a server in a different region: the underlying problem looks region-specific, and this error simply masked it — you never got to see what the venue actually replied.

One other cause produces a byte-identical message: Brotli 1.0.9/1.1.0 with aiohttp ≥3.13.3, where aiohttp passes a second argument older Brotli doesn't accept. That would have failed from *every* region, so the run on a server in a different region mostly rules it out — but it's a real latent bug, and it's why the Brotli pin is in this PR. If you can still run `python -c "import aiohttp,brotli;print(aiohttp.__version__,brotli.__version__)"` in that environment, it settles which of the two you were looking at.

cardosofede:
hey @dizpers the official lib is brotli not brotli-python as I see in pypi

dizpers:
@cardosofede

setup.py has this line:

```
"Brotli>=1.2.0",
```

this would be loaded from PyPI

environment.yml:

```
  - brotli-python>=1.2.0  # aiohttp>=3.13.3 calls Decompressor.process(data, max_length); Brotli<1.2.0 takes only one arg
```

this would be loaded from conda, where

brotli is a C library + CLI
brotli-python is actual python package

nikspz (approved):
> image
> ### **tl;dr So basically a lot of connectors are using Brotli (not gzip) to encode the responses. Brotli decoding isn't working correctly (for all connectors), package updates will fix it.**

reviewed that issue was fixed on server which previously can't handle it

- now pacifica_perpetual connected successfully ✅
- placed orders successfully ✅
- status, order_book --live matched with exchange ✅
- filled orders handling ✅

image

cardosofede:
okey but why not use the same lib installed from pip and add it as a python dep

cardosofede:
nvm I just checked the condo lib and it's anaconda verified

cardosofede (approved):

PR #8438, opened 2026-08-26
feat(gateway): trim the client's Gateway commands and fix their autocomplete

fengtality:
## Summary

Two things:

1. Trims the interactive client's Gateway command surface — LP and trading actions come out; connection, configuration, and balance/allowance stay. `gateway ping` folds into the bare `gateway` command.
2. Fixes five defects in the gateway autocomplete found while verifying the remaining commands. All five pre-date this PR.

Gateway itself and every Gateway-backed connector, script, strategy, and executor are untouched.

## Command surface after this PR

```
gateway                          - test node and chain/network status   <-- was `gateway ping`
gateway allowance <connector>    - check token allowances
gateway approve <connector> <t>  - approve tokens for spending
gateway balance [chain] [tokens] - check token balances
gateway config [namespace]       - show configuration
gateway config <ns> update [<path> <value>]
gateway connect <chain>          - view and add wallets for a chain
gateway generate-certs           - generate SSL certificates
gateway list                     - list available connectors
```

## 1. Removed (`577f67e`)

LP and trading actions belong to Gateway routes, HAPI, or Condor:

- `gateway lp`, `gateway pool`, `gateway swap`, `gateway token` — modules, parser entries, completers, tests
- `lp_command_utils.py`
- `GatewayCommandUtils.calculate_and_display_fees` and `.prompt_for_percentage` — the only two of the class's 13 helpers left without a caller once the LP command was gone

**`gateway ping` folds into bare `gateway`.** The `[chain]` filter is not carried over: an optional positional on the root parser would shadow the remaining subcommands (argparse would bind `gateway balance` to `chain="balance"`). Bare `gateway` reports every chain against its default network, which is what `gateway ping` with no argument already did. Discovery moves to `help gateway` and tab-completion.

## 2. Autocomplete fixes (`8d3c1df`)

Exercised the completer against populated `GATEWAY_CHAINS` / `GATEWAY_NAMESPACES` / `GATEWAY_ETH_DEXS`:

| Defect | Before | After |
|---|---|---|
| `gateway connect eth⇥` — chain completion only worked on an empty argument (predicate guarded on `len(text.split()) <= 2`) | `[]` | `['ethereum']` |
| `gateway approve uniswap/amm ⇥` — re-offered connectors where a token symbol belongs | connectors | `[]` |
| `gateway balance ethereum ⇥` — re-offered chains where a token list belongs | chains | `[]` |
| `gateway list ⇥`, `gateway generate-certs ⇥` — re-offered the whole subcommand list | subcommands | `[]` |
| `_gateway_namespace_completer`, `_gateway_connect_completer` built and never referenced | dead | removed |
| `_complete_gateway_chain`/`_network`/`_wallet_addresses` keyed off prompt text (`"Which chain do you want"`, `"Select a gateway wallet"`) that exists nowhere; feeders `set_gateway_chains`/`set_gateway_networks`/`set_list_gateway_wallets_parameters` had no callers | never fired | removed |

The six per-subcommand predicates collapse into one `_gateway_completer_for` resolver driven by a subcommand → positional-completer map, dispatched from a single **terminal** branch so an argument-less subcommand can no longer fall through to the generic subcommand completer. Adding completion for a new argument is now one line in the map.

```python
self._gateway_arg_completers = {
    "allowance": [self._gateway_eth_connector_completer],
    "approve":   [self._gateway_eth_connector_completer],
    "balance":   [self._gateway_chain_completer],
    "config":    [self._gateway_namespace_completer, self._gateway_config_action_completer],
    "connect":   [self._gateway_chain_completer],
}
```

## 3. Config error message (`7cffb24`)

`gateway config <namespace>` reported `Connection to Gateway localhost:15888 failed` for a namespace that does not exist — the connection was fine. Gateway answers an unknown namespace with a 404 that the client already turns into a `GatewayError` carrying the detail, but the handler collapsed every exception into the connection-failure branch. Verified against a live Gateway:

| Input | Before | After |
|---|---|---|
| `gateway config not-a-namespace` | `Connection to Gateway localhost:15888 failed` | `Namespace 'not-a-namespace' not found` |
| `gateway config <ns>`, port closed | `Connection to Gateway localhost:15999 failed` | unchanged |

`_update_gateway_configuration` swallowed the same way, and its "See log file for more details" never actually logged. It now surfaces Gateway'''s message and logs the unexpected case.

## 4. `gateway approve` was broken on every EVM network (`12abcdd`)

With a funded wallet, `gateway approve <connector> <token>` died before the confirmation prompt:

```
Error approving token: '''>''' not supported between instances of '''str''' and '''int'''
```

Gateway serialises the estimate-gas response as strings to preserve precision (`"fee": "0.0000438"`, `"maxFeePerGas": "0.146"`). `estimate_transaction_fee` passed them through untouched, but every consumer treats them as numbers — `display_balance_impact_table` compares `gas_fee > 0`, `display_transaction_fee_details` formats `{fee_in_native:.6f}`.

Coerced at the boundary in `estimate_transaction_fee` rather than at each call site. A non-numeric value now raises inside that method'''s existing try, surfacing as "Could not estimate transaction fees" with the reason in `result["error"]` instead of a TypeError mid-render.

After the fix, on ethereum mainnet:

```
=== Balance Impact After Approval ===
Token     Current Balance → After Transaction
  ETH            0.002410 →       0.002367
  WETH           0.001000
Transaction Fee Details:
  Max Fee Per Gas: 0.1431 gwei
  Estimated Gas Cost: ~0.000043 ETH
Do you want to proceed with the approval? (Yes/No) >>> no
Approval cancelled
```

Also exercised on **robinhoodchain** (chainID 4663) after switching `ethereum.defaultNetwork` with the `gateway config` command: 27 tokens listed, WETH/USDG already Unlimited, fees estimated at ~0.000016 ETH, prompt and cancel both clean. The default network was restored to `mainnet` afterwards.

## Live verification

All nine commands were exercised against a Gateway running on :15888 in `DEV=false` (mutual TLS) mode, invoking the real command methods and awaiting the coroutine each schedules:

| Command | Result |
|---|---|
| `gateway` (bare, folded ping) | online; ethereum/mainnet + solana/mainnet-beta both ✓ Connected |
| `gateway list` | 11 connectors rendered |
| `gateway config` / `<ns>` / `<bad ns>` | usage, full config, correct 404 message |
| `gateway balance` / `<chain>` / `<chain> <tokens>` | balances returned; chain and token filters applied |
| `gateway allowance` / `<connector>` | no-arg loops all ethereum connectors |
| `gateway connect` / `<chain>` | lists wallets, exits cleanly on option 3 |
| `gateway config <ns> update` — direct, invalid path, interactive | validation, write, and cancel all correct |
| `gateway approve` | stops at the wallet check before building any transaction |
| `gateway generate-certs` | wrote 8 files; certs backed up and restored afterwards |

## Verification

- new `test/hummingbot/client/ui/test_completer.py` covers the surface; its four regression cases **fail** against the previous implementation and pass against this one
- `grep` across the repo for every removed module, symbol, and command string: no dangling references
- `help` and `help gateway` render the surface shown above
- flake8 + isort clean
- full suite: `8492 passed, 2 failed` — both failures (`test_list_configs`, `test_price_and_size_limit_calculation_with_slippage_buffer`) reproduce unchanged on `development`


rapcmia:
PR update:
- Legacy Gateway TUI command removal 
  image

#### Gateway source configuration and wallet balance ✅ 
- Used the source Gateway instance on port 15888.
- Solana mainnet reports the configured URL-based RPC provider, a current block number, and `jupiter/router` as the swap provider.
- Added the Solana wallet to Gateway, then confirmed it was listed by the wallet endpoint.
- The added wallet returned non-zero SOL and USDC balances.

---

#### LP rebalancer executor lifecycle and Gateway polling ✅
- Ran the `lp_rebalancer` controller with `solana-mainnet-beta`, `orca/clmm`, and `SOL-USDC`.
- The controller opened range positions, received Gateway transaction signatures, and tracked each executor.
- Gateway polled the submitted transaction hashes and each observed add/remove order reached a completed state.
- Price movements crossed the configured boundaries, so the controller closed a position and created the next expected rebalancing position.
- The strategy stop closed the final position successfully; the persisted database shows zero active executors and the latest executors are terminated with `POSITION_HOLD` as their close type.

---

QA notes ❗ 
-  No longer add or manage Gateway wallets from the Hummingbot interactive client because the legacy Gateway TUI commands were removed.
- No longer use the interactive client to check gateway balance. Wallet and balance actions now belong to Gateway routes only unless user uses HAPI/Condor
- No longer use the interactive client to list supported Gateway connectors. Connector availability must be checked through Gateway routes 
- Hummingbot can still use Gateway-backed connectors and strategies, this PR removes the old command surface and does not disable Gateway trading integration.
- The tested source setup used Gateway's local HTTP endpoint on port 15888, `pnpm start --passphrase=x --dev`
- Users should update Gateway settings, such as RPC/node URLs and provider configuration, through Gateway routes

rapcmia:
Test update:
- Tested with ethereum-robinhoodchain
- Gateway command available again and worked ok, removed LP, pool, swap, and token commands were unavailable
- Gateway autocomplete, connect, config, balance, and approve suggested the expected values.
- Gateway configuration errors, an unknown namespace showed “namespace not found” instead of a connection failure.
- Gateway configuration validation, an invalid Ethereum RPC provider was safely rejected by schema validation.
- Gateway approval and allowance, gateway approve uniswap/clmm META displayed fee and balance-impact details without a type error, META showed unlimited allowance afterward.
- EVM token approval, gateway approve uniswap/clmm META displayed fee-per-gas, estimated cost, and balance impact without a type error.
- Fresh-install V2 LP rebalancer, a Gateway-backed Solana controller started, opened and closed positions, and stopped cleanly.

rapcmia (approved):

cardosofede (approved):

PR #8440, opened 2026-08-28
fix(lp_rebalancer): account for pre-existing LP positions at startup

AzothZephyr:
## Problem

`LPRebalancer` keeps no state across restarts. `executors_info` starts empty and `_initial_position_created` starts `False`, so a position opened by a previous process is invisible to the controller — and the first tick after a restart opens a second one.

The original position is then orphaned on-chain:

- nothing rebalances it,
- nothing closes it when price crosses its limit prices,
- its fees are never collected,
- and the wallet no longer holds enough to size the replacement correctly, so the new position is opened undersized or fails outright.

Any restart while a position is open reaches this — a config deploy, a crash, or a machine reboot. It is silent: the logs show a normal startup, and the only symptom is capital sitting in a pool that nothing is managing any more.

I hit this on a live Meteora DLMM deployment. A restart with ~$575 deployed left the position stranded while the controller opened a new one from the ~$48 still in the wallet.

## Change

A startup reconciliation pass. Before opening anything, the controller asks the connector which positions the wallet already holds in its configured pool, and acts on a new `startup_position_policy`:

| policy | behaviour |
| --- | --- |
| `recover` (default) | close the stray position, returning its liquidity and uncollected fees to the wallet, then start clean |
| `halt` | refuse to trade until it is dealt with manually |
| `ignore` | previous behaviour, kept as an escape hatch |

The failure paths are the point:

- **An unreadable chain is not an empty wallet.** If the position query fails, the controller waits rather than assuming a clean slate — opening blind is precisely the failure being prevented.
- **Recovery is verified, not assumed.** Positions are re-read after closing rather than trusting the close response.
- **It gives up safely.** After `startup_recover_max_attempts` failures it halts instead of opening on top of a position it could not clear.
- **The gate is startup-only.** Once reconciled, normal operation is untouched (covered by a test).

Reconciliation state appears in `to_format_status`, since it blocks trading.

## On the default

I chose `recover` because it keeps the controller autonomous, which is the reason to run it unattended, and closing a position is something this controller does routinely anyway — liquidity and fees both return to the wallet, so the cost is gas plus a few seconds out of market.

`halt` is the more conservative choice and I am happy to switch the default if maintainers prefer that a controller never take an unprompted on-chain action. The important part is that the default is no longer `ignore`.

## Testing

`test/hummingbot/strategy_v2/controllers/test_lp_rebalancer_startup_reconciliation.py` — 13 tests covering each policy, the unreadable-chain case, verification-after-close, retry-then-halt, a failing close, idempotence, config validation, and that trading resumes once reconciled.

```
13 passed
```

`flake8` clean against the repo config.

Beyond unit tests, the same logic has been running on a live mainnet Meteora SOL/USDC deployment. Restarting the bot with a position open now produces:

```
Startup reconciliation: recovering 1 pre-existing position(s) (attempt 1/3): Dg2Fw4BZ... [105.573239-108.093890] 2.5107 base + 306.74 quote (uncollected fees 0.00037/0.0529)
Startup reconciliation: closed stray position Dg2Fw4BZ... signature=4PRb6TXmPJ8Cc6JEx7ss8aaTRdDZZvG3rcCRiL3ftvnB59TacRahdW7PiR9XF2cVJjqAV1q7VuFymwfVhSpaJGCM
Startup reconciliation: all stray positions closed; capital returned to the wallet
Position created: 3eD8SVhf... base 2.689679 quote 287.499525
```

Full capital recovered and redeployed in about three seconds, where previously it would have been stranded.

## Possible follow-up

`LPExecutorConfig` has no field for an existing position, so an executor can only ever create one — which is why `recover` closes and reopens rather than resuming. Adding an `existing_position_address` and an adoption path would let a restart carry on managing the live position with no round-trip cost or time out of market. That touches executor state initialisation and P&L baselines for a position whose entry amounts are unknown, so it seemed better as its own change than bundled here. Happy to pick it up if there is interest.

### DIFF
--- a/controllers/generic/lp_rebalancer/lp_rebalancer.py
+++ b/controllers/generic/lp_rebalancer/lp_rebalancer.py
@@ -1,6 +1,6 @@
 import logging
 from decimal import Decimal
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Tuple
 
 from pydantic import Field, field_validator, model_validator
 
@@ -266,9 +266,11 @@ class LPRebalancer(ControllerBase):
         self._startup_reconciled: bool = False
         self._startup_halted: Optional[str] = None
         self._startup_recover_attempts: int = 0
-        # position address -> timestamp of the close submitted for it, so a
-        # close that has not settled yet is waited on rather than re-submitted
-        self._startup_closes_in_flight: Dict[str, float] = {}
+        # position address -> (timestamp of the last close attempt, whether the
+        # submission was accepted). Both outcomes pace the next attempt: a close
+        # that has not settled must not be re-submitted, and a submission that
+        # failed must not be retried at the controller's tick rate either.
+        self._startup_close_attempts: Dict[str, Tuple[float, bool]] = {}
         self._startup_last_wait_log: float = 0.0
 
         # Initialize rate sources
@@ -1010,9 +1012,9 @@ class LPRebalancer(ControllerBase):
 
         if not positions:
             self._startup_reconciled = True
-            if self._startup_closes_in_flight:
+            if self._startup_close_attempts:
                 self._pending_balance_update = True
-                self._startup_closes_in_flight.clear()
+                self._startup_close_attempts.clear()
                 self.logger().info(
                     "Startup reconciliation: all stray positions closed; capital returned to the wallet."
                 )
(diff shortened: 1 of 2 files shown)


AzothZephyr:
Thanks — both P1s were fair, and both are now fixed in 7f94bf6.

**Pending closes exhausting attempts** was worse than it first looks. Reconciliation runs on the controller's `update_interval`, which defaults to **one second**, and each pass counted as an attempt. So an unconfirmed close would be re-submitted every second and burn all three attempts within about three seconds of startup — permanently halting a controller whose close had actually succeeded. It only worked in my testing because Gateway's Meteora close happens to block until confirmation, which is connector-specific luck rather than a guarantee.

Submitted closes are now tracked per position address and given `startup_close_confirm_timeout` (default 120s) to settle before being treated as failed, so attempts measure real failures instead of latency. No re-submission happens while a close is in flight, and a submission that throws leaves nothing in flight and doesn't consume an attempt. Waiting states are log-throttled for the same tick-rate reason.

**On ownership, the default is now `halt`.**

I went looking for a way to scope recovery to positions this controller opened, and I don't think one exists today. Gateway reports positions per wallet, there's no on-chain marker tying a position to the workflow that opened it, and the obvious source of truth — the executors table, which does record `controller_id` alongside `position_address` — is only written on a clean shutdown via `store_all_executors()`. A crash leaves no record, and a crash is precisely the case this feature exists for. So any attribution I could implement would be reliable in exactly the situations that don't need it.

Rather than infer ownership from something weaker (matching price bands, assuming sole occupancy), the default now does the thing that is always safe: `halt` never touches funds, and it still prevents the second position that orphans the first, which is the actual bug. Stopping is a real cost for an unattended market maker, but it is a recoverable one, and quietly withdrawing another workflow's liquidity is not.

`recover` remains available for a dedicated wallet, and is now narrower: it refuses to act when it finds more than one position in the pool, since that is direct evidence the wallet is shared. Its config description says plainly that it acts on whatever the wallet holds in that pool, not only on positions this controller opened.

If maintainers would prefer this to go further — a `controller_id` recorded against opened positions so ownership becomes verifiable, or dropping `recover` from this PR entirely and shipping only the gate — I'm happy either way. My one strong opinion is that the default should not remain `ignore`, since that is the behaviour that silently loses funds.

Tests are now 16, covering the multi-position refusal, in-flight closes not re-submitting or consuming attempts, re-submission after the timeout, halting after max attempts, and a failed submission leaving state clean. `flake8` clean against the repo config.

One note on CI: the `ci` workflow is sitting at `action_required`, so the suite hasn't run here — it needs a maintainer to approve workflows for a first-time contributor.

PR #8447, opened 2026-09-01
(perf) bound the controller's executor view in backtesting, and report closing-tick executors

cardosofede:
## Problem

`BacktestingEngineBase.update_executors_info()` runs once per tick and assigned:

```python
self.controller.executors_info = active_executors_info + self.stopped_executors_info
```

`stopped_executors_info` grows monotonically for the whole run, so each tick allocated and copied a list whose length is "every executor created so far", and every controller that iterates `executors_info` paid that growing cost again. Total work is **O(n²)** in the number of executors.

This was found while diagnosing an incident where a backtest pinned a `hummingbot-api` process at 100% CPU for hours. A grid-style market maker creating ~8 executors per tick over a 30-day `1m` window (~43k ticks) never completed; `py-spy` showed the tick loop buried in the controller's scan of that list.

## What changed

**1. `(perf)` — the ledger and the controller's view are no longer the same list.**

- The **ledger** stays complete and is what `simulate_execution()` returns, via the new `collect_executors_ledger()`. `summarize_results()` and the `"executors"` payload are untouched.
- The **view** (`controller.executors_info`) is now the active executors plus a `deque` of those that terminated within `terminated_executors_window` seconds of *simulated* time, default one hour. `BacktestingEngineBase(terminated_executors_window=None)` restores the full history.

The window is **time-based rather than a fixed count**, because that is how controllers actually read terminated executors: `MarketMakingControllerBase.get_levels_to_execute()` looks back over `cooldown_time` for `STOP_LOSS` closes, and `DirectionalTradingControllerBase.can_create_executor()` over `cooldown_time` too. One hour covers both bases' defaults (15s and 300s) with wide headroom and stays O(1) in run length. A fixed count of 20-40 would silently break them.

Every close type is kept in the window, **`POSITION_HOLD` included**. It is tempting to drop them since their exposure is already tracked by the position-hold ledger and surfaced through `controller.positions_held`, but controllers read more than exposure off a terminated executor: a market maker built on `OrderExecutor` sees a *filled* maker order as a `POSITION_HOLD` termination, and that is the event its cooldown keys off before re-quoting the level.

**2. `(fix)` — executors created on the closing tick now reach the results.**

`simulate_execution()` returned the snapshot taken by the last `update_executors_info()`, which runs at the *start* of a tick, so an executor the controller created on the final tick went missing from the results while `_cumulative_volume` / `_executor_realized_pnl` had already booked it. The ledger is now rebuilt from the two places an executor can live — `active_executor_simulations` regenerated at the run's last timestamp, plus `stopped_executors_info` — which are disjoint, so nothing is counted twice.

Kept as a separate commit because, unlike the first, it **changes reported numbers**.

## Results

Reference backtests: `bollinger_v1` over mocked Binance klines, pinned window, so runs are deterministic.

Commit 1 — every `summarize_results` key identical before and after:

| run | executors | before | after | controller view (max) |
|---|---|---|---|---|
| 3d, 1 exec/side | 17 | 0.19s | 0.15s | 17 → 3 |
| 30d, 20 exec/side, tl=300s | 4 368 | 6.68s | 6.08s | 4 368 → 63 |
| 15d, dense signals, tl=60s | 21 302 | 35.86s | 23.10s | 21 302 → 61 |
| 30d, dense signals, tl=60s | 42 902 | 101.63s | 45.76s | 42 902 → 61 |

Doubling the window (15d → 30d) costs **2.83x** before and **2.03x** after: quadratic becomes linear. The controller's per-tick view is flat at 61 entries instead of growing to 42 902.

Commit 2 — no-op on runs that open nothing on the closing tick, recovers exactly the missing executor where they do:

| run | executors | net_pnl_quote |
|---|---|---|
| 3d, 1 exec/side | 17 → 17 | unchanged |
| 30d, 20 exec/side, tl=300s | 4 368 → 4 368 | unchanged |
| 30d, dense signals, tl=60s | 42 902 → **42 903** | -1202.1556 → **-1202.1756** |

## Live trading is unaffected

The change is confined to `BacktestingEngineBase`. In live, `controller.executors_info` is assigned by `StrategyV2Base.update_executors_info()` from `ExecutorOrchestrator.get_all_reports()`, which never goes through this engine.

## Tests

- `test_backtesting_executors_view.py` — bounded view vs. complete ledger, window contents and ordering, `POSITION_HOLD` retention, `window=None` escape hatch, default window.
- `test_backtesting_final_tick_ledger.py` — created-on-final-tick reaches the ledger and `summarize_results`, no duplicates, booked executors present with close info, stale-running-snapshot case.
- `test/hummingbot/strategy_v2/` — 631 passed. `pre-commit` clean.

## Notes for reviewers

- **Is one hour the right default?** It was chosen against the two shipped controller bases and the controllers in `controllers/`. A controller that genuinely needs deeper history would see a behaviour change; `terminated_executors_window=None` is the escape hatch, and the value is a constructor argument so callers can raise it.
- The mirror-image case of commit 2 — an executor *stopped* on the closing tick — is handled by the same code but is unreachable through a full run today: `_get_executor_max_timestamp()` caps every simulation's dataframe at the last index, so `update_executors_info()` has already terminated all of them before that tick's `determine_executor_actions()` runs. It is covered by a unit test that drives the engine methods directly.
- Unrelated pre-existing bug spotted while writing the tests, **not** fixed here: `summarize_results()`'s fallback drawdown branch (the one used when no `pnl_timeseries` is passed) raises `TypeError: unsupported operand type(s) for /: 'float' and 'decimal.Decimal'`, because `net_pnl_quote` is a `Decimal` and `max_draw_down` is cast to `float`. Dead for `run_backtesting()`, which always passes a timeseries, but it bites callers that summarize a bare executor list.

https://claude.ai/code/session_01AtaZ2cJgwwhkbjXJ6QbeDi

### DIFF
--- a/hummingbot/strategy_v2/backtesting/backtesting_engine_base.py
+++ b/hummingbot/strategy_v2/backtesting/backtesting_engine_base.py
@@ -292,7 +292,7 @@ class BacktestingEngineBase:
 
         # Final flush: convert any last-tick POSITION_HOLD executors into position holds
         self._update_positions_from_stopped_executors()
-        return self.collect_executors_ledger()
+        return self.collect_executors_ledger(last_index)
 
     def _reset_simulation_state(self):
         """Reset every accumulator a single backtesting run writes to."""
@@ -305,23 +305,26 @@ class BacktestingEngineBase:
         self.pnl_timeseries: List[Dict] = []
         self._executor_realized_pnl = 0.0
         self._cumulative_volume = 0.0
-        # Snapshot of the run ledger taken on every tick, see collect_executors_ledger().
-        self._ledger_active_executors_info: List[ExecutorInfo] = []
-        self._ledger_stopped_count: int = 0
         # Time-windowed view of terminated executors handed to the controller: the oldest ones
         # fall out of the left end. ``None`` means the controller gets the whole ledger.
         self._terminated_executors_view: Optional[Deque[ExecutorInfo]] = (
             None if self.terminated_executors_window is None else deque())
 
-    def collect_executors_ledger(self) -> List[ExecutorInfo]:
-        """The run ledger: the active and terminated executors of the whole run.
+    def collect_executors_ledger(self, timestamp: float) -> List[ExecutorInfo]:
+        """The run ledger at ``timestamp``: every executor of the run, terminated or not.
 
         This is what ``simulate_execution`` returns and what the results are summarized from, so
-        it must stay complete regardless of how much of it the controller gets to see. It is the
-        snapshot taken by the last ``update_executors_info()`` — the very list the controller used
-        to be handed before its view was bounded — so the reported metrics are unchanged.
+        it must stay complete regardless of how much of it the controller gets to see. It is built
+        from the two places an executor can live rather than from the view the controller was last
+        handed, so executors the controller created or stopped on the closing tick — after that
+        tick's ``update_executors_info()`` had already run — are reported too.
+
+        A simulation is either still in ``active_executor_simulations`` or has been moved to
+        ``stopped_executors_info``, never both, so nothing is counted twice.
         """
-        return self._ledger_active_executors_info + self.stopped_executors_info[:self._ledger_stopped_count]
+        still_simulating = [simulation.get_executor_info_at_timestamp(timestamp)
+                            for simulation in self.active_executor_simulations]
+        return still_simulating + self.stopped_executors_info
 
     def _record_terminated_executor(self, executor_info: ExecutorInfo):
         """Add a terminated executor to the complete ledger and to the controller's bounded view.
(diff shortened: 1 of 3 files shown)


rapcmia (approved):
- Long dense HAPI backtest, completed a 30-day one-minute Bollinger V1 run in approximately 16.1 seconds, returned all 897 unique executors, and left HAPI responsive with stable CPU and memory.
- Development comparison, produced identical executor records after generated IDs were removed, with matching PnL, volume, fees, direction totals, and final chart values.

---

❗ During my AI-assisted review, a potential POSITION_HOLD edge case was identified that I haven’t tested directly. It may be worth validating for chart and risk-metric consistency.

PR #8449, opened 2026-09-05
fix(hyperliquid): omit the builder field when the fee resolves to 0

AzothZephyr:
## Problem

Every order fails for any user who has not approved the Hummingbot Foundation builder:

```
Error submitting order 0x4f111ae6...: Builder fee has not been approved.
Failed to submit SELL order to Hyperliquid_perpetual. Check API key and network connection.
```

The connector is unusable on mainnet until the user approves a builder they may not know about and may not want to pay.

## Cause

`_initialize_builder_fee()` resolves the fee correctly — `min(approved, configured)`, which is `0` when `maxBuilderFee` returns `0`:

```python
self._builder_fee_tenths_bps = min(approved_max_tenths_bps, CONSTANTS.FOUNDATION_BUILDER_FEE_TENTHS_BPS)
```

But `_build_builder_field()` attaches the field regardless of the resolved fee:

```python
return {"b": self._builder_address.lower(), "f": self._builder_fee_tenths_bps}   # f == 0
```

Hyperliquid rejects an order carrying a builder the sender has not approved **even when the fee is 0**. The surrounding comments assume otherwise — "The fee only takes effect if the user has approved this builder in Condor; otherwise it is 0 bps" — which holds for the fee but not for the order.

## Fix

Omit the field when the resolved fee is 0. That matches what the existing comments already assume the behaviour to be, and it collects nothing either way.

An approved user is unaffected: the fee is non-zero and the field is attached exactly as before. Vault and testnet handling is untouched.

## Tests

Two existing tests asserted the field was still built at `f=0`, one of them noting *"Still attributes to the Foundation builder address, just at 0 bps"* — that is precisely the assumption the venue rejects. Both now assert the field is omitted.

Added:
- `test_builder_field_omitted_when_fee_is_zero`
- `test_place_order_omits_builder_key_when_fee_is_zero` — asserts the key is **absent** from the signed action rather than present with a zero value, driving the real `_place_order` path

`148 passed` for the full connector suite.

## Verification

Reproduced on mainnet with an account where `maxBuilderFee` returns `0`: every order failed with the message above. With the field omitted, orders submit and fill normally.

### DIFF
--- a/hummingbot/connector/derivative/hyperliquid_perpetual/hyperliquid_perpetual_derivative.py
+++ b/hummingbot/connector/derivative/hyperliquid_perpetual/hyperliquid_perpetual_derivative.py
@@ -681,8 +681,15 @@ class HyperliquidPerpetualDerivative(PerpetualDerivativePyBase):
 
     def _build_builder_field(self) -> Optional[Dict[str, Any]]:
         """The ``{"b": <address>, "f": <tenths_of_bps>}`` order field, or None when omitted. Address
-        is lowercased (the venue rejects mixed-case)."""
+        is lowercased (the venue rejects mixed-case).
+
+        Omitted when the resolved fee is 0, i.e. the user has not approved this builder (or the
+        lookup failed). Hyperliquid rejects an order carrying a builder the user has not approved
+        - "Builder fee has not been approved." - regardless of the fee being 0, so sending the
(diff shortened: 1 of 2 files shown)


fengtality:
Builder fee at 0 is used for volume attribution

Issue #8437, opened 2026-08-26
Regression: Binance BNB fee conversion with default Gate.io rate oracle causes continuous error storm


jilongjia:
### Describe the bug

When running Grid Executors on Binance Spot with trading fees paid in BNB, Hummingbot repeatedly fails to convert the BNB fee into USDT.

The generated `conf_client.yml` uses `gate_io` as the default rate oracle source:

rate_oracle_source:
  name: gate_io

After an order is filled and its fee is paid in BNB, the following exception is generated continuously:

ValueError: Could not find the exchange rate for BNB-USDT using the rate source gate_io rate oracle (please verify it has been correctly configured)

With two running Grid Executors, I observed approximately 1,102 occurrences and 12,240 log lines per minute. The repeated fee calculation consumes unnecessary CPU and produces very large amounts of Docker and database traffic. It may also cause cumulative fees and net PnL to be reported incorrectly.

Expected behavior:
- Hummingbot should successfully obtain the BNB-USDT conversion rate;
- or fall back to the active Binance connector/rate source;
- and should not retry and log the same conversion failure thousands of times per minute.

This appears related to previously closed issues #6964 and #6927.

### Steps to reproduce

1. Install the latest Hummingbot API using Docker.
2. Create an account using the generated default `conf_client.yml`, with `rate_oracle_source.name` set to `gate_io` and the global token set to `USDT`.
3. Connect a Binance Spot account that uses BNB to pay trading fees.
4. Start a Grid Executor on a Binance USDT spot trading pair.
5. Wait for an order to be filled and for Binance to charge the trading fee in BNB.
6. Check the Hummingbot API Docker logs.
7. Observe that `Error calculating fee paid in USDT` and the BNB-USDT rate-oracle exception are logged repeatedly on every update.

### Release version

20260729 (Hummingbot API 1.0.1)

### Type of installation

Docker

### Attach required files

[hummingbot_github_issue_bnb_fee_error.log](https://github.com/user-attachments/files/31470876/hummingbot_github_issue_bnb_fee_error.log)
[hummingbot_github_issue_conf_client_sanitized.yml](https://github.com/user-attachments/files/31470877/hummingbot_github_issue_conf_client_sanitized.yml)

matvt-cell:
Hi @jilongjia, I'd like to work on this. Before I start implementing, could a maintainer confirm this isn't already being addressed in another branch or PR?

[...]

PR #8453, opened 2026-09-10
fix: improve third-token fee conversion

matvt-cell:
## Summary

Addresses the third-token fee conversion path involved in #8437.

- allows `InFlightOrder.cumulative_fee_paid()` to use an explicitly supplied rate source
- lets GridExecutor use a strategy-provided market data provider for fee conversion when available
- rate-limits repeated fee-conversion exception logging for the same token/rate source to prevent continuous log storms
- continues attempting conversion on every call, so a newly available rate is used immediately
- clears log-throttle state after successful conversion and when a new fill arrives

This allows API executors to resolve fees such as BNB -> USDT through their market-data provider instead of relying exclusively on the legacy RateOracle path.

No partial cumulative fee value is cached across calls.

## Tests

- third-token BNB fee converts through a supplied rate source
- repeated unavailable-rate failures continue attempting conversion while duplicate logs are throttled
- conversion recovers immediately when the rate becomes available
- GridExecutor uses `strategy.market_data_provider` for fee conversion
- affected core suites: 66 passed

## Companion change

The companion API bridge is hummingbot/hummingbot-api#232, which exposes its existing `MarketDataService` through `AccountTradingInterface.market_data_provider`.

Related to #8437


matvt-cell:
Addressed the partial-fee caching finding in e04e14b69. Cached incomplete fee totals and conversion retry suppression were removed. Fee conversion is now attempted on every call so recovery is immediate when a rate becomes available; only repeated exception logging is rate-limited to 30 seconds. The affected suites remain green: 66 passed.

matvt-cell:
Ive addressed the original #8437 regression and the subsequent fee-completeness/risk-control findings, with the affected suites currently green (62 tests).

The latest automated review is now reaching broader GridExecutor forced-shutdown/accounting semantics: base-fee exposure handling and how completed orders with unresolved fee valuation should be persisted at the forced-shutdown boundary.

Before expanding this PR further, could a maintainer please review the current scope and advise whether these remaining forced-shutdown cases need to be handled in #8453 or should be follow-up work? I’d like to keep this fix scoped and mergeable.

Issue #8457, opened 2026-09-14
Order tracking - v2 strategy orders accepted by the exchange are dropped during a network outage, leaving an untracked position


rapcmia:
### Describe the bug

While monitoring a long-running bot, six sell orders were accepted by the exchange and received exchange order IDs before a DNS outage interrupted status, trade-update, and cancellation requests. Hummingbot later marked these orders as `OrderFailure` and removed them from lost-order tracking. The removal happened immediately after another DNS error, not after the exchange confirmed that the orders were cancelled, filled, or not found.

After the bot was stopped, the exchange account showed an accumulated `0.0006 BTC` short position that remained open. This exactly matches the combined size of the six lost sell orders (`6 x 0.0001 BTC`), while SQLite contains no TradeFill for any of them. This is strong evidence that the lost orders created untracked exchange exposure, although the exchange order and trade history is still needed to map each fill conclusively to the six order IDs.

<img width="1450" height="167" alt="Image" src="https://github.com/user-attachments/assets/5026ed7b-8852-46c0-89b7-2c048da0ea97" />

The behavior appears to come from shared Hummingbot order-tracking code rather than the kalshi_perpetual connector or the V2 controller. The active-order status handler sends generic exceptions, including network errors, to the order-not-found counter. If another polling task has already moved the order into lost-order tracking, a later network exception can remove it as though the exchange had confirmed it was missing. The same affected code is present on the tested branch, origin/development, and origin/master; Kalshi was only the connector where the behavior was observed.

---

The following sequence shows one order being accepted with an exchange order ID, receiving repeated DNS errors, being marked failed locally, and then being removed as “not found” immediately after another DNS error:

```text
2026-09-12 11:14:05,411 - 1891380 - hummingbot.core.event.event_reporter - EVENT_LOG - {"timestamp": 1789182845.0, "type": "OrderType.LIMIT_MAKER", "trading_pair": "BTC-USD", "amount": "0.000100", "price": "77328", "order_id": "HBOTSBCUD65b4097ec9796097bcd04a7b4df", "creation_timestamp": 1789182845.0, "exchange_order_id": "01a0939b-a048-7f47-b45a-0bf74612a485", "leverage": 3, "position": "OPEN", "event_name": "SellOrderCreatedEvent", "event_source": "kalshi_perpetual"}
[...]

PR #8461, opened 2026-09-17
(fix) race condition in api throttler acquire()

mnaamani:
## Problem - Race condition
`AsyncRequestContextBase.acquire()` checks capacity under the lock, releases it, then takes the lock again to record the task. When several requests are queued on the lock, releasing the lock between the two steps hands it to the next waiter, which runs its own capacity check before the first task is recorded.
Both see the same free capacity and both are admitted.

## Risk
Example scenario when a strategy/script makes a burst of order placements, api requests may exceed rate limit capacity, and risk client exceeding rate limits and potentially getting ip address blocked for some period of time.

## Fix
Under the same lock, do the capacity check and log the acquired rate limit into the tasks log.

## Tests
Added unit test `test_concurrent_acquirers_cannot_overbook_a_limit` in `test/hummingbot/core/api_throttler/test_async_throttler.py`

Potentially fixes https://github.com/hummingbot/hummingbot/issues/8457 

Supersedes #8460 (closed after renaming the branch to follow the `fix/` naming convention).


cardosofede:
@rapcmia this is ready to test

rapcmia (approved):
- Test repeated burst, created 10 requests at the same time and temporarily set the local  limit to 3 request per 10s. 
- The test ran 10x and each time, only 3 requests were allowed through and the other 7 waited
- Utilize agent to sent 10 small post-only BTCFDUSD buy orders at once, we temporarily set the local limit t o3 orders per 10s as well.
- Binance received the orders in a safe groups of 3, 3, 3 and 1. All 10 orders were cancelled afterward and none were filled
- We ran same controlled 10 small post-only tests on development branch. Even with a limit of 3, the development allowed all 10 request through which confirms the race condition exists there.
- Ran all related API throttler tests locally and all 16 passed

[live_binance_throttler_burst.py](https://github.com/user-attachments/files/32547398/live_binance_throttler_burst.py)

cardosofede (approved):

PR #8465, opened 2026-09-18
fix(xrpl): preserve custom market name when resolving token symbols

### LINKED ISSUE #8463: XRPL - custom market name is ignored, so two tokens with the same code overwrite each other in balances
### Describe the bug

On XRPL a token is identified by its currency code together with the account that issued it, so USDC from Circle and USDC from GateHub are different assets that share a code. A custom_markets entry was added for the GateHub token under the name USDC.gh-XRP and it did appear in the trading pair list, but balance kept showing a single USDC figure matching only one of the two holdings on the wallet.

Image

The name given to a custom market is not used when the connector decides what to call a token in balances. Built-in markets do use theirs, which is how the shipped list keeps Bitstamp's USD separate as USD.b, but custom entries are merged in without that step and fall back to the plain currency code unless trading_pair_symbol is filled in. Two markets for the same code then produce the same name, the balance list keeps one value per name, and one holding silently overwrites the other. Nothing is flagged as missing or wrong, the surviving figure depends on the order the ledger returns the trustlines, and that understated balance also feeds anything that sizes orders from available funds.

### Steps to reproduce

1. Use an XRPL wallet holding USDC from Circle `rGm7WCVp9gb4jZHWTEtGUr4dd74z2XuWhE` and USDC from GateHub `rcEGREd8NmkKRE8GE424sksyt1tJVFZwu`, with clearly different amounts on each.
2. Run `connect xrpl` with that wallet.
3. In conf/connectors/xrpl.yml, add a custom_markets entry named USDC.gh-XRP for the GateHub issuer and leave trading_pair_symbol unset.
4. Restart the client and run `balance`.
5. Only one USDC row appears, matching one holding; the other is absent even though the ledger shows both.

### Release version

dev-2.17

### Type of installation

Source

### Attach required files

na

Tyagiquamar:
### Description

When configuring XRPL with custom markets (where currencies share ticker names from different issuers, e.g. `USDC.gh` vs `USDC.rh`), `_make_xrpl_trading_pairs_request()` (`hummingbot/connector/exchange/xrpl/xrpl_exchange.py`) merged custom market dictionaries without populating `trading_pair_symbol` unless explicitly given in the config dictionary.

Because `trading_pair_symbol` remained `None`, `get_token_symbol_from_all_markets()` returned `None` when looking up tokens for custom pairs. The balance handler fell back to raw currency ticker symbols (e.g. `USDC`), causing balances from different issuers to overwrite each other under the same dictionary key.

### Changes
- In `_make_xrpl_trading_pairs_request()`, default `trading_pair_symbol` to the custom market key `k` if not explicitly specified, matching the behavior of default built-in market definitions.
- Added regression test `test_get_token_symbol_from_custom_market_without_explicit_trading_pair_symbol` in `test/hummingbot/connector/exchange/xrpl/test_xrpl_exchange_network.py`.

Closes #8463


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

